PR #78, opened 2021-08-03
Document how to launch non default version of julia

piever:
Thanks for this library, I find it super useful. This PR just adds a short note to the README on how to launch non default versions of julia (I had some trouble figuring it out from the docs).


davidanthoff:
Thanks!

Issue #26, opened 2021-05-24
Take JULIA_DEPOT_PATH into account


davidanthoff:
At the moment everything is hard coded to store Julia versions in `~/.julia/juliaup`. We should change that to check whether `JULIA_DEPOT_PATH` is configured and then use the value there if present.

heetbeet:
I would really love this feature, I snooped around a bit and found the following two places where the `~/.julia/juliaup` path is assumed:
https://github.com/JuliaLang/juliaup/blob/0bfe67d43677f50d148bd97deb47551d511ea825/Juliaup/src/Juliaup.jl#L19
https://github.com/JuliaLang/juliaup/blob/0bfe67d43677f50d148bd97deb47551d511ea825/launcher/launcher.cpp#L310

What I wonder is 
- How do the different versions of Julia handle the environment variable?
- Will it require more work that just using `DEPOT_PATH[0]` in the julia code and something like `Is JULIA_DEPOT_PATH empty? Then return "~/.julia/juliaup" else return JULIA_DEPOT_PATH (split at ";")` in the C++ part?
- Does this issue break version-compatibility? https://github.com/JuliaLang/julia/issues/31002
- And if version compatibility is broken, does it matter? I.e. what do we want to do when an old version of Julia doesn't understand `JULIA_DEPOT_PATH` and uses `~/.julia` or something else, while juliahub wants to use `JULIA_DEPOT_PATH/juliahub`. My current thinking is that this might actually be preferred since then at least juliahub will always be consistent and located where never versions would have wanted to place it.

I have a hunch that if we use `DEPOT_PATH[0]` in julia and add a preference for the environment variable `JULIA_DEPOT_PATH ` (split at `;`) in C++, we will be A-okay.

davidanthoff:
Yeah, I think that should pretty much be it. Couple of more specifics:
- We can't use `DEPOT_PATH` in the Julia code because it isn't populated when `juliaup` is compiled with PackageCompiler. So on the Julia side we'll also have to look for the env variable explicitly, parse it ourselves and extract the right values etc. Should not be hard at all, though.
- I think your logic for the C++ code is spot on (and the same would be used in the Julia code). Only "extra" thing I can think of is that on non Windows platforms `:` is also a valid splitter in `JULIA_DEPOT_PATH`, and we might just make it cross platform compatible from the beginning?
[...]

PR #83, opened 2021-08-04
Take JULIA_DEPOT_PATH into account

heetbeet:
https://github.com/JuliaLang/juliaup/issues/26

I cannot find a way to easily run the C++ code (or compile it from my computer), it does compile in GithubActions, so I will need a review on this. Also, we probably need some test to see if the DEPOT_PATH works as advertised.

Also, I agree with "on non Windows platforms `:` is also a valid splitter in `JULIA_DEPOT_PATH`, and we might just make it cross platform compatible from the beginning", but one of the bigger things to first figure out is how to compile the C++ code on Linux and OSX. The C++ function I added relies so heavily on the win32 API that it is probably moot to try and optimize for `:` at this moment.

### DIFF
--- a/Juliaup/src/Juliaup.jl
+++ b/Juliaup/src/Juliaup.jl
@@ -16,14 +16,14 @@ function julia_main()
     return 0
 end
 
-function get_juliauphome_path()	
-	depotpaths = get(ENV, "JULIA_DEPOT_PATH", nothing) 
-
-	if depotpaths === nothing
-		return joinpath(homedir(), ".julia", "juliaup")
-	else
-		return split(depotpaths, Sys.iswindows() ? ';' : ':')[1]
+function get_juliauphome_path()
+	depot_path = split(get(ENV, "JULIA_DEPOT_PATH", ""), Sys.iswindows() ? ';' : ':')[1]
+	
+	if isempty(depot_path)
+		depot_path =  joinpath(homedir(), ".julia", "juliaup")
 	end
+	
+	return depot_path
 end
 
 function get_juliaupconfig_path()
--- a/Juliaup/src/Juliaup.jl
+++ b/Juliaup/src/Juliaup.jl
@@ -19,7 +19,7 @@ end
 function get_juliauphome_path()	
 	depotpaths = get(ENV, "JULIA_DEPOT_PATH", nothing) 
 
-	if depotpaths == nothing
+	if depotpaths === nothing
 		return joinpath(homedir(), ".julia", "juliaup")
 	else
 		return split(depotpaths, Sys.iswindows() ? ';' : ':')[1]
--- a/Juliaup/src/Juliaup.jl
+++ b/Juliaup/src/Juliaup.jl
@@ -16,9 +16,14 @@ function julia_main()
     return 0
 end
 
-function get_juliauphome_path()
-	# TODO Handle JULIA_DEPOT env var here
-	return joinpath(homedir(), ".julia", "juliaup")
+function get_juliauphome_path()	
+	depotpaths = get(ENV, "JULIA_DEPOT_PATH", nothing) 
+
+	if depotpaths == nothing
+		return joinpath(homedir(), ".julia", "juliaup")
+	else
+		return split(depotpaths, Sys.iswindows() ? ';' : ':')[1]
+	end
 end
 
 function get_juliaupconfig_path()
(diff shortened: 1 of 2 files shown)


Juliaup/src/Juliaup.jl
```diff
@@ -16,9 +16,14 @@ function julia_main()
...
-function get_juliauphome_path()
-	# TODO Handle JULIA_DEPOT env var here
-	return joinpath(homedir(), ".julia", "juliaup")
+function get_juliauphome_path()	
+	depotpaths = get(ENV, "JULIA_DEPOT_PATH", nothing) 
+
+	if depotpaths === nothing
+		return joinpath(homedir(), ".julia", "juliaup")
+	else
+		return split(depotpaths, Sys.iswindows() ? ';' : ':')[1]
```
davidanthoff, line 25: One thing that is not clear to me is whether the base implementation has different separators per platform, or whether you can use potentially both separators on one of these platforms?

launcher/launcher.cpp
```diff
@@ -307,9 +307,26 @@ path GetHomedirPath() {
 	return buffer;
 }
 
+path GetDepotPath() {
+	string buffer;
+	buffer.resize(MAX_PATH);
+	DWORD pathLen = MAX_PATH;
+	
+	pathLen = GetEnvironmentVariableA("JULIA_DEPOT_PATH", &buffer[0], pathLen);
```
davidanthoff, line 315: I'm wondering whether we could use `std::getenv` here? If that works on Windows, then we could make this cross platform right away? It might look like this:
```cpp
const char *tmp = std::getenv("JULIA_DEPOT_PATH");

if(tmp) {
  string env_var {tmp};
  // Do the parsing etc here.
}
else {
  return GetHomedirPath() / ".julia";
}
```

davidanthoff (changes requested):
Fantastic, thanks a lot! I have two (hopefully small) comments.

I'll try to write up how one can compile and test this locally. A bit cumbersome, but it does work with stuff that is free.

Juliaup/src/Juliaup.jl
heetbeet: The `Sys.iswindows()` check will guide the seperator split, got the code from [here](https://stackoverflow.com/a/64595654/1490584). It doesn't seem like Julia base provides this functionality as per the stack overflow comment. I think it is a bit dangerous to use _both_ characters as a legal split, since `:` is found in a lot of windows paths like `c:\` and I think `;` is a legal filename character in Linux (not sure if it is found in practice).

I found where it is introduce in Julia: https://github.com/JuliaLang/julia/blob/0e094a818de1618b58edd56b759ed808accd3f9f/base/initdefs.jl#L98, I'll use something closer to that implementation.

launcher/launcher.cpp
heetbeet: Yes, I agree this will immediately be more platform independent. What to do then about the `;`/`:` split? Maybe we can get away with a
```
#ifdef _WIN32
    #define ENVPATHSEP ';'
#else
    #define ENVPATHSEP ':'
#endif
```
But it would be nicer if we don't need a directive.

heetbeet:
```
const char *tmp = std::getenv("JULIA_DEPOT_PATH");

if(tmp) {
  string env_var {tmp};

  auto cutoff = env_var.find(ENVPATHSEP);
  if(cutoff != std::string::npos){
    env_var.resize(cutoff);
  }
  
  if(env_var.size() == 0){
      return env_var;
  }
}
  
return fallback;
```

heetbeet: Maybe something like this? I can't really test this at the moment, so I'll just leave it here as a comment. (Maybe I'll get back to this sometime next week, but wouldn't be able to itch my FOSS addiction for the rest of this week at least)

Juliaup/src/Juliaup.jl
davidanthoff: Perfect, that makes sense!

launcher/launcher.cpp
heetbeet: Okay, I added some cross-platform code that work on Windows, but haven't been tested on Linux yet.

davidanthoff:
@heetbeet, I rewrote `juliaup` in Rust... See https://github.com/JuliaLang/juliaup/pull/89. I'm sorry for that 180, but it turns out there are many benefits to that, in particular when it comes to the cross platform solution. I'll try to incorporate this PR into the Rust branch before I merge it.

PR #110, opened 2021-08-11
[CI] automatically update AUR script

simeonschaub:
This should automatically update https://aur.archlinux.org/packages/juliaup/. This is untested, so no guarantees that this works. The SSH keys will also need to be added as a secret here. Is this generally a good idea, or would it be better to update it manually for now?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -246,6 +246,7 @@ jobs:
 
   deploy-aur:
     needs: [deploy-github-release]
+    environment: release-channel
     runs-on: ubuntu-latest
     if: startsWith(github.ref, 'refs/tags/')
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -243,3 +243,34 @@ jobs:
         $cred = New-Object System.Management.Automation.PSCredential $env:CLIENTID, $sec
         Set-StoreBrokerAuthentication -TenantId $env:TENANTID -Credential $cred
         Update-ApplicationSubmission -ReplacePackages -AppId 9NJNWW8PVKMN -SubmissionDataPath ".\Upload.json" -PackagePath ".\Upload.zip" -AutoCommit -Force
+
+  deploy-aur:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -243,3 +243,34 @@ jobs:
         $cred = New-Object System.Management.Automation.PSCredential $env:CLIENTID, $sec
         Set-StoreBrokerAuthentication -TenantId $env:TENANTID -Credential $cred
         Update-ApplicationSubmission -ReplacePackages -AppId 9NJNWW8PVKMN -SubmissionDataPath ".\Upload.json" -PackagePath ".\Upload.zip" -AutoCommit -Force
+
+  deploy-aur:
+    needs: [deploy-github-release]
+    runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/tags/')
```
davidanthoff, line 250: We should probably only run this for builds that we push into the release channel, right? So you would just add `environment: release-channel` as a line here, and then it is gated behind the release approval.

davidanthoff (changes requested):
Ah, that is great! And I think we can put this into place now.

Only caveat: I plan to change the name of the files that are attached to GitHub Releases, hopefully very soon, I assume that will break things on the other end? I'll try to get that change in ASAP, though.

simeonschaub:
> Only caveat: I plan to change the name of the files that are attached to GitHub Releases, hopefully very soon, I assume that will break things on the other end? I'll try to get that change in ASAP, though.

This only uses the source tarball, so it's probably still fine.

.github/workflows/release.yml
davidanthoff: Yes, it only runs for tags, but I think we want to gate it a bit more. I tag a lot, but many of these versions are crap and we don't want them to end up released. The environment stuff handles that, essentially.

simeonschaub: Sounds good. I have changed that now.

PR #113, opened 2021-08-11
fix unused import warning

simeonschaub:


davidanthoff:
We needed that for the Windows build, and with the latest `master` for all the other builds as well, because I finally did add more useful error handling.

PR #120, opened 2021-08-12
move AUR job to dev channel

simeonschaub:
Once development settles down a bit, we can always move this to a more stable channel, but I think for now it's reasonable if this just follows the dev releases to get new features out there right away. The key should be set up already, IIRC.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -246,7 +246,7 @@ jobs:
 
   deploy-release-channel-aur:
     needs: build-juliaup
-    environment: release-preview-channel
+    environment: dev-channel
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


davidanthoff:
Are you sure? :) The mode I'm in right now is that I tag dev channel builds when I want to try something out, so they are very often _completely_ broken. I sometimes for example publish builds that are on purpose non-functional so that I can test something out in the deployment pipeline. Whenever I have a build that is somewhat working, I can push it to the release-preview, so I think that might be a better match at the moment?

simeonschaub:
> Whenever I have a build that is somewhat working, I can push it to the release-preview, so I think that might be a better match at the moment?

Ok, sounds reasonable. Then lets leave it like it is currently.

PR #128, opened 2021-08-18
use `bail!` instead of `return(Err(anyhow!(...)))`

KristofferC:
just getting my feet wet a little


davidanthoff (approved):
Ah, that is much nicer :)

PR #132, opened 2021-08-18
make the download progress bar consistent with the Pkg theme

KristofferC:
I thought it would look nice if Pkg and juliaup were consistent with the Pkg progress bars.

Here is Pkg:
image

And here is juliaup with the PR:
image

### DIFF
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -5,6 +5,7 @@ use crate::jsonstructs_versionsdb::JuliaupVersionDB;
 use crate::utils::get_juliaup_home_path;
 use crate::utils::parse_versionstring;
 use anyhow::{anyhow, Context, Result};
+use console::style;
 use flate2::read::GzDecoder;
 use indicatif::{ProgressBar, ProgressStyle};
 use std::{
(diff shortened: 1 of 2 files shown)


src/operations.rs
```diff
@@ -86,7 +88,7 @@ pub fn install_version(
 
     std::fs::create_dir_all(target_path.parent().unwrap())?;
 
-    eprintln!("Installing Julia {} ({}).", version, platform);
+    eprintln!("{} Julia {} ({}).", style("Installing").green().bold(), version, platform);
```
davidanthoff, line 91: We might check whether that `console` package is smart enough to detect whether we are connected to a tty or not, and then only style things when we are.

davidanthoff (approved):
Yes, definitely better!

src/operations.rs
KristofferC: Looking at https://docs.rs/console/0.14.1/console/fn.colors_enabled.html, I think this should be disabled when the process is piped but good to confirm.

davidanthoff: I also looked through their issues, and I think this should be ok. We use the `atty` package to detect that, they also did at some point, then removed it because they said they had all that logic vendored, so clearly they are aware of this.

PR #134, opened 2021-08-19
improve error message when trying to set an uninstalled julia as default 

KristofferC:


Cargo.lock
```diff
@@ -108,7 +108,9 @@ dependencies = [
  "encode_unicode",
  "lazy_static",
  "libc",
+ "regex",
```
KristofferC, line 111: My IDE kept addinga this stuff...

PR #135, opened 2021-08-19
make cargo test pass

KristofferC:
I guess CI should be set up at some point to run the tests?


davidanthoff:
> I guess CI should be set up at some point to run the tests?

Yes, definitely, I just hadn't figured out how to do that. Although it probably is just another step with a cargo command, `test`.

Issue #84, opened 2021-08-04
How to distribute to Linux and OSX?


heetbeet:
Maybe we should initially bypass the Linux bureaucracy by creating an `sh` installer similar to Anaconda.

Here is the Anaconda installer header from `curl -r 0-26557 https://repo.anaconda.com/archive/Anaconda3-2021.05-Linux-x86_64.sh`: 
[Anaconda3-2021.05-Linux-x86_64.sh.zip](https://github.com/JuliaLang/juliaup/files/6934389/Anaconda3-2021.05-Linux-x86_64.sh.zip)

Here is an initial attempt into making it into something a bit more generic:
https://github.com/heetbeet/BashTemplateInstaller/blob/main/bash_installer.sh

This header seems to work for both Linux and OSX. I'm not sure though how the copyright will works, if it's a problem, it's easy enough to write something from scratch.

davidanthoff:
Yeah, I like that approach! Rust also does that, so we might also be able to copy from them. And with the rewrite in Rust we will have just two very small binaries and that should actually be enough to get started.

simeonschaub:
I have created a `juliaup` build script for arch [here](https://aur.archlinux.org/packages/juliaup/) in case anyone is interested. For now, that will need to be updated manually, but I think it should also be possible to do this automatically as part of CI.

davidanthoff:
Oh, and one other major question: how do we "brand" ourselves in third party stores? In the Windows Store we are just `Julia`, _not_ `juliaup`. I think that if we decide that this will become the official Julia setup story, then we should try to register everywhere as just `julia`. After all, this package _does_ put `julia` on the `PATH`, and it just also ships a helper binary that is called `juliaup`.

We might need a transition phase, though, where we register as `juliaup` until everything has gotten more stable.

KristofferC:
Why can't juliaup update itself?

davidanthoff:
> Couldn't juliaup just update itself?

So in the case of a standalone install, [this](https://github.com/JuliaLang/juliaup/issues/105) would do that. Or were you thinking auto-update itself? But when distributed via a third-party mechanism, then I'm not so sure whether we want it to self-update, right? Certainly for the Windows Store that won't work (because it doesn't allow us to change the files it distributes), and presumably other things like brew, conda etc would also not like that?

KristofferC:
>  But when distributed via a third-party mechanism, then I'm not so sure whether we want it to self-update, right? 

[...]

PR #141, opened 2021-09-09
added basic shell bootstrap script

ExpandingMan:
Hello all.  Following up on a discussion some of us had on slack: to a linux user this package looks a little bit weird.  The reason for this is that rust binaries are usually built from scratch with `cargo` if they are not available from the package manager.  That would be pretty awkward for `juliaup` since it's effectively asking users to install one entire package manager (`cargo`) to download another package manager (`Pkg`).  To put it another way: to a linux user it looks a little weird that this is anything other than shell scripts and Julia code.

To rectify this, my suggestion would be to make it as easy as possible to fetch the binary to an appropriate place (this can ultimately be reduced to pretty much two clicks and hitting enter).  This is [the same approach that rustup](https://rustup.rs/) takes.

To help get things going, I have added a basic shell script for downloading the binary and sticking it in what I deemed to be an appropriate location.  There's obviously a lot more that can be done here, most importantly handling of different architectures and weird edge cases, but I thought it might make a good starting point (and yes, it works).

If this is something you indeed want to do, it might be good to secure a julialang.org url for this script somewhere.

This is somewhat similar to the bootstrapping [install script](https://gitlab.com/ExpandingMan/dotfiles/-/blob/master/install) I've been using for quite a while for my dotfiles installer which is written in Julia.

In retrospect, I realize I was probably overzealous in making changes to the README since you may not want to expose this from this initial PR alone, so I'd be glad to revert that...

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,11 +7,21 @@ The installer also bundles a full Julia version manager called `juliaup`. One ca
 ## Using Juliaup
 
 If you want to try it, here is what you should do:
+
+### Windows Users
 - Make sure you don't have any version of Julia on your PATH. `Juliaup` will handle all `PATH` related aspects of your Julia installation.
 - Install Julia from the Windows Store [here](https://www.microsoft.com/store/apps/9NJNWW8PVKMN).
 
 Once you have that installed, `julia` is on the `PATH`, there is a start menu shortcut and it will show up as a profile in Windows Terminal. Any of those will start Julia. The VS Code extension will also automatically find this Julia installation.
 
+### Linux Users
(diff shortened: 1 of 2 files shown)


davidanthoff:
Did you see https://github.com/JuliaLang/juliaup/issues/84 and what @heetbeet has been working on? Same idea, right?

heetbeet:
Good reminder for me... I'll put some effort in today to see if I can get my script into working order. @ExpandingMan would you mind checking my work when I'm done, then we can compare notes

ExpandingMan:
No I hadn't seen it.  In fact, I was thinking this morning "Why didn't I just go through `rustup-init.sh` and re-use that?"

Looks like you have mostly done that already, so I'll close this.  Thanks all!

PR #149, opened 2021-09-26
support for linux aarch64

simeonschaub:
I have followed the example for Windows and MacOS here and used `~x64`
in the version specifier, but we might want to change that to `~aarch64`
instead.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -91,6 +91,11 @@ fn produce_version_db() -> Result<JuliaupVersionDB> {
                 format!("{}+0~x86", v),
                 JuliaupVersionDBVersion {url: format!("https://julialang-s3.julialang.org/bin/linux/x86/{}.{}/julia-{}-linux-i686.tar.gz", v.major, v.minor, v)}
             );
+        } else if target_os == "linux" && target_arch == "aarch64" {
+            db.available_versions.insert(
+                format!("{}+0~x64", v),
+                JuliaupVersionDBVersion {url: format!("https://julialang-s3.julialang.org/bin/linux/aarch64/{}.{}/julia-{}-linux-aarch64.tar.gz", v.major, v.minor, v)}
+            );
         } else if target_os == "macos" && target_arch == "x86_64"{
             db.available_versions.insert(
                 format!("{}+0~x64", v),
@@ -155,6 +160,19 @@ fn produce_version_db() -> Result<JuliaupVersionDB> {
                     format!("{}~x86", v),
                     JuliaupVersionDBChannel {
                         version: format!("{}+0~x86", v),
+                    },
+                );
+            } else if target_os == "linux" {
+                db.available_channels.insert(
+                    format!("{}", v),
+                    JuliaupVersionDBChannel {
+                        version: format!("{}+0~x64", v),
+                    },
+                );
+                db.available_channels.insert(
+                    format!("{}~x64", v),
+                    JuliaupVersionDBChannel {
+                        version: format!("{}+0~x64", v),
                     },
                 );
             } else if target_os == "macos" {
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -91,6 +91,11 @@ fn produce_version_db() -> Result<JuliaupVersionDB> {
                 format!("{}+0~x86", v),
                 JuliaupVersionDBVersion {url: format!("https://julialang-s3.julialang.org/bin/linux/x86/{}.{}/julia-{}-linux-i686.tar.gz", v.major, v.minor, v)}
             );
+        } else if target_os == "linux" && target_arch == "aarch64" {
+            db.available_versions.insert(
+                format!("{}+0~x64", v),
+                JuliaupVersionDBVersion {url: format!("https://julialang-s3.julialang.org/bin/linux/aarch64/{}.{}/julia-{}-linux-aarch64.tar.gz", v.major, v.minor, v)}
```
davidanthoff, line 97: So the idea here is that the platform in the `format!` call and the one in the URL always match here, so in this case it should be `format!("{}+0~aarch64", v)`, I think.

For Mac and Windows if the `target_arch` is `aarch64`, I currently just ship the x64 (or was it x86?) builds and then let emulation on those platforms handle things. But that is only the case because we don't have native arm builds for Windows and Mac yet (or at least not for a released version).

build.rs
```diff
@@ -157,6 +162,19 @@ fn produce_version_db() -> Result<JuliaupVersionDB> {
...
+                    JuliaupVersionDBChannel {
+                        version: format!("{}+0~x64", v),
+                    },
+                );
+                db.available_channels.insert(
+                    format!("{}~x64", v),
+                    JuliaupVersionDBChannel {
+                        version: format!("{}+0~x64", v),
+                    },
+                );
```
davidanthoff, line 177: This should all be `aarch64`. I think Linux doesn't have emulation, right? If it did, we could add additional platforms here, i.e. that would then add the ability to say run the x64 binaries on aarch64 on Linux, if that was supported.

davidanthoff (changes requested):
You already pointed this out: we should name all of this `aarch64`, not `x64`.

The Windows and Mac situation is such that the `aarcht64` channels at the moment point to the x64 binaries because there isn't an `aarch64` build, but we don't have to do that for Linux, I believe, given that we actually have proper native builds.

davidanthoff:
I'm updating this branch right now, will push soon.

davidanthoff (approved):

davidanthoff:
@simeonschaub I updated the branch and to me it all looks good now. Could you also take another look whether this makes sense to you?

davidanthoff:
And then once you can confirm that this works, we need to add it to our Github Actions release pipeline as well, of course!

simeonschaub:
Ok, I can confirm that it still builds just fine on my android phone. GitHub doesn't allow me to approve my own PRs, but feel free to merge whenever.

> And then once you can confirm that this works, we need to add it to our Github Actions release pipeline as well, of course!

Have you ever looked into cross compilation with Rust? I must admit I don't have any experience in that regard. I almost wonder whether it might be easiest to just take advantage of the BinaryBuilder infrastructure, I believe it does support Rust as well.

davidanthoff:
> Have you ever looked into cross compilation with Rust?

Yes, and it is very easy and all fully integrated into the existing GitHub Actions. I'll update the workflow.

davidanthoff:
@simeonschaub could you try the binaries that are now build as part of the release process https://github.com/JuliaLang/juliaup/releases/tag/v1.2.11 ? I also added M1 Apple builds while I was at it.

simeonschaub:
I noticed that the binaries didn't have the executable bit set, otherwise everything worked great though.

PR #158, opened 2021-11-20
Fix Ctrl-C handling

### LINKED ISSUE #157: Ctrl+C ruins the terminal
a video is worth probably a few thousand words, so here is the issue in action

https://user-images.githubusercontent.com/35898736/142609384-27707309-fb4f-4b1f-8153-64f97880f8ea.mp4

some (most?) ctrl+C actions errors and quits Julia, and worst than that when relaunching julia in the same terminal, the input seems to be scrambled. The only solution is to start a new terminal

This only happens with juliaup (not if installed directly) and only when launched from another terminal (cmd too), it doesn't happen when I click directly on the julia app

davidanthoff:
Fixes #157. The idea is that we handle Ctrl-C, but then just don't do anything in the launcher process, as we just want the Julia child process to handle this entirely.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,9 +39,9 @@ checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
 name = "bitflags"
-version = "1.2.1"
+version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


davidanthoff:
@simeonschaub I'm going to merge this and release in the release preview channel, maybe you could then test whether this also works on non Windows, or at least the platform that you usually use?

simeonschaub:
I can confirm that this fixes the issue for me on Linux.

PR #190, opened 2021-12-25
Locking

### LINKED ISSUE #66: Lock config file during update
`juliaup` should lock the config file while it is making changes. That way we can essentially use that as a lock.

davidanthoff:
Fixes #66.

The design now is that we create an empty file just to us as a global lock.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,6 +171,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "cluFlock"
+version = "1.2.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


davidanthoff:
@simeonschaub I should probably have waited for a review from you on this PR :) If you still want to review after the fact, it would be great!

simeonschaub:
No, this looks great! I am not sure how to properly test whether the lock does what it's supposed to, but I have been using this a bit and so far haven't run into any issues. The code changes also all look reasonable to me.

davidanthoff:
The tests that I ran is to open a couple of terminals, execute `juliaup add SOMETHING` in one for a channel that I know is not installed, and then immediately run `julia` or `juliaup add SOMETHING ELSE` in one of the other terminal windows. The two second windows should then wait before they do anything until the first one is finished.

PR #193, opened 2021-12-27
Add support for cron auto update job

davidanthoff:


src/command_selfuninstall.rs
```diff
@@ -0,0 +1,68 @@
...
+        Ok(val) => {            
+            std::process::Command::new("schtasks.exe")
+                .args([
+                    "/delete",
+                    "/tn",
+                    &format!("Juliaup self update for WSL {} distribution", val),
+                    "/f",
+                ])
+                .output()
+                .with_context(|| "Failed to remove Windows task for juliaupa.")?;
```
deoradh, line 20:
```suggestion
                .with_context(|| "Failed to remove Windows task for juliaup.")?;
```

simeonschaub:
So is the idea that this always gets run by the install script or will this need to be set up manually? Perhaps we can also give this a different name to make clear that this just sets up a cron job and doesn't really install Juliaup itself.

davidanthoff:
My medium term plan is that we move as much stuff as possible from the shell script into `juliaup self install`, and then always run `juliaup self install` at the end of the shell script. That in particular includes all the steps we do in the shell script right now to add `juliaup` to the `PATH`.

But for now I was going to merge this as is, just to move forward, and then we can tweak and extend. I think before we move the shell script stuff, we minimally need to add some interactivity to `juliaup self install` to ask whether users are OK with all of this, and also provide some command line args. I was thinking things like `--enable-background-update=true/false` and `--enable-path-modification=true/false`, and then those could also be specified with the shell script and it would just pass these through.

I think it is OK, though, to have the entire `self *` set of commands not be perfect before we merge them, really the only two deployment methods that I consider production ready is the Windows Store and AUR story, and neither is affected by this.

simeonschaub:
Yes, I think that sounds like a plan. Does `juliaup self install` already get called at the end of the shell script or not yet?

davidanthoff:
> Does `juliaup self install` already get called at the end of the shell script or not yet?

Not yet, and maybe we leave it like that until we're more comfortable that it all works as we want it to?

simeonschaub:
Yes, that's what I was going to say as well.

simeonschaub (approved):
Would be good to fix the unused import warnings, but otherwise this looks good to me.

PR #198, opened 2021-12-28
Use clap for configuration

davidanthoff:
@simeonschaub I think this might be a bit nicer, to just rely on `clap` for the parsing of the specific configuration options? We can then also use the clap help system, for example.

I also slightly changed the config thing to only serialize the setting if it is `true`.


simeonschaub:
Ah, I didn't know clap could do that! What I tried to do in #165 was to make adding new config options easier by being able to reuse a lot of the logic in `run_command_config` and only needing to add a new case to the `match` statement. By moving each config option to its own function, I'm slightly worried that a lot of the code in there will need to be copy-pasted for new options. Maybe that's thinking a bit too far ahead for now though.

davidanthoff:
I guess we'll have two types of configuration options: things like `symlinks` where a change of a configuration option actually triggers some pretty involved logic, and more "standard" cases where we just change a value and then the program changes its behavior (but we don't have one like that yet). The next think I wanted to add was a configuration setting for the cron job stuff, which is similar to `symlinks` in that a change would trigger some pretty involved logic.

I think for the case of the "involved" configurations, a common command won't help much, much of the fun is unique in any case, and I kind of like to have separate functions. For the "standard" case we could just have the `clap` match block call a common function where we pass the name of the property, like in your original PR, right? In that case we could still use `clap` for validation, help text (all of that also tied nicely into conditional compilation), but then share code?

simeonschaub:
My thinking was that both cases could be covered with this approach, since the `match` bodies can still contain arbitrary logic for what to do if a setting was changed as in the `symlinks` case. I'll leave it up to you whether you want to merge this PR for now though, we can also discuss this further once we are actually adding new config options.

davidanthoff:
I'm going to merge for now, just because I like the UI integration we get from clap. If this turns out too cumbersome, happy to go back to something different later again!

PR #206, opened 2021-12-30
Modify path as part of config

davidanthoff:


simeonschaub:
This is great! It's quite bash specific right now though -- we should probably at least think about zsh as well since I believe that's now the default shell on macOS and it's also quite popular on Linux.

Also, are you aware of the warning in https://github.com/JuliaLang/juliaup/blob/602ad0c394447effc63e2bebc565325b535bdc59/src/operations.rs#L260? It would probably make sense to ask if Juliaup should automatically add that directory to the path. (This might not necessarily be the same directory as the `juliaup` executable though.)

davidanthoff:
> It's quite bash specific right now though -- we should probably at least think about zsh as well since I believe that's now the default shell on macOS and it's also quite popular on Linux.

Yes, agreed! I'll merge this PR and we can add in a follow up PR. I think essentially over time we'll just have to add specific code for any shell we find on the system....

> It would probably make sense to ask if Juliaup should automatically add that directory to the path.

So in a user install that folder will always be the parent of `juliaup` itself, right? So that case should be handled already? And if it is a system install, then it is putting things into `~/.local/bin`? Would the suggestion be that we add that folder to the `PATH` if it isn't? I'm a bit nervous about that, it seems that is modifying a bit too much?

In general, I think this whole script file modifying business is pretty weird... I think we need it, but I was also thinking that we could also detect the situation where `~/.local/bin` _is_ already on the `PATH`, and the just add symlinks there and not modify any files? I still think that is generally a much nicer solution... But all for future PRs :)

Issue #170, opened 2021-11-22
Disabling certificate validation


darleybarreto:
I folks, I tried to install julia 1.6.4 on a Windows 10 machine with some security constraints and I got this error

```bash
$ juliaup add 1.6.4
Installing Julia 1.6.4+0 (x64).
Error: Failed to download from url `https://julialang-s3.julialang.org/bin/winnt/x64/1.6/julia-1.6.4-win64.tar.gz`

Caused by:
	0: https://julialang-s3.julialang.org/bin/winnt/x64/1.6/julia-1.6.4-win64.tar.gz: Connection Failed: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
	1: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
```

Would it be possible to disable certificate validation?

davidanthoff:
I don't think so, that is really the only defense we have against some really bad attack vectors at the moment. If we were to disable this, anyone between your machine and the server where the files are hosted could just inject whatever they wanted and send you that. I think what this error suggests is that your system doesn't rust the CA issuer that created the certificate for the domain from which we are fetching the data. Unclear to me why that would be the case, maybe @staticfloat has an idea?

darleybarreto:
I see. I wonder what rustup do, then, because I was able to install it here without issues.

davidanthoff:
Maybe their certificate was issued by a different authority that is actually whitelisted on your system?

Also, note to self: I should check how the Rust crate we are using is actually handling this. I'm assuming it is just looking at some system store of whitelisted certificate authorities, but we could double check.

staticfloat:
Both of the certificates are valid; so either the root is not trusted by your system for some reason (does going there in a browser work for you?), or someone is indeed MITM'ing you.  :P

darleybarreto:
> does going there in a browser work for you?

Yes, that works.

StefanKarpinski:
The best option is to use a native system TLS engine which is both more secure (does certificate revocation checks on Windows and macOS) and is more likely to have MITM firewall certs added to it by the IT department. If you can point me at where the code for doing the downloads that's failing is, I might be able to save you some pain since I've already been through a lot of this for Julia itself.

davidanthoff:
[...]

PR #216, opened 2022-01-07
Use system certs for TLS

davidanthoff:
Fixes https://github.com/JuliaLang/juliaup/issues/170.

@StefanKarpinski could you take a look at https://github.com/rustls/rustls-native-certs#should-i-use-this-or-webpki-roots? This PR would move us to using what is described there, but there are some caveats mentioned (i.e. not using OS level verification and poor quality on debian based Linux) that I have a hard time evaluating. Are those caveats OK?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -254,6 +254,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "core-foundation"
+version = "0.9.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6888e10551bb93e424d8df1d07f1a8b4fceb0001a3a4b048bfc47554946f47b3"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
+]
+
+[[package]]
+name = "core-foundation-sys"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5827cebf4670468b8772dd191856768aedcb1b0278a04f989f7766351917b9dc"
+
+[[package]]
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
That's actually a package that I looked into porting code from for Julia at one point and I do think this is a good solution. Caveat: it turns out to be even better (as they mention in the README) to just let the system validate TLS certificates for you since that, in addition to handling transparent MITM proxies, lets the system do certificate revocation checking, which there's no (good) way to do with PEM files. Not sure how feasible that is in Rust. We had to change the way we configure libcurl and libgit2, etc. to use native TLS engines on Windows and macOS. This is ok though—Linux doesn't even check for revoked certificates at all; browsers like Chrome do it themselves, but CLI tools on Linux don't do it at all.

davidanthoff:
I think the other alternative would be to use https://github.com/sfackler/rust-native-tls (docs https://docs.rs/native-tls/latest/native_tls/). If I understand that correctly, it would just call some OS API, and might be even better? It seems a bit more involved to use that package than this PR as one needs to make some code changes as well, but if you think that would be even better, we might as well use that.

davidanthoff:
@StefanKarpinski bump, any thoughts whether we should use this `rust-native-tls` instead?

tedbyron:
> whether we should use this `rust-native-tls` instead?

I would be interested in adding a feature to the crate for the option of native TLS. I'll send a PR

davidanthoff:
Closing this in favor of https://github.com/JuliaLang/juliaup/pull/249.

PR #237, opened 2022-01-25
clarify availability on Arch Linux in the README

simeonschaub:


davidanthoff:
Is there a canonical command to install it from there? You could even add something similar to the `brew` instructions that tell people what to type exactly?

simeonschaub:
Unfortunately there isn't really one standard AUR helper, so it will depend on the system. There are literally dozens of them: https://wiki.archlinux.org/title/AUR_helpers

davidanthoff:
Oh wow :) Sounds good.

PR #259, opened 2022-03-12
Add openSUSE TumbleWeed in Installation

uncomfyhalomacro:
Hello. I am maintaining juliaup in openSUSE. It is available now in opensuse:Factory. :partying_face:

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,16 @@ Note that the Mac and Linux version are considered prerelease, have known bugs a
 
 On Arch Linux, Juliaup is available [in the Arch User Repository (AUR)](https://aur.archlinux.org/packages/juliaup/).
 
+### openSUSE Tumbleweed
+
+On openSUSE Tumbleweed, Juliaup is available now in Factory. To install, run with root privileges:
+
+```sh
+zypper addrepo https://download.opensuse.org/repositories/openSUSE:Factory/standard/openSUSE:Factory.repo
+zypper refresh
+zypper install juliaup
+```
+
 ## Using Juliaup
 
 Once you have installed Juliaup, `julia` is on the `PATH`, and on Windows there is a start menu shortcut and it will show up as a profile in Windows Terminal. Any of those will start Julia. The VS Code extension will also automatically find this Julia installation.
--- a/README.md
+++ b/README.md
@@ -50,8 +50,6 @@ On Arch Linux, Juliaup is available [in the Arch User Repository (AUR)](https://
 On openSUSE Tumbleweed, Juliaup is available now in Factory. To install, run with root privileges:
 
 ```sh
-zypper addrepo https://download.opensuse.org/repositories/openSUSE:Factory/standard/openSUSE:Factory.repo
-zypper refresh
 zypper install juliaup
 ```
 
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ On Arch Linux, Juliaup is available [in the Arch User Repository (AUR)](https://
 
 ### openSUSE Tumbleweed
 
-On openSUSE Tumbleweed, Juliaup is available now in Factory. To install, run with root privileges:
+On openSUSE Tumbleweed, Juliaup is available. To install, run with root privileges:
 
 ```sh
 zypper install juliaup


davidanthoff:
Very cool! Happy to merge this, but I have a few questions first:
1. I think we eventually want to make the `juliaup` story the "official" Julia deployment story. At that point it would be great if `zypper install julia` would actually install `juliaup`. I assume we can change that later without too much trouble?
2. I don't fully understand the relationship between TumbleWeed and Factory :) https://en.opensuse.org/Portal:Factory makes it sound as if anything in Factory that passes some QA will automatically be pushed to TumbleWeed? Is that right? Is `zypper addrepo https://download.opensuse.org/repositories/openSUSE:Factory/standard/openSUSE:Factory.repo` then even needed?
3. For both brew and arch we automatically push new builds out to them whenever we release something, see https://github.com/JuliaLang/juliaup/blob/master/.github/workflows/release.yml#L413 and https://github.com/JuliaLang/juliaup/blob/master/.github/workflows/release.yml#L426. Would something similar make sense here to?

uncomfyhalomacro:
Hello again.

To answer the following questions:

1. `julia` is already available in openSUSE https://build.opensuse.org/package/show/openSUSE:Factory/julia. However, the version there is 1.6.3 which is outdated. If we want to make `julia` actually provide `juliaup`, there might be some magic in RPM spec that I need to tweak. I am new to openSUSE and packaging so I am still not knowledgeable if it allows that. Otherwise, we have to talk to the current maintainer of `julia`.

2. I think I also got a little confused too. If I understand correctly, once it passes openQA in Factory, it will become available in openSUSE TumbleWeed, hence the default -> `http://download.opensuse.org/tumbleweed/repo/oss/ (openSUSE:Factory)`. Therefore, no need to `zypper addrepo https://download.opensuse.org/repositories/openSUSE:Factory/standard/openSUSE:Factory.repo`. I have to clarify and ask them.

3. I do not know if there is a github action/workflow that pushes new releases to openSUSE Build Service (OBS). I will ask about that.

davidanthoff:
Great!

Lets forget about the `julia`/`juliaup` thing for now. We can revisit that once we flip the switch officially here on the Julia side and declare `juliaup` the default way to get `julia`.

Should we wait with merging this until you know whether we actually need the `zypper addrepo` instruction or not?

uncomfyhalomacro:
According to https://en.opensuse.org/Portal:Tumbleweed and https://www.suse.com/suse-defines/definition/opensuse-factory/
> Tumbleweed is based on Factory, openSUSE's main development codebase. Tumbleweed is updated once Factory's bleeding edge software has been integrated, stabilized and tested. Tumbleweed contains the latest stable applications and is ready and reliable for daily use.

> The openSUSE Factory project is the development codebase that creates openSUSE Tumbleweed, which is the rolling-release version of the openSUSE Project. The Factory receives a constant flow of packages from developers working on the openSUSE distribution.
>
>The Factory repository holds the raw, untested packages and they receive automated testing via openQA. Once the testing is done and the repository is deemed to be integrated, stabilized and consistent, it is synched to the download mirrors, and two or three times a week it is published as openSUSE Tumbleweed.
>
> The Tumbleweed version is primarily used by power users, software developers, and openSUSE contributors who need to work on a reliable platform that is as close to openSUSE Factory as possible. The Linux kernel is frequently updated in Tumbleweed, so users who rely on 3rd party kernel drivers often need to update all their drivers from source whenever they download a fresh version of Tumbleweed. Users who just need a reliable, stable working version of openSUSE and don’t want to worry about compiling kernel modules prefer to use the openSUSE Leap version of the distro, which is updated much less frequently.

I submitted this package from my `home` repo project -> request for submission in `science` repo -> accepted -> request for submission in `Factory` repo. And this undergoes openQA once in Factory. If it passes it will be synced to the download mirrors and published as openSUSE TumbleWeed.

So therefore doing `sudo zypper in juliaup` will install `juliaup` since it is already in the download mirrors. I can confirm this with

```sh
λ zypper info juliaup                                                                             
Loading repository data...
Reading installed packages...


Information for package juliaup:
--------------------------------
Repository     : Main Repository (OSS)
Name           : juliaup
Version        : 1.5.37-1.1
Arch           : x86_64
Vendor         : openSUSE
Installed Size : 6.3 MiB
Installed      : Yes
Status         : up-to-date
Source package : juliaup-1.5.37-1.1.src
Upstream URL   : https://github.com/JuliaLang/juliaup
Summary        : Julia installer and version multiplexer
Description    : 
    A cross-platform installer for the Julia programming language.

    The installer also bundles a full Julia version manager called juliaup.
    One can use juliaup to install specific Julia versions, it alerts users when new Julia versions are released and provides a convenient Julia release channel abstraction.
```

davidanthoff:
Awesome!

So now we just need to figure out what triggers an update of `juliaup` there, right? Does that require you manually updating things?

uncomfyhalomacro:
> Awesome!
> 
> So now we just need to figure out what triggers an update of `juliaup` there, right? Does that require you manually updating things?

As for the automated github actions/workflows, i havent received any info about that yet. I will keep looking though. So for now, it is manual updating. I am automatically assigned to maintain `juliaup` in the science repository since it got approved. I do use `juliaup` most of the time and keep it updated so that should not be something I should worry for now. 

Thanks! I will keep you posted when I got more information :tada:

PR #264, opened 2022-03-16
Make `winres` a Windows-only build dependency

devmotion:
`winres` is only used on Windows (and using it on other platforms is undefined behaviour: https://crates.io/crates/winres). I was playing around with the Fedora packaging system and this build dependency is problematic since (unsurprisingly) there's no Fedora package of the `winres` crate. Therefore this PR makes `winres` a Windows-only build dependency and replaces `cfg!` with a `cfg` directive in build.rs, as explained in the documentation (https://crates.io/crates/winres).

(BTW it seems one still has to patch Cargo.toml if one wants to create a Fedora package but one does not have to patch build.rs anymore: https://docs.fedoraproject.org/en-US/packaging-guidelines/Rust/#_nightly_features_and_dependencies_for_other_platforms)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,12 +45,14 @@ windows = { version = "0.29", features = ["alloc", "std", "Win32_Foundation", "W
 
 [build-dependencies]
 anyhow = "1"
-winres = "0.1"
 itertools = "0.10"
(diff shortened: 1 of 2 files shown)


davidanthoff:
Excellent, thanks!

PR #272, opened 2022-03-21
Handle Rosetta not installed case on MacOS M1

davidanthoff:


src/bin/juliainstaller.rs
```diff
@@ -188,6 +188,24 @@ pub fn main() -> Result<()> {
...
     println!("");
 
+    #[cfg(all(target_os="macos", target_arch="aarch64"))]
+    {
+        match std::process::Command::new("arch")
+                .args(["-x86_64", "/bin/bash", "-c", "arch"])
+                .output() {
+                    Ok(value) => {
+                        if String::from_utf8(value.stdout)? != "i386" {
+                            println!("It seems that you have not yet installed Rosetta, please install it with `softwareupdate --install-rosetta` before you try to install Julia.");
```
fingolfin, line 198: Why is this necessary, though? I think a comment explaining that (or even a message to the user doing so) would be helpful.

PR #281, opened 2022-03-30
Add 1.8.0-beta2 and 1.8.0-beta3

frankier:


davidanthoff:
Thanks! I pushed two more commits that a) remove 1.8.0-beta2, which was not released, and b) adds 1.6.6 which was also tagged today. I'll merge once CI passes.

PR #303, opened 2022-06-01
new rc

anandijain:


anandijain:
do you need to tag a release? `juliaup self update` and `juliaup list` don't give me new rc

PR #324, opened 2022-07-16
Add 1.8.0-rc3

pthariensflame:


davidanthoff:
I've added it on `master`, and also made the switch that the `aarch64` builds are now the default for Mac starting with 1.8.0-rc3. Thanks for the PR, though :)

PR #329, opened 2022-07-17
Change the default self-install location

davidanthoff:


davidanthoff:
I'm merging this because I want to build a version that I can test, but more reviews after the fact would be great!

PR #330, opened 2022-07-17
Use valid semver identifiers for available_versions

### LINKED ISSUE #140: Make the "Version" entries in `juliaup.json` be valid semver versions
Right now, the versions in `juliaup.json` are written like `"Version": "1.6.2+0~x64"`. I don't really understand the purpose behind the `~` stuff and further, this is not a valid semver version string. I think it would be better if they were valid semver versions because then they could be treated as `semver::Version` throughout the code base instead of as strings.

I started a bit with this here: https://github.com/KristofferC/juliaup/tree/kc/types

Not sure what to do about backwards compatibility. There is probably some good rusty way of versioning schemas like this.

### LINKED ISSUE #290: Windows: Tilde as arch separator causes firewall issues
Turns out that it's currently impossible to add a firewall rule for Julia installations via juliaup because of the  `~` in the dir name.
```
PS> netsh advfirewall firewall add rule name="Julia" dir=out action=allow program="%USERPROFILE%\.julia\juliaup\julia-1.7.2+0~x64\bin\julia.exe" enable=yes

The application name could not be resolved.
```
While I'm pretty sure this is a Windows bug, it would nevertheless be nice to change the convention used by juliaup.

davidanthoff:
Fixes #290 and starts to fix #140.

The strings we use to identify a specific version now are like `1.8.0-rc2+0.x64`, and that then translates into folder names on disc like `julia-1.8.0-rc2+0.x64`.

The general idea is that we supply two pieces of build metadata in the semver, separated by a `.`: 1) our own build number (always 0 at the moment, but this allows us to replace all shipped binaries if we need to) and 2) the platform string.

Channel names are unaffected by this, i.e. they still read like `release~x64` etc.

@pfitzseb and @KristofferC does that make sense? @KristofferC that still does not use the `Version` type in the code, but once we have changed the identifiers everywhere, we should be able to make that change also sometime down the road.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -80,51 +80,51 @@ fn produce_version_db() -> Result<JuliaupVersionDB> {
     for v in &original_available_versions {
         if target_os == "windows" && target_arch == "x86_64" {
             db.available_versions.insert(
-                format!("{}+0~x64", v),
+                format!("{}+0.x64", v),
                 JuliaupVersionDBVersion {url_path: format!("bin/winnt/x64/{}.{}/julia-{}-win64.tar.gz", v.major, v.minor, v)}
             );
             db.available_versions.insert(
-                format!("{}+0~x86", v),
+                format!("{}+0.x86", v),
                 JuliaupVersionDBVersion {url_path: format!("bin/winnt/x86/{}.{}/julia-{}-win32.tar.gz", v.major, v.minor, v)}
             );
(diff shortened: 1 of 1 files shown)


davidanthoff:
Same here, merging so that I can test it better, but reviews after the fact would be greatly appreciated.

Issue #321, opened 2022-07-12
Support older glibc versions (e.g. glibc 2.12)


DilumAluthge:
Right now, it looks like the distributed `juliaup` binaries for Linux do not support glibc 2.12.

To reproduce, first, run the following command:

```
docker run --rm -it centos:6
```

Now, inside the Docker container, run the following commands:

```
/lib64/libc.so.6
curl -fsSL https://install.julialang.org | sh
```

Here is the output that I get:

```
info: downloading installer
Warning: Not enforcing strong cipher suites for TLS, this is potentially less secure
Warning: Not enforcing TLS v1.2, this is potentially less secure
/tmp/tmp.sFLAXI1M2z/juliainstaller: /lib64/libc.so.6: version `GLIBC_2.15' not found (required by /tmp/tmp.sFLAXI1M2z/juliainstaller)
/tmp/tmp.sFLAXI1M2z/juliainstaller: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by /tmp/tmp.sFLAXI1M2z/juliainstaller)
/tmp/tmp.sFLAXI1M2z/juliainstaller: /lib64/libc.so.6: version `GLIBC_2.18' not found (required by /tmp/tmp.sFLAXI1M2z/juliainstaller)
```

evetion:
Would be great to have this 👍🏻

Wongboo:
Really a issue

tpmerrifield:
Also ran into this on Red Hat Enterprise version 7.9, GLIBC_2.18 not found.

DilumAluthge:
My understanding is that statically linking glibc is strongly discouraged.

Statically linking musl is an option. One challenge is the DNS resolver. For example, musl's DNS resolver does not support TCP, so if the result is too big to fit in a UDP packet, musl truncates it and returns only some of the results. I'm not sure if this would be a problem in practice, but it's at least a theoretical concern.

We already have the infrastructure in Buildkite for building Julia with support for old glibc. (The support varies based on architecture. For x86_64, we support glibc as old as 2.12.2.) So it's really just a matter of using our existing tooling to build the $ARCH-linux-gnu binaries for Juliaup.

LilithHafner:
glibc is tier 1, so I think this blocks release and advertisement on the downloads page of juliaup as the official Julia version manager.

StefanKarpinski:
[...]

Issue #320, opened 2022-07-07
Consider stripping binaries


darleybarreto:
Hi,

I tested this
```bash
cargo build --release --bins --features selfupdate
```

`strip` disabled:
juliainstaller - 7.8MB
julialauncher - 4MB
juliaup - 7.8MB

`strip` enabled:
juliainstaller - 4.4MB
julialauncher - 2MB
juliaup - 4.5MB

You can enable strip on release by adding `strip = true` to `profile.release`.

davidanthoff:
Would we still get the same error messages, though?

darleybarreto:
Oh, I missed that. So when panicking happens, you lose the stack trace. But if the errors are somewhat dealt with, you can still have nice messages.

ViralBShah:
This doesn't seem like a worthwhile saving.

PR #334, opened 2022-07-19
CI: For some jobs, use `ubuntu-18.04` instead of `ubuntu-latest`

DilumAluthge:
This could potentially help us support _slightly_ older glibc versions.

See also: #321

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -29,7 +29,7 @@ jobs:
             os: windows-latest
             features: windowsstore
           - target: i686-unknown-linux-gnu
-            os: ubuntu-18.04
+            os: ubuntu-latest
             features: selfupdate
           - target: aarch64-unknown-linux-gnu
             os: ubuntu-18.04
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,7 +32,7 @@ jobs:
             os: ubuntu-latest
             features: selfupdate
           - target: aarch64-unknown-linux-gnu
-            os: ubuntu-18.04
+            os: ubuntu-latest
             features: selfupdate
           - target: aarch64-apple-darwin
             os: macos-latest
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,39 +12,30 @@ on:
 jobs:
 
   build-juliaup:
-    runs-on: ${{ matrix.os }}-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        target: [
-          x86_64-pc-windows-msvc,
-          x86_64-apple-darwin,
-          x86_64-unknown-linux-gnu,
-          i686-pc-windows-msvc,
-          i686-unknown-linux-gnu, 
-          aarch64-unknown-linux-gnu,
-          aarch64-apple-darwin
-          ]
         include:
           - target: x86_64-pc-windows-msvc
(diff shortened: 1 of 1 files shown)


davidanthoff:
> either vendoring openssl or switching to rustls

We should be using rustls on Linux already in any case.

SaadiSave:
> We should be using rustls on Linux already in any case.

Then do we need binaries dynamically linked to glibc?

davidanthoff:
Probably not, I like your suggestion here very much. The more self-contained we can make `juliaup` and `julialauncher` the better.

SaadiSave:
Great, I'll test static musl binaries first thing in the morning tomorrow.

davidanthoff:
> Great, I'll test static musl binaries first thing in the morning tomorrow.

Awesome!

> I'll use the rustls PR for the test.

Oh, you can ignore that PR, `rustls` is used on `main` right now for Linux.

SaadiSave:
> Oh, you can ignore that PR, `rustls` is used on `main` right now for Linux.

Why don't I see it in `Cargo.toml` as a dependency then?

SaadiSave:
@davidanthoff, the musl binaries work! Completely statically linked. No dynamic libraries, not even libc.

davidanthoff:
Amazing!

The size for `juliaup` is perfectly fine, the thing we do need to worry about (a bit) is the size of `julialauncher`. I've tried to keep the size of that really as small as possible, given that that sits in the "load path" for every Julia invocation and we want it to start as quickly as possible. At some point I had `julialauncher` at around 400 KB on Windows, but I'm not actually sure whether that is still the case. Could you check how big that one is with the static musl build? And maybe we could also try and see whether the ideas in https://github.com/JuliaLang/juliaup/issues/320 would make the `julialauncher` smaller with the static musl?

I should say, though, that the benefits of having everything self-contained probably even trump any size considerations we might have :) Really fantastic to see your results here!

SaadiSave:
>the thing we do need to worry about (a bit) is the size of `julialauncher`

Then you won't be happy to hear that size has ballooned to 4 MB
```text
❯ du target/x86_64-unknown-linux-musl/release/julialauncher
4092	target/x86_64-unknown-linux-musl/release/julialauncher
```

Even with stripping the symbols, it's still 2 MB:
```text
❯ du target/x86_64-unknown-linux-musl/release/julialauncher   
2016	target/x86_64-unknown-linux-musl/release/julialauncher
```

How should I measure the performance impact?

SaadiSave:
@davidanthoff, I can safely say at this point that using musl will have no adverse effects.

davidanthoff:
> That's not true anymore

Yep, it is what I feared :) Probably actually due to me just using additional crates since I originally tested this.

@DilumAluthge does this all make sense to you as well?

@SaadiSave do you want to open a PR where this is added? Or would it be easier if I add this to the (admittedly very) cumbersome and convoluted workflow. I think with the info you provided me I could do this really quickly.

SaadiSave:
> @SaadiSave do you want to open a PR where this is added? Or would it be easier if I add this to the (admittedly very) cumbersome and convoluted workflow. I think with the info you provided me I could do this really quickly.

Okay, do it.

davidanthoff:
@DilumAluthge I think https://github.com/JuliaLang/juliaup/pull/370 takes care of this, so OK to close this? Do protest if you disagree :)

Issue #356, opened 2022-08-02
Juliainstaller Crash Report on `x86_64-apple-darwin`: `Failed to open juliaup config file`


DilumAluthge:
Log message:

```
Now installing Juliaup
Well, this is embarrassing.

Juliainstaller had a problem and crashed. To help us diagnose the problem you can send us a crash report.

We have generated a report file at "/var/folders/dg/7mtlrlqd6fz01jdb9djb33bm0000gp/T/report-7fe674d3-a33e-46db-8626-81b807edeedb.toml". Submit an issue or email with the subject of "Juliainstaller Crash Report" and include the report as an attachment.

- Homepage: https://github.com/JuliaLang/juliaup

We take privacy seriously, and do not perform any automated error collection. In order to improve the software, we rely on people to submit reports.

Thank you kindly!
```

Log file:

```
name = 'Juliainstaller'
operating_system = 'unix:OSX'
crate_version = '1.6.12'
explanation = '''
Panic occurred in file 'src/operations.rs' at line 559
'''
cause = '''
called `Result::unwrap()` on an `Err` value: Failed to open juliaup config file.

Caused by:
    No such file or directory (os error 2)'''
method = 'Panic'
backtrace = '''

   0: 0x105226cac - juliaup::command_config_modifypath::run_command_config_modifypath::h3eb12daf0cd2c1d2
   1: 0x105136a76 - juliainstaller::main::h7abd7c0c48f4b2f4
   2: 0x105120a36 - std::sys_common::backtrace::__rust_begin_short_backtrace::h5636139a57dca0d6
   3: 0x10513aa51 - _main'''
```

## Steps to reproduce

1. `rm -rf ~/.julia`
2. `curl -fsSL https://install.julialang.org | sh`
3. Select "Customize installation".
4. Answer the questions as follows:
    - "Enter the folder where you want to install Juliaup ›": press Enter to select the default value.
    - "Do you want to add the Julia binaries to your PATH by manipulating various shell startup scripts? (y/n) ›": press `n`.
    - "Do you want to add channel specific symlinks? (y/n) ›": press `n`.
    - "Enter minutes between check for new version at julia startup, use 0 to disable (1440) ›": type `0` and press Enter.
    - Enter minutes between check for new version by a background task, use 0 to disable (0) ›": type `0` and press Enter.
5. Select "Proceed with installation".

DilumAluthge:
[...]

PR #357, opened 2022-08-02
Improve a "failed to open file" error message by printing the file path that we failed to open

DilumAluthge:
This should help debug cases such as #356.


DilumAluthge:
Even though I was able to debug #356 without this PR, I do think this is a good addition to have. In general, I think that if we fail to open a file, we should include the filename as part of the error message.

davidanthoff (approved):

PR #362, opened 2022-08-05
Fix a bug during remove of PATH on Mac

### LINKED ISSUE #356: Juliainstaller Crash Report on `x86_64-apple-darwin`: `Failed to open juliaup config file`

davidanthoff:
Fixes #356.

The problem is that we used a list of files to modify on remove from path that included the ZHS file even if it doesn't exist. 

Including that file during add to path makes sense, even if it doesn't exist, because that is the default on Mac, but on remove we should only include it if it exists.

### DIFF
--- a/src/bin/juliainstaller.rs
+++ b/src/bin/juliainstaller.rs
@@ -223,7 +223,7 @@ pub fn main() -> Result<()> {
             .ok_or(anyhow!("Could not determine the path of the user home directory."))?
             .join(".julia")
             .join("juliaup"),
-        modifypath_files: find_shell_scripts_to_be_modified()?,
+        modifypath_files: find_shell_scripts_to_be_modified(true)?,
     };
 
     print_install_choices(&install_choices)?;
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -520,7 +520,7 @@ fn remove_path_from_specific_file(path: PathBuf) -> Result<()> {
     Ok(())
 }
 
-pub fn find_shell_scripts_to_be_modified() -> Result<Vec<PathBuf>> {
+pub fn find_shell_scripts_to_be_modified(add_case: bool) -> Result<Vec<PathBuf>> {
     let home_dir = dirs::home_dir().unwrap();
 
     let paths_to_test: Vec<PathBuf> = vec![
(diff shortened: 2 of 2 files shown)


DilumAluthge:
Thanks!

Issue #274, opened 2022-03-23
Support for firewall proxy?


kbarros:
My workplace requires all web traffic to go through their proxy. For example, I have configured the environment variable `$http_proxy`, and this setting is picked up by most command line utilities. Does juliaup has the capability to use this proxy setting for its connection to julialang.org? At the moment, I am seeing the error message "Connection refused". Thank you.

DhruvaSambrani:
I do not have RUST skills, but from a precursory reading of juliaup's code and some websearching, I think the line - 

https://github.com/JuliaLang/juliaup/blob/18ea73b8d95ac02349bfb206c1438b11d7cd5803/src/operations.rs#L72

should include a new Agent with proxy support as in [the official docs](https://docs.rs/ureq/latest/ureq/#example-using-http-connect)

```rust
// Configure an http connect proxy. Notice we could have used
// the http:// prefix here (it's optional).
let proxy = ureq::Proxy::new(env::var("https_proxy"))?;
let agent = ureq::AgentBuilder::new().proxy(proxy).build();

// This is proxied.
let resp = agent.get(url).call()
```

kbarros:
Adding to this, it looks like the [env_proxy](https://crates.io/crates/env_proxy) package could look up the appropriate environment variables according to a given `url` string.

kbarros:
Indeed, here is where they are using `env_proxy` to get the appropriate proxy address from the target url and the environment variables:
https://github.com/rust-lang/rustup/blob/bc18467b500c8cb892cec84316b23e96207553c4/download/src/lib.rs#L350

davidanthoff:
And maybe on Windows we should also use [WinHttpGetDefaultProxyConfiguration](https://docs.microsoft.com/en-us/windows/win32/api/winhttp/nf-winhttp-winhttpgetdefaultproxyconfiguration) or something similar to get the system default proxy config?

DhruvaSambrani:
As a temporary workaround, people may find `proxychains-ng` to be of great help. Install and RTFM.

davidanthoff:
[...]

PR #364, opened 2022-08-06
Wolthom/feature/use proxy

wolthom:
This PR uses the `env_proxy` crate (as suggested in #274) to respect environment variables setting a proxy.
Typically, corporate laptops may require this feature (such as my work laptop).

Because wrongly configured proxies may lead to weird errors (e.g. DNS resolution errors or refused connections) with suboptimal error messages, I added logging traces to the proxy-resolution in case debugging is required. The logging statements are in a separate commit and can be removed if preferred.

Right now, this is only tested on an Ubuntu notebook. Later today, I will verify that this settings works on a Windows behind a corporate firewall. With the current release, `juliaup` does not work on said laptop.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -398,6 +398,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "env_proxy"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3a5019be18538406a43b5419a5501461f0c8b49ea7dfda0cfc32f4e51fc44be1"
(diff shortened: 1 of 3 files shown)


wolthom:
Small upate:
The changes seem to fix the proxy issues in my specific work environment. If other people could verify the functionality in different environments (with / without proxy), this would be great.

davidanthoff:
This looks great! I'm going to merge right now, it seems to me that this is an improvement no matter what, even if it were to not work in all circumstances.

Issue #363, opened 2022-08-05
Find alternatives to Mozilla cert list on Linux


davidanthoff:
I think this should probably be reverted once the `pemfile` dependency is upgraded to include https://github.com/rustls/pemfile/pull/7. Yes, the Mozilla certs a known good, so why not use them? Because many users are behind MITM proxying firewalls and the connection will fail using only public certs. On such systems, a sysadmin will typically have added the CA root cert for the firewall to the system so that secure connections can be made. By using Mozilla certs, that will stop working.

_Originally posted by @StefanKarpinski in https://github.com/JuliaLang/juliaup/issues/335#issuecomment-1206779507_

davidanthoff:
@StefanKarpinski and I just had another chat, and for now the strategy would be to a) use system certs, but b) detect Debian and on Debian don't use them by default, but have users opt-in to using system certs via some env var.

davidanthoff:
https://crates.io/crates/os_info might be useful (thanks to @wolthom for finding it!).

StefanKarpinski:
Here's the logic we use in NetworkOptions:

https://github.com/JuliaLang/NetworkOptions.jl/blob/791fa05928f03b7f18ff5/src/ca_roots.jl#L85-L110

I'm not sure if juliaup uses OpenSSL on Linux; if it does, it can probably skip the logic in there that skips over pem files that only contain OpenSSL-specific `BEGIN TRUSTED CERTIFICATE` blocks. The list of places where we look is here:

https://github.com/JuliaLang/NetworkOptions.jl/blob/791fa05928f03b7f18ff5/src/ca_roots.jl#L58-L68

I based this off of a combination of what Rust and Go do when looking at these files and it seems to work pretty well (we needed to do some reordering at some point). If you wanted, we could put this list somewhere common that you could download it during the rustup build process and likewise NetworkOptions could grab it from there. Otherwise we could just try to keep the lists in sync—they shouldn't have to change often. A comment saying to change the other if one of theses changes might suffice.

StefanKarpinski:
Looking at this issue in an attempt to close out the public release for Linux & MacOS milestone.

wolthom:
@StefanKarpinski @davidanthoff
I'm a bit out of depth regarding the certificate roots / native certificate stores vs bundled ones. If the points below do not actually address this issue, feel free to skip over this comment.

After searching around in the `ureq` and `rustls` docs for a bit, I wonder if this could be an option:
[...]

PR #365, opened 2022-08-06
Use HttpClient for downloads on Windows

davidanthoff:
This completely drops the ureq dependency on Windows and instead uses https://docs.microsoft.com/en-us/uwp/api/windows.web.http.httpclient for all HTTP/download stuff.

The benefit of this is that we are getting all the default certificate and proxy handling that Windows has.

The downside is that this is Windows 10 and newer only. It might make sense to only use this in the Windows Store builds (they are Win 10 and newer by definition) and for the standalone installer use the previousl ureq download story.


davidanthoff:
@StefanKarpinski you spent some quality time on figuring out the download story for Julia base lately, right? Does this here broadly make sense?

davidanthoff:
Also, note to self: merge https://github.com/JuliaLang/juliaup/pull/364 first.

davidanthoff:
@wolthom Is there a chance that you could also try this PR on your Windows machine? The idea here is that it doesn't rely on more Unixy env variables, but instead (hopefully) just uses the standard Windows system proxy configuration.

wolthom:
@davidanthoff Sure. Just build the most recent commit to the 'win-httpclient' branch and see what happens?

As a warning though:
The corporate setup I'm in can be quite infuriating at times (this is me being diplomatic). In case things don't work there, it could still be a working solution for most people.

davidanthoff:
> Just build the most recent commit to the 'win-httpclient' branch and see what happens?

Yes, and then just try to do some `juliaup add 1.0.2` or something for a version that you definitely don't have on your machine :)

> The corporate setup I'm in can be quite infuriating at times

The general idea with this PR here is that it should pick up any Windows system configuration. My hope would be that _if_ something like Microsoft Edge works on your machine, then this here should also work without any further configuration.

wolthom:
@davidanthoff  Quick update:
This approach seems to work as well :+1: 
I was able to install and launch a Julia version with the local build results that I definitely did not have on the system before.

Also, I got a compilation error because `ureq` is not a dependency for Windows machines, but the `get_proxy` function in `operations.rs` is always compiled => A `#[cfg(not(windows))]` should fix that issue, though.

This this part of the code is currently being touched:
What is / was the reason to treat MacOS and other Unix target differently regarding `native-tls` in ureq? Could this be deduplicated?

davidanthoff:
Ah, thanks, missed that conditional compile when I merge `main` :) Hopefully fixed now.

> What is / was the reason to treat MacOS and other Unix target differently regarding native-tls in ureq? Could this be deduplicated?

I wanted to get the native certificate story in before Juliacon, but then I found some info that made me wonder whether that is safe on Unix (essentially apparently Debian has outdated certificate lists by default) and then I ran out of time to really sort it out and thought that just sticking with the Mozialla cert list on Linux was the safe thing to do for now. The current plan is to change the Linux story here to bring it more in line with the Mac story, details at https://github.com/JuliaLang/juliaup/issues/363.

wolthom:
Understood, thanks for the link! Not an expert on that topic, otherwise I'd love to help out.

Also, thank you for the fantastic work on `juliaup` (and associated tools). That in combination with `Pkg` and the `Artifacts` / `BinaryBuilder` ecosystem is becoming pretty amazing for simple, reliable system setups :)

Regarding the linked issue:
https://crates.io/crates/os_info => This crate may come in handy (in case you're not already using it anyway).

davidanthoff:
Ok, I'm merging this.

I think this is essentially the best way to do downloads on Windows, in particular it seems pretty much the only way to correctly pick up system proxy configurations. Even libcurl still has that on the backlog :)

I think that probably means we will have a more robust (or at least less configuration required) download story in Juliaup than in Downloads.jl and Pkg.jl on Windows. I'm not sure how useful that is, it probably means we will end up with folks that install Julia versions but not packages without further environment variable configuration, but presumably the right move would be that at some point Downloads.jl also picks up system proxy configuration on Windows, rather than relying on env vars.

PR #370, opened 2022-08-06
Link crt statically on Linux

davidanthoff:


DilumAluthge:
Note that regardless of what platform we build these on, we should be testing them on `x86-linux-gnu`, because we need to make sure that our binaries work on systems that have glibc on the system (and do not have musl on the system).

It's fine to test on `x86_64-linux-musl`, as long as we also test on `x86_64-linux-gnu`.

davidanthoff:
Right now I'm under the impression that we can either link glibc or musl statically. If that holds for all the platforms, which one should we chose? glibc somehow seems the safer option to me, or not?

davidanthoff:
The original suggestion for all of this from @SaadiSave had the following additional arguments:
```
-Clto=fat -Cembed-bitcode=yes
```
@SaadiSave should we use those? It is not clear to me what they do :)

The other open question is: are we ok to just statically link glibc here? Would there be any benefit in using musl instead? @DilumAluthge any thoughts on that?

DilumAluthge:
I'm far from an expert here, but it was my understand that glibc cannot be linked statically without breaking some things.

SaadiSave:
> The original suggestion for all of this from @SaadiSave had the following additional arguments:
> ```
> -Clto=fat -Cembed-bitcode=yes
> ```
> @SaadiSave should we use those? It is not clear to me what they do :)

lto is link-time-optimization
embed-bitcode is required for lto to work

It's not really required, that's just what I use when building locally. But it might provide a speed boost because the compiler will be able to inline more and eliminate more dead code and unused branches. Upto you whether you want to use it.

> The other open question is: are we ok to just statically link glibc here? Would there be any benefit in using musl instead? @DilumAluthge any thoughts on that?

My original suggestion had the target `x86_64-unknown-linux-**musl**`. Static linking doesn't work well with glibc, while musl is designed for static linking.

davidanthoff:
Merged a different version of this on `main`.

PR #373, opened 2022-08-09
Reduce image file sizes (losslessly)

pthariensflame:


davidanthoff:
Hm, so on one hand this is neat (how did you actually pull this off), but on the other hand I'm reluctant to replace these pictures because I think it would trigger some further evaluations in the store submission process, and the less we rock that boat the better ;) Especially in this case, where I don't think the current size of these images is really an issue.

pthariensflame:
Understood!  I basically made this as a drive-by contribution and don't mind if it's not acceptable.  As for the how:  https://imageoptim.com/ (I used the app version).

Issue #131, opened 2021-08-18
Use the published `versions.json` instead of listing versions in source code?


KristofferC:
Instead of hardcoding version information in the build file:

https://github.com/JuliaLang/juliaup/blob/623684d248c50a34e912ff0223bdfd154df4f0d5/build.rs#L23-L52

it perhaps makes sense to use the published https://julialang-s3.julialang.org/bin/versions.json file (that among other things is used for CI). That would decouple the version information from the juliaup executable which feels like a good idea.

There is, however, no way to determine which is the latest LTS from that file AFAIU. That could be hardcoded to be `1.0.x` for now (where the latest such version from the file is used).

PallHaraldsson:
>no way to determine which is the latest LTS from that file AFAIU. That could be hardcoded to be 1.0.x for now

That's of course no longer the case, so you want to have Julia 1.6.4 for LTS, that fixes a security issue. It's a question if you want to warn (or block?) against any older version.

davidanthoff:
I looked into this in a bit more detail, and at the moment the official `versions.json` is missing a fair number of specific archives that we need. I didn't look through everything, but minimally this:
- For Mac, it only lists the `.dmg` files, not the `.tar.gz` files
- For Windows, it only lists the `.zip` files, not the `.tar.gz` version

It is not clear to me how this could be fixed without breaking lots of stuff. Presumably a lot of existing downstream code relies on the fact that there is only ever one entry in the `files` list as `archive` for a given `triplet` value, so just adding another entry to `files` is probably not a good idea. Also, even if that wouldn't break lots of things, one would then have to look at file extensions in the URL to figure out which entry uses what kind of archive, which also seems like a bad design.

I guess maybe one option would be to allow a value of `targzarchive` in the `kind` field, and that add lots of new entries with that? But that of course also has the potential to break downstream stuff...

CCing a random list of folks that might have input on this @staticfloat, @SaschaMann, @ararslan

SaschaMann:
I wasn't even aware those existed. The main thing we can do without breaking downstream stuff is adding keys, so my initial thought was that we could add an `alternates: []` key to the archives and perhaps list the others in there? Though I have to say that feels like a bodge.

davidanthoff:
Another option would be that we ignore everything in `files`, and just use the list of versions that we can get from `versions.json`, and continue to construct download URLs ourselves... That might actually be the easiest short-term solution.

uncomfyhalomacro:
[...]

PR #417, opened 2022-10-03
 Use the published versions.json

xgdgsc:
fix https://github.com/JuliaLang/juliaup/issues/131 

Basically just uses the version from json file on runtime instead of build time. LTS version is still hard-coded. Just adding a LTS version file on this repo or use from somewhere like https://github.com/Homebrew/homebrew-cask-versions/blob/master/Casks/julia-lts.rb might work.

Tested on linux with musl and Windows 11 22H2, with and without JULIAUP_SERVER env (works with mirror in China).


StefanKarpinski:
What needs to happen to move this forward?

wolthom (commented):
I added some general remarks to this PR.
Right now, one thing that could be improved is the handling of conditionals regarding the versions/channels/architectures. Especially localized match statements can make this a bit cleaner.

Additionally, some unit tests would be nice (either for the os/arch handling in particular, or just verifying some properties of the final DB).

wolthom:
One more thought:
Would it make sense to cache the versions.json, or is it explicitly desired to download it every time `juliaup` is used to manage versions? This would also make it more robust to usage while offline. Currently, using `juliaup list` would fail when offline (arguably, not failing would not provide any major benefits).

SaschaMann:
Caching might make sense if there's a check to see if it has changed recently. Otherwise I feel like people would be annoyed when a new release has been announced and they want to try it out ASAP. Maybe checking `last-modified` first would be a workaround.

src/versions_file.rs
```diff
@@ -1,32 +1,982 @@
...
+use std::env;
+use ureq;
 
-pub fn load_versions_db() -> Result<JuliaupVersionDB> {    
-    let vendored_db = include_str!(concat!(env!("OUT_DIR"), "/versionsdb.json"));
+pub type JsonVersion = HashMap<String, JsonVersionValue>;
 
+#[derive(Serialize, Deserialize)]
+pub struct JsonVersionValue {
+    #[serde(rename = "files")]
```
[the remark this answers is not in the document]
xgdgsc: This is generated with quicktype. So not sure whether to add that.

src/versions_file.rs
```diff
@@ -1,32 +1,982 @@
...
+    }
+    let mut db = JuliaupVersionDB {
+        available_versions: HashMap::new(),
+        available_channels: HashMap::new(),
+    };
+    let target_os = env::consts::OS;
+    let target_arch = env::consts::ARCH;
+
+    for v in &original_available_versions {
+        if target_os == "windows" && target_arch == "x86_64" {
```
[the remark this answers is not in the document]
xgdgsc: Did a partial switch. Do you think the approach is ok to follow in the next if else blocks?

src/versions_file.rs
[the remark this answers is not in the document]
wolthom: Gotcha. From my side it would be fine to leave it as-is for now.

src/versions_file.rs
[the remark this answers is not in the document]
wolthom: I like the changes! Personally I find this cleaner to read and the compiler will remind you of the fall-back branch (as opposed to if-else chains).

One more improvement (clippy kept bugging me):
In the `channel_versions` function, you can directly return the result of the `match`-expression. I.e.:
```rust
fn channel_versions(...) -> ... {
match (...) {
    ...
}
}
```

I forgot about that simplification in the Rust-playground snippet.

wolthom:
@SaschaMann I'd suggest to implement the caching in a separate PR. This works reasonably well for now and then we can get some initial experiences while implementing the caching.

@davidanthoff Are there any recent changes that should be tagged/released before merging this larger change eventually?

xgdgsc:
Added a test of listing urls that' s not in the online json file:

```
running 1 test
https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5.0-rc1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4.7-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4.0-rc4-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.7/julia-0.7.0-alpha-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7.0-beta2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.1/julia-1.1.0-rc1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.3/julia-1.3.0-rc5-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6.4-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5.0-rc4-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.1/julia-1.1.0-rc2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.3-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.0/julia-1.0.0-rc1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.5-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6.3-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.12-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4.0-rc2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.7-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5.0-rc4-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.1/julia-1.1.0-rc1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5.1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.7/julia-0.7.0-beta2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.3/julia-1.3.0-rc2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.3/julia-1.3.0-rc4-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6.0-pre.beta-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.3-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5.0-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.12-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5.0-rc0-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.7-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4.4-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.11-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4.0-rc1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4.4-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.4/julia-1.4.0-rc1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4.0-rc3-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4.2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5.0-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6.0-rc1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5.0-rc0-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.3/julia-1.3.0-rc1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.10-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4.5-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4.2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6.0-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.8-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5.2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.4/julia-1.4.0-rc2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6.0-pre.alpha-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5.0-rc1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.3/julia-1.3.0-rc2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5.1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.3/julia-1.3.0-rc3-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5.0-rc3-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4.3-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.1/julia-1.1.0-rc2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.7/julia-0.7.0-beta-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6.0-rc2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.0/julia-1.0.0-rc1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.3/julia-1.3.0-rc3-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.3/julia-1.3.0-alpha-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6.2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4.0-rc4-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6.0-rc3-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4.1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.2/julia-0.2.0-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.7/julia-0.7.0-rc1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.2/julia-0.2.0-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6.0-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.6-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4.7-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.0-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4.0-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.2/julia-0.2.1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.0-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7.0-rc2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.11-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7.0-rc1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7.0-rc3-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4.0-rc3-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6.0-rc2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.2/julia-1.2.0-rc3-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.8-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.4/julia-1.4.0-rc1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.2/julia-0.2.1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5.0-rc2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4.0-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.5-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4.5-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6.2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.1/julia-0.1.2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.9-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6.0-rc3-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.2/julia-1.2.0-rc2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6.4-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5.0-rc3-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.2/julia-1.2.0-rc1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.2/julia-1.2.0-rc2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6.1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.7/julia-0.7.0-rc3-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.6-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4.3-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.2/julia-1.2.0-rc1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.4-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7.0-alpha-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.3/julia-1.3.0-rc5-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.10-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6.1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6.0-rc1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.4/julia-1.4.0-rc2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6.3-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4.1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4.6-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.4/julia-0.4.6-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.3/julia-0.3.4-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.1/julia-0.1.2-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4.0-rc2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.3/julia-0.3.9-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5.2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.3/julia-1.3.0-alpha-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6.0-pre.beta-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.4/julia-0.4.0-rc1-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.3/julia-1.3.0-rc4-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.7/julia-0.7.0-rc2-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6.0-pre.alpha-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/0.7/julia-0.7.0-beta-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x86/1.3/julia-1.3.0-rc1-win32.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/1.2/julia-1.2.0-rc3-win64.tar.gz
https://julialang-s3.julialang.org/bin/winnt/x64/0.5/julia-0.5.0-rc2-win64.tar.gz
```

xgdgsc:
Removed online check of julialauncher.

tests/command_list_test.rs
```diff
@@ -1,15 +1,52 @@
...
+    }
+
+    let vdb = load_versions_db().unwrap();
+    for (_, v) in vdb.available_versions {
+        let full_url = format!(
+            "https://julialang-s3.julialang.org/{}",
+            v.url_path.to_string()
+        );
+        if !exist_urls.contains_key(&full_url) {
+            println!("{}", &full_url);
```
wolthom, line 48: Could you add a comment or two regarding what you are testing here?

Also, unless you instruct someone to run the tests with `--nocapture`, this `println!`s will not show up.
Is this intended?

src/bin/julialauncher.rs
```diff
@@ -134,12 +105,6 @@ fn get_julia_path_from_channel(
                 .installed_versions.get(version)
                 .ok_or_else(|| anyhow!("The juliaup configuration is in an inconsistent state, the channel {} is pointing to Julia version {}, which is not installed.", channel, version))?.path;
 
-            check_channel_uptodate(channel, version, versions_db).with_context(|| {
```
wolthom, line 137: Could you briefly explain, why you removed the online check from the `julialauncher` binary?

tests/command_list_test.rs
xgdgsc: I initially thought every url should be in the versions.json. But as it is not the case, I just added the code to show the diff and wait for your opinions.

src/bin/julialauncher.rs
xgdgsc: Because If you use julialauncher offline or in bad/slow network condition. It would block the running or add a noticable delay. So removing online checking of julialauncher seems more reasonable. Make `juliaup` the explicit update manager. And `julialauncher` the simple launcher.

xgdgsc:
Tried a windows on arm64 build. Adding last commit make it work. Will add the workaround needed to build arm64 on `cargo.toml` later if needed.

tests/command_list_test.rs
wolthom: Hmm. Interesting point. I don't the reason behind this. Based on a quick screening, it does not look like any actual post-1.0-versions are missing. All of the entries seem to be RCs and alpha versions.
I'd guess that it's unlikely that people will require these versions, if the actually released versions seem to be available.

About the pre-1.0 versions I'm not sure. @davidanthoff Do you have any ideas on this? Are the missing Windows versions going to be an issue?

StefanKarpinski:
Just did a little DMing with @wolthom about caching the `versions.json` file. It should be pretty straightforward: just download the file to a location and re-download periodically. The main consideration is just making sure that races between multiple `juliaup` instances don't corrupt or confuse things. For that, I'd use two strategies:

1. File locking (advisory). Rust seems to have good support for this (and I suspect it's already being used for other things by `juliaup`).
2. Making file replacement as atomic as possible: download to a temporary location that's on the same file system and then replacing the old copy by doing an atomic `mv` operation.

Each of these is mostly reliable but not 100% but together they should work on all but the most broken os/file system combinations (some distributed file systems might not implement advisory locks or atomic move).

DilumAluthge:
> Just did a little DMing with @wolthom about caching the `versions.json` file. It should be pretty straightforward: just download the file to a location and re-download periodically.

@StefanKarpinski @wolthom Can we have a mechanism for the user to force the latest `versions.json` file to be downloaded? Otherwise, I can see a scenario where a new Julia release has just been announced, and the user is trying to install the new version, but Juliaup chooses not to re-download `versions.json` because it was downloaded relatively recently.

Maybe something like `juliaup refresh` or something?

cc @davidanthoff to weight in on the command syntax.

wolthom:
@DilumAluthge  Personally, I'd prefer to add this as a flag to the `juliaup list` command. E.g. `juliaup list --online`, `juliaup list --refresh`, or something along those lines.
The caveat being that this option may end up harder to find. But that can be addressed with documentation and an improved help message of the `juliaup list` command.

xgdgsc:
What about when using `julialauncher` it would trigger a background periodic update of  cache file (via call to `juliaup`) which doesn' t affect launching latency. And when using `juliaup` every time the cache is updated, so there' s no need to add any command to juliaup.

wolthom:
For now I'd really focus on the caching and get that implemented reliably. When that works we can still discuss when / where the availability of updates should be checked.

wolthom:
@xgdgsc Just wanted to check whether you need more input on caching the `versions.json` / intend to add that functionality to this PR.

xgdgsc:
No. I' m fine with the current status.

davidanthoff:
Sorry that I'm only chiming in now, I should have engaged much earlier, but was (and still am) busy with other things at the moment and was not able to keep up with Julia stuff at all.

I have been working on supporting this scenario since the springish (nothing pushed yet) and generally have a design and parts (really just a little bit) of an implementation that is quite different from this here. The primary differences are:
- the client doesn't use `versions.json` directly, instead we host separate versions specifically for juliaup that get auto-generated from `versions.json`. Many benefits of this: a) we can have smaller files that only contain the info we need (important because we'll have to parse the version file at each `julia` startup), b) we can have additional info in there (like LTS info), c) we can handle schema updates _much_ more elegantly (this is actually really crucial, because we must make sure that even a completely outdated `juliaup` version that hangs around on some random machine somewhere still works in five years), d) it integrates much better with the idea of the dev, release-preview and release channel that we have for juliaup, e) it has a mechanism to only download a new version db if there have been updates.
- It handles all the concurrency problems when updating the locally cached version. This is another hard requirement (an earlier version of Juliaup actually _did_ already download a version.json, but it didn't handle the concurrency properly and it became very clear very quickly that this didn't work), and the infrastructure for this is all in place already because there is a locking story for updates to the config database in place already that we can just re-use for this.
- I only skimmed the discussion about online/offline here, but it essentially handles that scenario properly, i.e. the locally cached version auto-updates in the background.
- It properly integrates with the first-install scenario, i.e. we must make sure that when things are deployed via the Windows Store, that one can run `julia` even when one is offline, i.e. when there is currently _no_ `versions.json` around it should fall back to the info that came in the MXIS package.

I am not entirely sure when I'll be able to finish this, my hope is maybe this calendar year. The whole thing is pretty involved because it also involves a fair number of Github Workflows that auto-do things...

I know folks are eager to get rid of the manual stuff (and I am as well), but on the other hand for now it is actually not very cumbersome to do things by hand, and I have (finally) written down how to do that, so that others could help with that as well, see https://github.com/JuliaLang/juliaup/wiki/Adding-a-Julia-version and https://github.com/JuliaLang/juliaup/wiki/Release-process.

KristofferC:
> but on the other hand for now it is actually not very cumbersome to do things by hand,

Pretty much every release there are complaints about juliaup doesn't have the latest version. I would say getting something like this in should be highest priority. If this PR works OK in practice I would favor merging it and if someone has better implementations juliaup can always be updated to that since it is just changes to internals.

DilumAluthge:
I think it's fine if we can just get more people that are able to do the release process quickly after a new Julia release. Now that the process is documented, it's just a matter of adding more people to the list of approved reviewers. I've added @staticfloat, @IanButterworth, @ararslan, and @KristofferC  as reviewers (in addition to @davidanthoff). I've also created a team (the ` @JuliaLang/juliaup-release-approvers` team), so it's easy to add additional reviewers just by adding them to the team.

davidanthoff:
I'm going to close this, I'm almost done with my implementation. Much of it is already on `main` now, I'm pretty sure I can finish it in a few days the most.

@xgdgsc I'm sorry I didn't tell you earlier that I had parts of an implementation already and that I would finish that rather than go with a different PR. But thanks so much for engaging here, and I hope you might want to help with future features!

Issue #406, opened 2022-09-18
1.8.1 not recognised as the latest version?


ghyatzo:
Hello,

I am encountering this message everytime i start julia
```
The latest version of Julia in the `release~aarch64` channel is 1.7.2+0~aarch64. You currently have `1.8.1+0.aarch64` installed. Run:

  juliaup update

to install Julia 1.7.2+0~aarch64 and update the `release~aarch64` channel to that version.
```
I have installed the `release`, `release~aarch64` and `release~x64` versions, and a similar message pops up with all of them.
Of course i ran `juliaup update` and I am at the latest version.

The installed julia versions run flawlessly though, so the functionality is not affected.

I am on an M1 mac with MacOS 12.5.1

albheim:
What version of juliaup are you running, `juliaup --version`? Have you tried `juliaup self update`?

Don't really know what I'm talking about though, but my thought was that the `self update` is to update juliaup itself, and if you have an old juliaup maybe it has old definitions of the channels? Unclear how you managed to get the newer versions to begin with though.

ghyatzo:
juliaup version is `1.7.23`.
`juliaup self update` return 
```
Checking for self-updates
Juliaup unchanged on channel 'release' - 1.7.23
```

Not really sure what you mean by this though

> Unclear how you managed to get the newer versions to begin with though.

albheim:
Okay, no clue then.

> Unclear how you managed to get the newer versions to begin with though.

was just in case
> if you have an old juliaup maybe it has old definitions of the channels

was true, but since it seems it wasn't it can be disregarded.

ghyatzo:
After checking the code, and try to guess where it was taking bad informations, I suspected that I had some issues with the installation folders.
I tried to uninstall and reinstall and now it works.

by the way, the uninstalling process is waaay to aggressive. I know I had some misconfigured folders, but still, it deleted a lot of folders it wasn't supposed to. Without asking anything or doublechecking. It cost me an afternoon to reinstall everything the way it was. Please fix this.
Also, maybe don't delete the `config` folder in `$JULIA_DEPOT_PATH`, it would be nice to keep the `startup.jl` file between reinstalls.

shall I close this issue and open another one with these comments?

albheim:
I just tried the uninstall, and it seemed reasonable to me, the config was still there and so was any other julia related things I wouldn't expect juliaup to remove. How did you configure your julia?

[...]

PR #429, opened 2022-10-31
Manage the installation folder path through env variable.

ghyatzo:
This started from #406. 
While the install location of the various julia versions can be customised through the `JULIA_DEPOT_PATH` env var, the self installation folder was not, and instead the program relied on the current position of the running executable (or symlink to it).
This PR introduces the env var `JULIAUP_SELF_HOME` which controls the paths to the juliaup binaries and selfconfiguration files. 

To avoid too much book keeping to manage an eventual custom location, I disabled the logic for letting the user type the location itself, and instead only rely on setting up the right env variable, similarly to how rustup does.

### DIFF
--- a/src/bin/juliainstaller.rs
+++ b/src/bin/juliainstaller.rs
@@ -12,17 +12,17 @@ fn run_individual_config_wizard(install_choices: &mut InstallChoices, theme: &di
 
     trace!("install_location pre inside the prompt function {:?}", install_choices.install_location);
 
-    let new_install_location = Input::with_theme(theme)
-        .with_prompt("Enter the folder where you want to install Juliaup")        
-        .validate_with(|input: &String| match input.parse::<PathBuf>() {
-                Ok(_) => Ok(()),
-                Err(_) => Err("Not a valid input".to_owned())
-        })
-        .with_initial_text(install_choices.install_location.to_string_lossy().clone())        
-        .interact_text()?;
+    // let new_install_location = Input::with_theme(theme)
+    //     .with_prompt("Enter the folder where you want to install Juliaup")        
+    //     .validate_with(|input: &String| match input.parse::<PathBuf>() {
+    //             Ok(_) => Ok(()),
+    //             Err(_) => Err("Not a valid input".to_owned())
+    //     })
+    //     .with_initial_text(install_choices.install_location.to_string_lossy().clone())        
+    //     .interact_text()?;
 
-    let new_install_location = shellexpand::tilde(&new_install_location)
-        .parse::<PathBuf>().unwrap();
+    // let new_install_location = shellexpand::tilde(&new_install_location)
+    //     .parse::<PathBuf>().unwrap();
 
     let new_modifypath = match Confirm::with_theme(theme)
         .with_prompt("Do you want to add the Julia binaries to your PATH by manipulating various shell startup scripts?")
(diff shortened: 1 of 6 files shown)


davidanthoff:
I left some comments at the end of #406, which maybe are easier and more robust ways to solve this.

I think we should generally try to not introduce another env variable, at the end of the day that is another state that is stored somewhere that can get out of sync with the rest of the state and then we have to handle all the potential errors that can come about from that.

ghyatzo:
Regarding the book keeping part I agree, that is why I also proposed to modify the installation script to only be controlled by the env variable (similar to how rustup handles things). That way you don't have to keep track if the user selects a different folder and juggle a possible dual state, where your env variable points somewhere but the internal state chosen by the user points somewhere else. You basically keep part of the state (the self referential one) outside the program.

If a user has the need to specify a custom installation folder she can do so only through the env var, and It is mentioned during the installation script where you are presented the current installation configuration.

davidanthoff:
I think the original problem is solved by https://github.com/JuliaLang/juliaup/pull/454, and beyond that I don't really see a benefit of introducing another state here via the env variable, so I'm tempted to close this, unless @ghyatzo thinks there are other benefits of this PR.

ghyatzo:
Late reply, differente time zone here.

The benefits of not relying on the `current_exe()` reference is that it enables running the juliaup program from anywhere within the filesystem (ie symlinking). Otherwise you are implicitly assuming that whichever folder the program is running from is the self home folder which is wrong. #406 as an example. 

The change you implemented don't solve #406, since the issue is the reference to `paths.juliaupselfhome` which calls ` let my_own_path = std::env::current_exe()` which reference the wrong folder if running juliaup from a symlink and not `, In that case you would be deleting non existing folders.

davidanthoff:
Ah, I had assumed that `current_exe()` would always return the path of the binary, even if the binary was launched via a symlink. Is that not the case?

ghyatzo:
Apparently it is [platform dependant](https://doc.rust-lang.org/std/env/fn.current_exe.html) also [here](https://github.com/rust-lang/rust/issues/43617). I am on Mac so in my case it doesn't follow the symlinks.

davidanthoff:
So maybe the right strategy would be to a) check whether the path returned by `current_exe` is a symbolic link, b) if it is, dereference it to find the target of the link, and then c) delete things relative to that path?

Or maybe there is just something similar to `current_exe` that always returns the actual path of the binary, not the path of the symbolic link?

ghyatzo:
I found the following example to be working:
`test.rs`
```
fn main() {
    let mypath = std::env::current_exe();
    let _actual_path = match std::fs::read_link(mypath.as_ref().unwrap()) {

        Ok(link) => {
            println!("path is a link pointing to {:?}", link)
        }
        Err(_) => {
            println!("path points to real exec {:?}", mypath)
        }
    };
}
```
and with
```
> rustc test.rs
> ln test hard_link
> ln -s test soft_link
> ./test 
path points to real exec Ok(".../test")
> ./hard_link
path points to real exec Ok(".../hard_link")
> ./soft_link
path is a link pointing to "test"
```

The only annoying thing is that `read_link` does not necessarily return an absolute path, but most likely a relative one.
The relative path is relative to the symlink, so we can use `current_exe()` to obtain the path of the symlink and join it with the relative path from the symlink contents to reconstruct the real path.

ghyatzo:
Even better, I found [this](https://doc.rust-lang.org/std/fs/fn.canonicalize.html) which resolves symlinks.

PR #475, opened 2022-11-25
In the "Update Version DB" PRs, add a note that the user must close a nd reopen the PR to trigger CI

DilumAluthge:


davidanthoff:
We probably don't need to run CI on these PRs, they are really just a data file update. But still, the message can't hurt.

PR #483, opened 2022-11-26
Enable ci tests of musl build

davidanthoff:
For some reason this breaks, if someone wants to help getting these tests pass, please :)


evetion:
> For some reason this breaks, if someone wants to help getting these tests pass, please :)
Might be fixed by #486

PR #486, opened 2022-11-26
Fix musl build can't find gcc-musl

evetion:
See https://github.com/actions-rs/toolchain/issues/102.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -88,6 +88,8 @@ jobs:
             rustflags:
     steps:
     - uses: actions/checkout@v3
+    - if: ${{ contains(matrix.target, '-musl') }}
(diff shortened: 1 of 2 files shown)


davidanthoff:
Ah, fantastic! Could you also add that to the `release.yml` test block https://github.com/JuliaLang/juliaup/blob/main/.github/workflows/release.yml#L120?

evetion:
> Ah, fantastic! Could you also add that to the `release.yml` test block [https://github.com/JuliaLang/juliaup/blob/main/.github/workflows/release.yml#L120](https://github.com/JuliaLang/juliaup/blob/main/.github/workflows/release.yml?rgh-link-date=2022-11-26T17%3A10%3A28Z#L120)?

Done, for both the build as the test command, to be sure.

PR #495, opened 2022-11-28
add win arm64 build

xgdgsc:
Need some workaround of `ring` dependency to build and run on windows arm64.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ codegen-units = 1
 [dependencies]
 clap = { version = "4.0.15", features = ["derive"] }
 dirs = "4"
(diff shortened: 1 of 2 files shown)


davidanthoff:
The reason I didn't enable this so far is that the x64 version works just fine on arm64 builds, we don't have native aarch64 Julia builds in any case and that this would significantly increase the build and publish times of the Windows version. I think if we ever get an arm64 build of Julia itself for Windows this definitely is something we should do, but before then, I don't really see a huge upside?

davidanthoff:
Closing this for now, we can revisit if there ever is a arm build for Windows of Julia itself.

PR #496, opened 2022-11-28
Configure for target_os, not nonexistant family

jakobnissen:
`cfg(macos)` is invalid, because `macos` is not a target family, but a target os. This causes the cfg to be ignored.
See more here:
https://rust-lang.github.io/rust-clippy/master/index.html#mismatched_target_os


jakobnissen:
It appears there are some issues in the branch that is now compiled on MacOS. I can try to look into it, but without a Mac computer and with me knowing very little about networking, it might be hopeless.

wolthom:
Hey @jakobnissen 

This is just a quick guess, but I suspect it could be related to the OS-specific dependencies section:
https://github.com/JuliaLang/juliaup/blob/main/Cargo.toml#L43-L51

I suspect the same error happened there. The `'cfg(macos)'` parts may need to be `'cfg(target_os="macos")'`. (See here: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies)

Could you try this out? I don't have a Mac available to try

PR #497, opened 2022-11-28
Various lints and minor improvements

jakobnissen:
See GitHub PR for explanation of changes

### DIFF
--- a/src/bin/juliainstaller.rs
+++ b/src/bin/juliainstaller.rs
@@ -111,39 +111,39 @@ fn print_install_choices(install_choices: &InstallChoices) -> Result<()> {
     use console::style;
 
     println!("Juliaup will be installed into the Juliaup home directory, located at:");
-    println!("");
+    println!();
     println!("  {}", install_choices.install_location.to_string_lossy());
-    println!("");
+    println!();
     println!("The {}, {} and other commands will be added to", style("julia").bold(), style("juliaup").bold());
     println!("Juliaup's bin directory, located at:");
-    println!("");
+    println!();
     println!("  {}", install_choices.install_location.join("bin").to_string_lossy());
-    println!("");
+    println!();
 
     if install_choices.modifypath {
         println!("This path will then be added to your {} environment variable by", style("PATH").bold());
         println!("modifying the profile files located at:");
-        println!("");
+        println!();
         for p in &install_choices.modifypath_files {
             println!("  {}", p.to_string_lossy());
         }
-        println!("");
+        println!();
     }
 
     if install_choices.backgroundselfupdate>0 {
         println!("The installer will configure a CRON job that checks for updates of");
         println!("Juliaup itself. This CRON job will run every {} seconds.", install_choices.backgroundselfupdate);
-        println!("");
+        println!();
     }
 
     if install_choices.startupselfupdate>0 {
         println!("Julia will look for a new version of Juliaup itself every {} minutes when you start julia.", install_choices.startupselfupdate);
-        println!("");
+        println!();
     }
 
     if install_choices.symlinks {
         println!("Julia will create a symlink for every channel you install that is named julia-<CHANNELNAME>.");
-        println!("");
+        println!();
     }
 
     Ok(())
@@ -186,7 +186,7 @@ pub fn main() -> Result<()> {
     use log::{info, trace, debug};
 
     println!("{}", style("Welcome to Julia!").bold());
-    println!("");
+    println!();
 
     if is_juliaup_installed() {
         println!("It seems that Juliaup is already installed on this system. Please remove the previous installation of Juliaup before you try to install a new version.");
@@ -211,7 +211,7 @@ pub fn main() -> Result<()> {
             return Ok(());
         }
 
-        println!("");
+        println!();
     }
 
     let mut install_choices = InstallChoices {
@@ -222,14 +222,14 @@ pub fn main() -> Result<()> {
         install_location: dirs::home_dir()
             .ok_or(anyhow!("Could not determine the path of the user home directory."))?
             .join(".juliaup"),
-        modifypath_files: find_shell_scripts_to_be_modified(true)?,
+        modifypath_files: find_shell_scripts_to_be_modified(true),
     };
 
     print_install_choices(&install_choices)?;
 
     println!("You can uninstall at any time with {} and these", style("juliaup self uninstall").bold());
     println!("changes will be reverted.");
-    println!("");
+    println!();
 
     debug!("Next running the prompt for default choices");
 
@@ -243,7 +243,7 @@ pub fn main() -> Result<()> {
 
     trace!("choice is {:?}", answer_default);
 
-    println!("");
+    println!();
 
     if answer_default == 1 {
         debug!("Next running the individual config wizard");
@@ -349,14 +349,14 @@ pub fn main() -> Result<()> {
     println!("Julia was successfully installed on your system.");
 
     if install_choices.modifypath {
-        println!("");
+        println!();
         println!("Depending on which shell you are using, run one of the following");
         println!("commands to reload the {} environment variable:", style("PATH").bold());
-        println!("");
+        println!();
         for p in &install_choices.modifypath_files {
             println!("  . {}", p.to_string_lossy());
         }
-        println!("");
+        println!();
     }
 
     Ok(())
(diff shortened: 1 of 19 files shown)


src/bin/juliainstaller.rs
```diff
@@ -111,39 +111,39 @@ fn print_install_choices(install_choices: &InstallChoices) -> Result<()> {
     use console::style;
 
     println!("Juliaup will be installed into the Juliaup home directory, located at:");
-    println!("");
+    println!();
```
jakobnissen, line 114: Like in Julia, there is no need to create an empty string to pass to `println!`

src/bin/julialauncher.rs
```diff
@@ -6,7 +6,6 @@ use juliaup::versions_file::load_versions_db;
 use normpath::PathExt;
 use std::path::Path;
 use std::path::PathBuf;
-use ctrlc;
```
jakobnissen, line 9: `ctrlc` is already in scope - no need to bring it in.

src/bin/julialauncher.rs
```diff
@@ -52,8 +51,7 @@ fn run_versiondb_update(config_file: &juliaup::config_file::JuliaupReadonlyConfi
     if let Some(val) = versiondb_update_interval {
         let should_run = if let Some(last_versiondb_update) = config_file.data.last_version_db_update {
             let update_time = last_versiondb_update + chrono::Duration::minutes(val);
-
-            if Utc::now() >= update_time {true} else {false}
+            Utc::now() >= update_time
```
jakobnissen, line 54: The pattern `if [cond] {true} else {false}` can be expressed as just `[cond]`

src/bin/julialauncher.rs
```diff
@@ -161,7 +159,7 @@ fn get_julia_path_from_channel(
         JuliaupConfigChannel::LinkedChannel { command, args } => {
             return Ok((
                 PathBuf::from(command),
-                args.as_ref().map_or_else(|| Vec::new(), |v| v.clone()),
+                args.as_ref().map_or_else(Vec::new, |v| v.clone()),
```
jakobnissen, line 162: This is similar to in Julia replacing `x -> f(x)` with `f`

src/bin/julialauncher.rs
```diff
@@ -219,8 +217,8 @@ fn run_app() -> Result<i32> {
     if args.len() > 1 {
         let first_arg = &args[1];
 
-        if first_arg.starts_with("+") {
-            julia_channel_to_use = Some(first_arg[1..].to_string());
+        if let Some(stripped) = first_arg.strip_prefix('+') {
+            julia_channel_to_use = Some(stripped.to_string());
```
jakobnissen, line 221: This way, we avoid manually accessing strings, which can throw errors like in Julia, and also only do the string stripping once.

src/bin/julialauncher.rs
```diff
@@ -250,7 +248,7 @@ fn run_app() -> Result<i32> {
     }
 
     for (i, v) in args.iter().skip(1).enumerate() {
-        if i > 1 || !v.starts_with("+") {
+        if i > 1 || !v.starts_with('+') {
```
jakobnissen, line 251: Various places I've replaced single-char `str` with `char` for efficiency and simplicity.

src/command_add.rs
```diff
@@ -13,7 +13,7 @@ pub fn run_command_add(channel: String, paths: &GlobalPaths) -> Result<()> {
     let required_version = &version_db
         .available_channels
         .get(&channel)
-        .ok_or(anyhow!(
+        .ok_or_else(|| anyhow!(
```
jakobnissen, line 16: By using `ok_or_else` with a closure, we avoid instantiating the error before it is needed. It is similar to the `if ... else` vs `ifelse` tradeoff in Julia.

src/command_info.rs
```diff
@@ -40,7 +40,7 @@ pub fn run_command_info(paths: &GlobalPaths) -> Result<()> {
     let dbversion_url = juliaupserver_base.join(dbversion_url_path)
         .with_context(|| format!("Failed to construct a valid url from '{}' and '{}'.", juliaupserver_base, dbversion_url_path))?;
 
-    let online_dbversion = download_juliaup_version(&dbversion_url.to_string())
+    let online_dbversion = download_juliaup_version(dbversion_url.as_ref())
```
jakobnissen, line 43: Various places, I've avoided cloning, or copying a `str` to a `String`, if not needed.

src/global_paths.rs
```diff
@@ -21,7 +21,7 @@ fn get_juliaup_home_path() -> Result<PathBuf> {
 
     let path = match std::env::var("JULIA_DEPOT_PATH") {
         Ok(val) => {
-            let path = PathBuf::from(val.to_string().split(entry_sep).next().unwrap()); // We can unwrap here because even when we split an empty string we should get a first element.
+            let path = PathBuf::from(val.split(entry_sep).next().unwrap()); // We can unwrap here because even when we split an empty string we should get a first element.
```
jakobnissen, line 24: Unneeded cloning here, since `str` implements `split`

src/operations.rs
```diff
@@ -684,44 +685,38 @@ fn remove_path_from_specific_file(path: PathBuf) -> Result<()> {
...
         .iter()
         .filter(|p| p.exists() ||
             (add_case && p.file_name().unwrap()==".zshrc" && std::env::consts::OS == "macos") // On MacOS, always edit .zshrc as that is the default shell, but only when we add things
         )
-        .map(|p|p.clone())
-        .collect();
-
-    Ok(result)
+        .cloned()
+        .collect()
```
jakobnissen, line 705: `.cloned()` is equivalent to mapping `clone` to each element.

src/operations.rs
```diff
@@ -684,44 +685,38 @@ fn remove_path_from_specific_file(path: PathBuf) -> Result<()> {
...
+        add_path_to_specific_file(bin_path, &p).unwrap();
     });
-
-    Ok(())
 }
 
-pub fn remove_binfolder_from_path_in_shell_scripts() -> Result<()> {
-    let paths = find_shell_scripts_to_be_modified(false)?;
+pub fn remove_binfolder_from_path_in_shell_scripts() {
+    let paths = find_shell_scripts_to_be_modified(false);
```
jakobnissen, line 717: These two functions cannot error either, so just return the unit type.

src/bin/julialauncher.rs
davidanthoff: Oh, that is nice!

src/config_file.rs
```diff
@@ -256,7 +254,7 @@ pub fn save_config_db(juliaup_config_file: &mut JuliaupConfigFile) -> Result<()>
         .with_context(|| "Failed to set len to 0 for config file before writing new content.")?;
 
     serde_json::to_writer_pretty(&juliaup_config_file.file, &juliaup_config_file.data)
-        .with_context(|| format!("Failed to write configuration file."))?;
+        .with_context(|| "Failed to write configuration file.".to_string())?;
```
davidanthoff, line 257: Do we even need the `.to_string()`? As far as I can tell not? There are more changes like that further down that probably also don't need that `.to_string()`.

src/operations.rs
```diff
@@ -684,44 +685,38 @@ fn remove_path_from_specific_file(path: PathBuf) -> Result<()> {
     Ok(())
 }
 
-pub fn find_shell_scripts_to_be_modified(add_case: bool) -> Result<Vec<PathBuf>> {
+pub fn find_shell_scripts_to_be_modified(add_case: bool) -> Vec<PathBuf> {
```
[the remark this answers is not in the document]
davidanthoff: I think it only can't return an error because my implementation is sloppy ;) The implementation right now has a bunch of `.unwrap()` statements in the function body, they should really be replaced with `?` and then the return type should be preserved, right?

src/operations.rs
davidanthoff: Same thing, I believe the proper fix is to replace the `unwrap` calls in the function body with `?` and then leave the return type as is.

davidanthoff (changes requested):
Oh, this is amazing, thanks!

I was very much learning Rust while writing Juliaup, so there are some things that I dind't know about when I wrote the code initially, and then other things I'm still not fully grasping, so this is super helpful :)

The only thing is that I think the functions that currently can't return errors should be changed to do so, their current implementation really is sloppy on my end and we should rather fix that.

src/config_file.rs
jakobnissen: You're right. I've removed the `.to_string()` here and a few similar places.

jakobnissen:
Your comments have been addressed.

davidanthoff (approved):
Excellent, merging now. Thanks :)

PR #509, opened 2022-11-29
Modify startup scripts as byte vectors instead of strings

### LINKED ISSUE #433: Don't crash on invalid shell startup files
Hi everyone,

when trying to install Juliaup on my machine, the installer itself crashes immediately after downloading the Juliaup file. During the installation I just kept all settings on their defaults. To install Juliaup I used the recommended `curl` method described within the README file.

The generated crash report:
```
name = 'Juliainstaller'
operating_system = 'unix:Arch'
crate_version = '1.7.24'
explanation = '''
Panic occurred in file 'src/operations.rs' at line 675
'''
cause = 'called `Result::unwrap()` on an `Err` value: stream did not contain valid UTF-8'
method = 'Panic'
backtrace = '''

   0: 0x55c073266e04 - juliainstaller::main::h84fedc54940da11e
   1: 0x55c073251d23 - std::sys_common::backtrace::__rust_begin_short_backtrace::hed5660ccc12c4739
   2: 0x55c073251fd1 - std::rt::lang_start::{{closure}}::h2c3639270a40ed75
   3: 0x55c07326acf2 - main
   4: 0x7f41f3890290 - <unresolved>
   5: 0x7f41f389034a - __libc_start_main
   6: 0x55c073251539 - _start
   7:        0x0 - <unresolved>'''
```

wolthom:
This PR is intended to fix #433 by matching on and replacing pure byte streams. This way, the assumption of valid UTF-8 data in the startup files is loosened. A typical case where this could cause issues is when operating system paths that are not valid UTF-8 data are used.

### DIFF
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -11,6 +11,7 @@ use bstr::ByteSlice;
 use console::style;
 use flate2::read::GzDecoder;
 use indicatif::{ProgressBar, ProgressStyle};
+use indoc::formatdoc;
 use std::io::Seek;
 use std::io::Write;
 use std::os::unix::prelude::OsStrExt;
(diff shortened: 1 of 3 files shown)


wolthom:
Small update:
I just realized that even retrieving the byte sequence of a path is highly platform specific (i.e. on Windows the `as_bytes` method is not implemented for `OsStr`).

The easiest way to handle this seems to be taking a dependency on `os_str_bytes`. It's a widely used crate that is actively maintained and has minimal dependencies (https://crates.io/crates/os_str_bytes).

@davidanthoff At this point, I'm wondering if we actually want to support non-UTF-8 paths for the `juliaup`-relevant paths.
A middle-ground could be to gracefully handle whatever is already-present in the startup scripts (via this PR), but only add valid UTF-8 from the `juliaup` side and report an error if a user tries to supply a non-UTF-8 path for the `juliaup` installation.

StefanKarpinski:
FWIW, we have occasionally seen Windows users in East Asian countries with home directories that have invalid UTF-16 in them. Not sure if that makes it worth trying to support, but it's not something that doesn't ever happen. Two things: 1. It's awful that OSes allow invalid Unicode paths; 2. Every operation you're doing on these paths would work just fine if Rust worked like bstr and Julia.

wolthom:
@StefanKarpinski Oh wow, so you actually did encounter that before. I stumbled over this when trying to write additional tests that actually exercise the invalid UTF-8 case. Do you have any examples on hand for Windows?
And yeah, Rust's pedantic handling here is technically correct but clunky when trying to work with these edge-cases.

What is your opinion on trying to support that?

StefanKarpinski:
I found an example surprisingly easily: https://github.com/JuliaLang/julia/issues/33486 (searched "Windows" and "Korean"). I'd say that it's not worth trying to handle this currently. At the moment `libuv` doesn't handle this well which forces Julia to also not handle it well. I've got a languishing PR to `libuv` that would fix it, which would eventually fix it for Julia as well, but given that Julia doesn't currently handle it, it's find for `juliaup` not to as well.

src/operations.rs
```diff
@@ -573,34 +574,57 @@ pub fn uninstall_background_selfupdate() -> Result<()> {
...
 const E_MARKER: &[u8] = b"# <<< juliaup initialize <<<";
+const HEADER: &[u8] = b"\n\n# !! Contents within this block are managed by juliaup !!\n\n";
 
-fn get_shell_script_juliaup_content(bin_path: &Path, path: &Path) -> Vec<u8> {
+fn get_shell_script_juliaup_content(bin_path: &Path, path: &Path) -> Result<Vec<u8>> {
     let mut result: Vec<u8> = Vec::new();
 
+    let bin_path_str = match bin_path.to_str() {
+        Some(s) => s,
+        None =>  bail!("Could not crate UTF-8 string from passed-in binary application path. Currently only valid UTF-8 paths are supported"),
```
uncomfyhalomacro, line 584: typo: "crate"

wolthom:
Okay the current implementation should be able to handle non-UTF-8 data in the file it is working with and properly modify that file.

It checks that the inserted path is itself valid UTF-8 and returns a (hopefully) helpful error message if not.
In case it is valid UTF-8, the corresponding bytes are inserting into the byte-blob and written back to the file.

src/operations.rs
```diff
@@ -575,38 +576,55 @@ const S_MARKER: &[u8] = b"# >>> juliaup initialize >>>";
...
-    buf.extend_from_slice(b"    export PATH=");
-    buf.extend_from_slice(p.as_os_str().as_bytes());
-    buf.extend_from_slice(b"${{PATH:+:${{PATH}}}};;\n");
-    buf.extend_from_slice(b"esac\n");    
+fn append_sh_content(buf: &mut Vec<u8>, path_str: &str) {
+    // If the variable is already contained in $PATH, do nothing
+    // Otherwise prepend it to path
+    // ${PATH:+:${PATH}} => Only append :$PATH if $PATH is set
+    let content = formatdoc!("
+            case \":$PATH:\" in
```
wolthom, line 617: Does this first case actually work properly? I copied it from the original code, but this looks wrong to me.

davidanthoff: Uh, good question. Starting at it right now, it does not look right to me either. But I'm also really not a bash expert. My guess is that it also doesn't do any harm, so I think I'll merge this PR now and then we can potentially fix this later.

davidanthoff (approved):
Looks all good to me!

Issue #257, opened 2022-03-04
using multiple DEPOT_PATH s with one julia installation


jlapeyre:
You might have `julialauncher` look in the default depot path as a fallback in case `DEPOT_PATH[1]` has no `juliaup` directory.

I'd like to use various main depots (the first entry in `DEPOT_PATH`) without running juliaup each time to install julia in the depot.

More precisely, I'd like to use juliaup to install to `~/.julia` and then later this same julia with various other depots. But, `~/.juliaup/bin/julia`, which links to the launcher of course can't find julia.

I am working on a project that manages a depot, etc. with scripts. If `~/.juliaup/bin/julia` is the first on my system path, then the real julia binary will not be found.  I imagine that other scenarios like this might arise in the future.

davidanthoff:
So the idea would simply be we search through all the paths in `DEPOT_PATH`, until we find one where with a `juliaup` install, right? That sounds like a good idea to me.

jlapeyre:
That's not what I had in mind, but it sounds like a good idea as well. I don't have a good grasp on different practical workflows that involve manipulating `DEPOT_PATH`.  Checking a few scenarios: If I do "export JULIA_DEPOT_PATH=~/newdepot:", then this depot is prepended the usual contents of `DEPOT_PATH`. If `~/newdepot` is empty, then the current behavior is `~/.juliaup/bin/julia` begins to download a julia installation. If I interrupt this, and start julia again. Then julialauncher sees `~/newdepot/juliaup` with no config file in it and says: "The Julia launcher failed to figure out ...".  In this case, searching through `DEPOT_PATH` would have found the juliaup installs.  So, it's useful if people do this.

 (Here is the the  [doc for `DEPOT_PATH`](https://docs.julialang.org/en/v1/base/constants/#Base.DEPOT_PATH)  and [for `JULIA_DEPOT_PATH`](https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH).)


But, I am doing something different. What I am trying to do is *replace* the first entry in `DEPOT_PATH`, because I want to force new packages to be installed there (I have a reason that I omit for brevity). If I leave `~/.julia` in `DEPOT_PATH`, then packages will be found there. I wouldn't be surprised if there are other use cases for this.

[...]

Issue #276, opened 2022-03-26
juliaup in a multi-user environment


dmbates:
We would like to install Juliaup as root on a departmental Linux cluster and have it available to users.  I haven't found any documentation specific to this use.  I presume we should use a custom installation but I am not entirely clear what directories would go where.  For a personal installation the are both ~/.juliaup and ~/.julia/juliaup directories created.  What would be system-wide equivalents?

johnnychen94:
Installs in `/opt/juliaup` and make a symlink to `/usr/local/bin/juliaup`?

dmbates:
Thanks for the replies @johnnychen94 and @davidanthoff .  Can you point me to some information on JULIA_DEPOT, David?  I should know what that environment variable refers to but I don't.  Essentially I want users to share the installed versions of Julia and the multiplexer and its configuration.  I would expect that the SysAdmin would need to do the installations of new versions but I hope that a non-privileged user could check
```
juliaup st
```
to find what versions are currently installed.

dmbates:
I found the description of JULIA_DEPOT in https://docs.julialang.org/en/v1/manual/environment-variables/

davidanthoff:
Well, first I used the wrong name, the env var is called `JULIA_DEPOT_PATH`, see https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH.

So, I think you are interested in some kind of variant of b). In your system, do you want users to have the ability to install packages, or is that also meant to be systemwide?

At the moment, `juliaup` treats individual Julia versions as if they were a bit like packages...

dmbates:
I'm fine with individual users having their own package repositories as it quickly becomes too much of a headache to try to maintain a system-wide package repository.

I guess I would prefer if the installations of julia versions were shared.  If they are installed for each user then the users could just install their own version of juliaup. Each installation of a Julia version weighs in at 400-500 MB and we would like to keep at least a dozen versions available in case someone needs to go back to v1.3.2 for some reason.

dmbates:
That seems reasonable.  Thanks for considering this.  For the time being we will go back to having several different `julia-1.x.y` links in our equivalent of `usr/local/bin/`

StefanKarpinski:
Getting updates would be a good task for a cron job. The tricky bit would be making sure it's atomic in case someone tries to use a Julia version right when it's being updated.

davidanthoff:
[...]

PR #510, opened 2022-11-30
Search depot path for existing juliaup dir

### LINKED ISSUE #323: Empty `DEPOT_PATH` segments throws an error
From the documented behaviour of  `JULIA_DEPOT_PATH` in https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH I'd expect passing an empty segment in `JULIA_DEPOT_PATH` to still launch `julia`, but `juliaup` throws an error when it's provided, as shown below. I can kind of work around it by pre-expanding it to what `julia` would do, but seems like `juliaup` should be able to handle this for me instead.

I assume it's probably just a case of adjusting https://github.com/JuliaLang/juliaup/blob/34b41f4f34d7dd5e5fca76761495c4bc4dfa5508/src/global_paths.rs#L24-L26 to expand to the user's home directory when encountering an empty string instead of throwing?

```
[mike@desktop ~]$ JULIA_DEPOT_PATH=":/home/mike/custom-depot" julia
Error: Trying to load all global paths.

Caused by:
    The `JULIA_DEPOT_PATH` environment variable contains a value that resolves to an an invalid path ``.
[mike@desktop ~]$ JULIA_DEPOT_PATH=":/home/mike/custom-depot" .julia/juliaup/julia-1.7.3+0~x64/bin/julia
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.7.3 (2022-05-06)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

julia>
```

jakobnissen:
Old behaviour would pick the first segment of JULIA_DEPOT_PATH to find the juliaup dir. This commit finds the first segment with a juliaup/juliaup.json file. If no such are found, it picks the first segment. Empty segments are expanded to the default path (which is ~/.julia/juliaup). If JULIA_DEPOT_PATH is not set, or it is set to the empty string, it will also use the default path.

__NOTE__: Julia docs on `JULIA_DEPOT_PATH` says that if the variable is set to the empty string, it means `DEPOT_PATH` is empty. It would be nice to be consistent with Base, and have it so an empty JULIA_DEPOT_PATH would resolve to no path instead of the default. However, this won't work for Juliaup, since it needs a directory to check for new versions. Hence, an empty variable just returns the default instead.

Note also that this approach does _not_ solve #257. If I understand that issue correctly, @jlapeyre wants Juliaup to search in the default depot path, even if that is not included in `JULIA_DEPOT_PATH`, such  that Julia can be forced to install _packages_ in a custom location, without forcing Juliaup to update to the same custom location. To solve this issue, we need to come up with a principled strategy for which directories Juliaup works in.

This might not solve #276. The user can set the env variable to `my_own_dir:systemwide_juliaup_dir`, and Juliaup will correctly pick the systemwide directory. However, the package manager will also pick that, if packages are not found in the self directory.

It might be necessary to have the Juliaup directory be separate from the package directory. Though we would need to think about exactly how. As I see it, there are two use cases:

* Common use case: User just wants to use their normal depot path both to load and install packages, and to use Juliaup and install new versions. This is what Juliaup already does.
* Uncommon use case: There is a systemwide `Juliaup` install, where individual users don't have (or should have) write permissions. Users want Juliaup to find versions in that location, but does not want Juliaup to begin writing to that location, since that would result in a permission error. Packages should be installed to a user-specific directory.

It seems to me a solution could be to have a specific env variable for Juliaup, independent of JULIA_DEPOT_PATH. Juliaup would then, if this variable is set, only look at that path. If it is not set, it would do as it does now.

Closes #323


jlapeyre:
We need some solution to decoupling the depots from the julia installation. There are reasonable scenarios where you would want this, and juliaup does not support it at the moment.
> a solution could be to have a specific env variable for Juliaup

This one seems simple.

I think that this PR *does* fix #257 (there may be some unforeseen problem.) I think the statement of the problem could be more clear in that issue.

The trailing colon in the following is interpreted as an empty path, which is expanded to the default depot.
```shell
$ export JULIA_DEPOT_PATH="~/newdepot:"
$ julia -e "println(DEPOT_PATH)"
["/home/lapeyre/newdepot", "/home/lapeyre/.julia", 
 "/home/lapeyre/packages/julias/julia-1.8/local/share/julia",
 "/home/lapeyre/packages/julias/julia-1.8/share/julia"]
```
The current behavior of juliaup is to install julia in `newdepot`.

If I understand this PR correctly:  If `newdepot` has no juliaup-installed julia, juliaup will next look in the default depot and find the exisiting installation.

davidanthoff:
Yes, exactly. I think @jlapeyre's situation would be solved in this way:
1. Run Juliaup once with `JULIA_DEPOT_PATH` empty. This will create `~/.julia/juliaup/juliaup.json` etc. At this point the `~/.julia` depot will have a Juliaup installation in it.
2. Now change things to `JULIA_DEPOT_PATH="~/newdepot:"`. At this point Juliaup will use the `~./julia` depot for everything, while the package manager is going to use the `~/newdepot`.

I think that is the whole scenario, right?

For the multi-user scenario I think the same thing would work? I.e. install Juliaup into a depot `/some/shared/path`, then do something like `JULIA_DEPOT_PATH=":/some/shared/path"`. At this point Juliaup will use the depot that is on a shared path, but packages would be installed into `~/.julia` depot, right? The only thing we need to properly handle here is that the user might not have write permissions for `/some/shared/path`. I think the correct behavior there is simply that any command that would modify things errors, with a good error message.

> It seems to me a solution could be to have a specific env variable for Juliaup, independent of `JULIA_DEPOT_PATH`. Juliaup would then, if this variable is set, only look at that path. If it is not set, it would do as it does now.

Agreed. It could just be named `JULIAUP_DEPOT_PATH`. It could either a) not exist, or b) exist hand have one path in it.

I'll try to review the PR properly later today.

jlapeyre:
> something like `JULIA_DEPOT_PATH=":/some/shared/path"`

That appends `/some/shared/path` to the standard contents of `DEPOT_PATH`. If I understand this PR correctly, then yes, this would allow using a shared julia installation.

I think the usual case is that the user does not have write permission for  `/some/shared/path`.  It would be expected that a privileged user maintains this. IIUC the user would lose the ability to easily change the default julia. Because this is written in `~/.julia/juliaup`. You might modify juliaup to store this information in `~/.juliaup` instead. This better separates more ephemeral user preferences from the installation. I wonder if rust supports things like this. I suppose not, since this is modeled on rustup.

src/global_paths.rs
```diff
@@ -19,35 +19,60 @@ pub struct GlobalPaths {
...
+                let subpath = path.join("juliaup").join("juliaup.json");
+                if subpath.is_file() {
+                    return Ok(path.join("juliaup"));
+                }
+            }
+            // If such a file does not exist, we pick the first segment in JULIA_DEPOT_PATH.
+            // This is guaranteed to be nonempty due to the properties of str::split.
+            let first_path = paths.iter().next().unwrap();
+            if !first_path.is_dir() {
+                bail!("The `JULIA_DEPOT_PATH` environment variable contains a value that resolves to an invalid directory `{}`.", first_path.display());
```
davidanthoff, line 53: I'm wondering whether in this case we should create that folder? I think we are running into a bootstrapping problem otherwise?

davidanthoff (changes requested):
Left one comment about non-existing Julia depots.

src/global_paths.rs
davidanthoff: @jakobnissen I just pushed a commit that I believe fixes this. Essentially, I just remove the check whether the path exists and return the first element no matter what. If it doesn't exist, then Juliaup will create a depot and install there, which is the behavior we want, I believe. Could you briefly review whether this all makes sense to you? If it does, I would merge.

davidanthoff (approved):

src/global_paths.rs
jakobnissen: Yes, this makes sense. Julia in general assumes a writeable first entry of `JULIA_DEPOT_PATH`, so I think it's valid.

thofma:
Hm, does that fix the problem? I updated `juliaup`, but it still downloads things for custom `JULIA_DEPOT_PATH`:

```
 ~  juliaup --version
Juliaup 1.8.16
 ~  ls .julia/juliaup/
julia-1.0.5+0.x64.apple.darwin14     julia-1.8.2+0.aarch64.apple.darwin14
julia-1.6.7+0.x64.apple.darwin14     julia-1.8.5+0.aarch64.apple.darwin14
julia-1.7.3+0.x64.apple.darwin14     juliaup.json
julia-1.8.0+0.aarch64.apple.darwin14 versiondb-aarch64-apple-darwin.json
 ~  JULIA_DEPOT_PATH=/tmp/bla3 julia
Installing Julia 1.8.5+0.aarch64.apple.darwin14
  Downloading: [=>                  ] 6.83 MiB/102.59 MiB eta: 20s
```

davidanthoff:
You need to put _both_ depot paths into `JULIA_DEPOT_PATH`. In your example, Juliaup won't know about the `~/.julia` depot at all.

jakobnissen:
@davidanthoff If I understand correctly, this happens because `juliaup` does not populate `DEPOT_PATH` the same was as Julia does. If `JULIA_DEPOT_PATH` ends with a colon, then the default values are appended to the value of the environmental variable. That is, if `JULIA_DEPOT_PATH` is `a:b:`, then DEPOT_PATH will be `["a", "b", "/home/jakni/.julia"
 "//home/jakni/.julia/juliaup/juliaup/julia-1.8.3+0.x64.linux.gnu/local/share/julia"
 "/home/jakni/.julia/juliaup/juliaup/julia-1.8.3+0.x64.linux.gnu/share/julia"]`

PR #511, opened 2022-11-30
Fix mismatch types

uncomfyhalomacro:
Running `cargo build --release --feature selfupdate` will have a mismatch type error in line 225 in `src/bin/juliainstaller.rs`. Just a small fix.


jakobnissen:
Perhaps it's a good idea to leave the `rustfmt` run as a separate PR, as that is a separate decision whether this repository should use standardized formatting. Other than that, LGTM.

davidanthoff:
Thanks, I fixed this is part of https://github.com/JuliaLang/juliaup/pull/512.

PR #514, opened 2022-12-01
fix MS store link

kafisatz:
the link should point to
https://apps.microsoft.com/store/detail/julia/9NJNWW8PVKMN


davidanthoff:
The current link is what the store management portal suggests as the link to share, so I think we should keep that as it presumably is the link that MS treats as the permalink to the app.

PR #526, opened 2022-12-15
Use kill to set signal

davidanthoff:
Maybe fixes https://github.com/JuliaLang/juliaup/pull/525#issuecomment-1352694997.

@maleadt and @staticfloat if you think this is worth trying I would merge and publish a new releasepreview build with this version.


davidanthoff:
Oh, and https://users.rust-lang.org/t/is-sigsegv-handled-by-rust-runtime/45680/2 is somewhat useful. My best guess is that Rust by default installs some signal handlers that behave differently for different signals and that is why `raise` by itself might not be enough.

davidanthoff:
Ah, @staticfloat has found a better method.

PR #551, opened 2023-01-12
Make wording on mac and linux versions less worrysome

aramirezreyes:
Juliaup offers a good experience in mac and on linux, the previous wording could be more scary than needed for the great majority of users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Juliaup can be installed on Linux or Mac by executing
 curl -fsSL https://install.julialang.org | sh
 ```
 
-in a shell. Note that the Mac and Linux version are considered prerelease, have known bugs and might often break.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -32,7 +32,7 @@ Juliaup can be installed on Linux or Mac by executing
 curl -fsSL https://install.julialang.org | sh
 ```
 
-in a shell. Note that the Mac and Linux version are considered prerelease, have known bugs and might often break.
+in a shell. Although `juliaup` offers a reliable and seamless experience on Mac and Linux, these versions are still in development and thus are considered prerelease.
```
StefanKarpinski, line 35:
```suggestion
in a shell.
```
I would just get rid of this comment altogether. Not clear to me in what meaningful sense the Linux and Mac versions are any less mature than the Windows version at this point.

StefanKarpinski (approved):
Ultimately @davidanthoff's call, but it feels like this is ready enough to not warn people (at least!).

PR #582, opened 2023-02-12
Add more cli parameters for automated install

JBlaschke:
Added some more CLI flags that allow for an automated install to a custom location. Flags added are:
1. `--path` the location where to install Juliaup to
2. `--add-to-path` when set, do not append `.bashrc` etc addind Juliaup to PATH
3. `--background-selfupdate` and `--startup-selfupdate` update intervals

This PR allows Juliaup to be used as an administrative tool -- where it:
1. can't be installed to the home directory (it might be invoked by a setup tool running as a privileged/admin user)
2. mustn't mess with local config files (such as bashrc)
3. shouldn't automagically update itself

As an example, this it how I plan to use it at NERSC:
1. Install into a global location (set by modulefiles):
```
JULIA_DEPOT_PATH=/path/to/global/shared/depot ./juliainstaller --path=/models/bin/path --add-to-path=no --background-selfupdate=0 --startup-selfupdate=0 --yes
```
2. Use juliaup to download julia versions (no more wget+untar!!!)

### DIFF
--- a/deploy/shellscript/juliaup-init.sh
+++ b/deploy/shellscript/juliaup-init.sh
@@ -37,7 +37,7 @@ USAGE:
 
 FLAGS:
     -y, --yes                   Disable confirmation prompt.
-        --no-add-to-path        Don't add to user PATH environment variable
+        --add-to-path           Add to user PATH environment variable
     -h, --help                  Prints help information
         --background-selfupdate Background self-update interval
         --startup-selfupdate    Statup self-update interval
--- a/deploy/shellscript/juliaup-init.sh
+++ b/deploy/shellscript/juliaup-init.sh
@@ -38,10 +38,9 @@ USAGE:
 FLAGS:
     -y, --yes                   Disable confirmation prompt.
         --add-to-path           Add to user PATH environment variable
-    -h, --help                  Prints help information
         --background-selfupdate Background self-update interval
         --startup-selfupdate    Startup self-update interval
-
+    -h, --help                  Prints help information
 
 OPTIONS:
     -p, --path              Custom install path
--- a/deploy/shellscript/juliaup-init.sh
+++ b/deploy/shellscript/juliaup-init.sh
@@ -36,9 +36,12 @@ USAGE:
     juliaup-init [FLAGS] [OPTIONS]
 
 FLAGS:
-    -y, --yes               Disable confirmation prompt.
-        --no-add-to-path    Don't add to user PATH environment variable
-    -h, --help              Prints help information
+    -y, --yes                   Disable confirmation prompt.
+        --no-add-to-path        Don't add to user PATH environment variable
+    -h, --help                  Prints help information
+        --background-selfupdate Background self-update interval
+        --startup-selfupdate    Statup self-update interval
+
 
 OPTIONS:
     -p, --path              Custom install path
--- a/deploy/shellscript/juliaup-init.sh
+++ b/deploy/shellscript/juliaup-init.sh
@@ -40,7 +40,7 @@ FLAGS:
         --add-to-path           Add to user PATH environment variable
     -h, --help                  Prints help information
         --background-selfupdate Background self-update interval
-        --startup-selfupdate    Statup self-update interval
+        --startup-selfupdate    Startup self-update interval
 
 
 OPTIONS:
--- a/src/bin/juliainstaller.rs
+++ b/src/bin/juliainstaller.rs
@@ -13,7 +13,7 @@ fn run_individual_config_wizard(install_choices: &mut InstallChoices, theme: & d
     trace!("install_location pre inside the prompt function {:?}", install_choices.install_location);
 
     let new_install_location = Input::with_theme(theme)
-        .with_prompt("Enter the folder where you want to install Juliaup")        
+        .with_prompt("Enter the folder where you want to install Juliaup")
         .validate_with(|input: &String| match input.parse::<PathBuf>() {
                 Ok(_) => Ok(()),
                 Err(_) => Err("Not a valid input".to_owned())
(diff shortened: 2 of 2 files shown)


deploy/shellscript/juliaup-init.sh
```diff
@@ -36,9 +36,12 @@ USAGE:
     juliaup-init [FLAGS] [OPTIONS]
 
 FLAGS:
-    -y, --yes               Disable confirmation prompt.
-        --no-add-to-path    Don't add to user PATH environment variable
-    -h, --help              Prints help information
+    -y, --yes                   Disable confirmation prompt.
+        --no-add-to-path        Don't add to user PATH environment variable
```
StefanKarpinski, line 40: I would try to express this without the negative. Otherwise if one adds values to choose from for option you end up in the situation where you have `--no-add-to-path=false` and then everyone is left scratching their heads trying to decode the double negative. The alternative is to end up with `--add-to-path` and `--no-add-to-path`, but having `--add-to-path=[yes|no|ask]` seems like the saner eventuality (and is how Julia command-line options work).

JBlaschke: Makes sense, I've renamed it to `--add-to-path` with a default value of `yes`. If the user enter neither `yes` or `no`, the installer prints a helpful error message: `Error: --add-to-path can be either 'yes' or 'no', you entered: 'true'`

I was tempted to make it be able to parse `true`, and `1` for `yes` and so on, but I wanted to keep all these changes minimal.

deploy/shellscript/juliaup-init.sh
```diff
@@ -36,9 +36,12 @@ USAGE:
     juliaup-init [FLAGS] [OPTIONS]
 
 FLAGS:
-    -y, --yes               Disable confirmation prompt.
-        --no-add-to-path    Don't add to user PATH environment variable
-    -h, --help              Prints help information
+    -y, --yes                   Disable confirmation prompt.
+        --add-to-path           Add to user PATH environment variable
+    -h, --help                  Prints help information
```
davidanthoff, line 41: Maybe this should be listed last?

deploy/shellscript/juliaup-init.sh
```diff
@@ -36,9 +36,12 @@ USAGE:
...
+        --background-selfupdate Background self-update interval
+        --startup-selfupdate    Statup self-update interval
```
davidanthoff, line 43:
```suggestion
        --startup-selfupdate    Startup self-update interval
```

src/bin/juliainstaller.rs
```diff
@@ -99,7 +99,19 @@ struct Juliainstaller {
...
+    alternate_path: String,
+    /// Control adding the Juliaup dir to the PATH
+    #[clap(long = "add-to-path", default_value = "yes")]
+    add_to_path: String,
+    /// Manually specify the background self-update interval
+    #[clap(long = "background-selfupdate", default_value = "0")]
+    background_selfupdate_interval: String,
+    /// Manually specify the statup self-update interval
+    #[clap(long = "startup-selfupdate", default_value = "1440")]
+    startup_selfupdate_interval: String
```
davidanthoff, line 114: Is there a way to use more specific types than `String` here? And have `clap` do the basic validation and type conversion?

src/bin/juliainstaller.rs
```diff
@@ -234,17 +256,23 @@ pub fn main() -> Result<()> {
     }
 
     let mut install_choices = InstallChoices {
-        backgroundselfupdate: 0,
-        startupselfupdate: 1440,
+        backgroundselfupdate: args.background_selfupdate_interval.parse().unwrap(),
+        startupselfupdate: args.startup_selfupdate_interval.parse().unwrap(),
```
davidanthoff, line 260: In general we should not use `unwrap` for anything where a user might input incorrect values, the error message is not pretty for it ;) I think in this case the right strategy would be to use a numeric data type in `clap`, and then we don't even have to worry about this.

src/bin/juliainstaller.rs
```diff
@@ -234,17 +256,23 @@ pub fn main() -> Result<()> {
...
         symlinks: false,
-        modifypath: true,
+        modifypath: args.add_to_path.eq("yes"),
```
davidanthoff, line 262: I wonder whether clap has some way to make the Rust type here a bool, and then in clap specify which string value is mapped to `true` and `false`?

src/bin/juliainstaller.rs
```diff
@@ -234,17 +256,23 @@ pub fn main() -> Result<()> {
...
         install_location: dirs::home_dir()
             .ok_or(anyhow!("Could not determine the path of the user home directory."))?
             .join(".juliaup"),
         modifypath_files: find_shell_scripts_to_be_modified(true)
             .with_context(|| "Failed to identify the shell scripts that need to be modified.")?,
     };
 
+    // args.alternate_path is used to overwrite the default install location (if
+    // specified). This is also the location use if disable_confirmation_prompt
+    if ! args.alternate_path.is_empty() {
```
davidanthoff, line 272: I think instead of checking for an empty string, there might be a way to use an option and check that?

davidanthoff (changes requested):
Great! I left some smallish comments here and there. If you know how to address these, great! If not, ping me and I might try to push to this branch with some small changes.

deploy/shellscript/juliaup-init.sh
JBlaschke: Makes sense. Moving.

src/bin/juliainstaller.rs
JBlaschke: It's been a while, but I've spent this time ;earning more about rust and clap -- so this is me taking another stab at it.

src/bin/juliainstaller.rs
JBlaschke: See my resolved comments above -- I've cleaned up the parser.

src/bin/juliainstaller.rs
JBlaschke: I had to mess about with `BoolishValueParser`s -- and the whole thing had to be wrapped in `Option<bool>` to avoid it from becoming a flag. I think the solution now is a good one.

src/bin/juliainstaller.rs
JBlaschke: Makes sense now that I've learned quite a bit more about rust. Using `Option<String>` + a `match` statement now.

JBlaschke:
After learning some more about rust and clap, I think this is pretty much ready. @davidanthoff Let me know what you think.

davidanthoff (approved):
Great, thanks!

Issue #584, opened 2023-02-17
Add ability to relink a channel (mainly dev)


ndinsmore:
It would be great if you could relink a channel like dev to a new location without having to undefault it then remove it first.

AbhinavMir:
Something like `juliaup link update dev <DIR>`? I can take you up on this. Just need this clarification (or if you have a sample command in mind).

AbhinavMir:
Command update already exists, any other name we could think of?

https://github.com/JuliaLang/juliaup/blob/051886cff975564d3fdf09dcd5330d9764ccefc8/src/bin/juliaup.rs#L180

ndinsmore:
The problem is that right now you have to do 4 things if a link channel is your default
1. Change default to a different channel
2. Delete the link channel
3. recreate link to new location
4. change the default back to that link

It would be great if that was a single step. So if relink does all those steps that is great.

AbhinavMir:
Best to also write out tests, I was OOO yesterday so didn't get to work on this. Will add some tests.

PR #593, opened 2023-03-09
Add ability to relink a channel

AbhinavMir:
Fixes https://github.com/JuliaLang/juliaup/issues/584


AbhinavMir:
I forgot to get back to you (didn't get a notification!). I have to test this out still. We could probably replace link but it makes sense to have two different commands in my opinion from an explicitness point of view.

davidanthoff:
To me just having `link` always work seems way easier? We could prompt if a channel name already exists with something like `Channel X already exists, do you want to replace the existing channel? (y/N)`, and a `-f` flag to force override.

PR #595, opened 2023-03-14
run cargo format on all files

KristofferC:
I was working on https://github.com/JuliaLang/juliaup/issues/257#issuecomment-1403391982 but all the trailing whitespace and unconventional formatting made the diff kind of annoying.
I think it is a good idea to do a big format once and for all. To not lose blame information we can do a `.git-blame-ignore-revs` file after (https://michaelheap.com/git-ignore-rev/)

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -8,31 +8,38 @@ extern crate winres;
 mod jsonstructs_versionsdb;
 
 use anyhow::Result;
+use serde_json::Value;
 use std::env;
 use std::fs::File;
-use std::path::PathBuf;
 use std::path::Path;
-use serde_json::Value;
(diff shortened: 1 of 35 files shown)


DilumAluthge:
Should we also have a CI job that runs the formatter and makes sure that there are no formatting issues? So then, we won't accidentally merge a PR with bad formatting?

DilumAluthge (approved):

KristofferC:
> Should we also have a CI job that runs the formatter and makes sure that there are no formatting issues? 

Yeah, probably a good idea.

KristofferC:
I'll merge this and make a PR to exclude this from git blame.

PR #596, opened 2023-03-14
disentangle JULIA_DEPOT_PATH from Juliaup and use JULIAUP_HOME instead

KristofferC:
The current "piggybacking" of `JULIA_DEPOT_PATH` in juliaup makes no sense to me. The depot path is a Julia setting but it does not (and should not) configure the Julia installation itself. Doing so makes it annoying when you want to change this Julia setting (e.g use a single temporary depot path) with your currently installed Julia version. Losing access to the Julia installation is not something desirable when changing the depot path.

This PR removes any usage of the depot path in juliaup. Instead, it stores all config files and julia installations in `~/.juliaup` by default, changeable with the env variable `JULIAUP_HOME` (in the same way as rustup). The subdir in the juliaup home directory for the installs is `juliainstallations`.

Fixes: https://github.com/JuliaLang/juliaup/issues/257

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The installer also bundles a full Julia version manager called `juliaup`. One ca
 
 ## Status
 
-This installer is considered production ready. 
+This installer is considered production ready.
 
 ## Installation
 
@@ -105,6 +105,11 @@ All of these channels can be combined with the `~x86`, `~x64` or `~aarch64` suff
 To launch the default Julia version simply run `julia` in your terminal.
 
 To launch a specific Julia version, say in channel `release`, run `julia +release`.
+
+## Path used by Juliaup
+
+Juliaup will by default use `~/.juliaup` to store Julia versions and log files. This can be changed by setting
+the `JULIAUP_HOME` environment variable.
 
 ## Juliaup server
 
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -147,6 +147,7 @@ fn get_julia_path_from_channel(
     config_data: &JuliaupConfig,
     channel: &str,
     juliaupconfig_path: &Path,
+    juliainstalls_path: &Path,
     julia_version_from_cmd_line: bool,
 ) -> Result<(PathBuf, Vec<String>)> {
     let channel_info = if julia_version_from_cmd_line {
@@ -177,9 +178,7 @@ fn get_julia_path_from_channel(
                     channel
                 )
             })?;
-            let absolute_path = juliaupconfig_path
-                .parent()
-                .unwrap() // unwrap OK because there should always be a parent
+            let absolute_path = juliainstalls_path
                 .join(path)
                 .join("bin")
                 .join(format!("julia{}", std::env::consts::EXE_SUFFIX))
@@ -235,6 +234,7 @@ fn run_app() -> Result<i32> {
         &config_file.data,
         &julia_channel_to_use,
         &paths.juliaupconfig,
+        &paths.juliainstalls,
         julia_version_from_cmd_line,
     )
     .with_context(|| {
--- a/src/global_paths.rs
+++ b/src/global_paths.rs
@@ -8,6 +8,7 @@ pub struct GlobalPaths {
     pub juliaupconfig: PathBuf,
     pub lockfile: PathBuf,
     pub versiondb: PathBuf,
+    pub juliainstalls: PathBuf,
     #[cfg(feature = "selfupdate")]
     pub juliaupselfhome: PathBuf,
     #[cfg(feature = "selfupdate")]
@@ -17,62 +18,24 @@ pub struct GlobalPaths {
 }
 
 fn get_juliaup_home_path() -> Result<PathBuf> {
-    let entry_sep = if std::env::consts::OS == "windows" {
-        ';'
-    } else {
-        ':'
-    };
-
-    match std::env::var("JULIA_DEPOT_PATH") {
+    match std::env::var("JULIAUP_HOME") {
         Ok(val) => {
-            // Note: Docs on JULIA_DEPOT_PATH states that if it exists but is empty, it should
-            // be interpreted as an empty array. This code instead interprets it as a 1-element
-            // array of the default path.
-            // We interpret it differently, because while Julia may work without a DEPOT_PATH,
-            // Juliaup does not currently, since it must check for new versions.
-            let mut paths = Vec::<PathBuf>::new();
-            for segment in val.split(entry_sep) {
-                // Empty segments resolve to the default first value of
-                // DEPOT_PATH
-                let segment_path = if segment.is_empty() {
-                    get_default_juliaup_home_path()?
-                } else {
-                    PathBuf::from(segment.to_string())
-                };
-                paths.push(segment_path);
-            }
-
-            // First, we try to find any directory which already is initialized by
-            // Juliaup.
-            for path in paths.iter() {
-                let subpath = path.join("juliaup").join("juliaup.json");
-                if subpath.is_file() {
-                    return Ok(path.join("juliaup"));
-                }
-            }
-            // If such a file does not exist, we pick the first segment in JULIA_DEPOT_PATH.
-            // This is guaranteed to be nonempty due to the properties of str::split.
-            let first_path = paths.iter().next().unwrap();
-            return Ok(first_path.join("juliaup"));
+            return Ok(std::path::PathBuf::from(val));
         }
-        Err(_) => return get_default_juliaup_home_path(),
+        Err(_) => {
+            // Return ~/.julia/juliaup, if such a directory can be found
+            let path = dirs::home_dir()
+                .ok_or_else(|| anyhow!("Could not determine the path of the user home directory."))?
+                .join(".juliaup");
+            if !path.is_absolute() {
+                bail!(
+                    "The system returned an invalid home directory path `{}`.",
+                    path.display()
+                );
+            };
+            Ok(path)
+        }
     }
-}
-
-/// Return ~/.julia/juliaup, if such a directory can be found
-fn get_default_juliaup_home_path() -> Result<PathBuf> {
-    let path = dirs::home_dir()
-        .ok_or_else(|| anyhow!("Could not determine the path of the user home directory."))?
-        .join(".julia")
-        .join("juliaup");
-
-    if !path.is_absolute() {
-        bail!(
-            "The system returned an invalid home directory path `{}`.",
-            path.display()
-        );
-    };
-    Ok(path)
 }
 
 pub fn get_paths() -> Result<GlobalPaths> {
@@ -94,6 +57,8 @@ pub fn get_paths() -> Result<GlobalPaths> {
 
     let lockfile = juliauphome.join(".juliaup-lock");
 
+    let juliainstalls = juliauphome.join("juliainstalls");
+
     #[cfg(feature = "selfupdate")]
     let juliaupselfhome = my_own_path
         .parent()
@@ -110,6 +75,7 @@ pub fn get_paths() -> Result<GlobalPaths> {
         juliaupconfig,
         lockfile,
         versiondb,
+        juliainstalls,
         #[cfg(feature = "selfupdate")]
         juliaupselfhome,
         #[cfg(feature = "selfupdate")]
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -339,7 +339,7 @@ pub fn install_version(
         .join("BundledJulia");
 
     let child_target_foldername = format!("julia-{}", fullversion);
-    let target_path = paths.juliauphome.join(&child_target_foldername);
+    let target_path = paths.juliainstalls.join(&child_target_foldername);
     std::fs::create_dir_all(target_path.parent().unwrap())?;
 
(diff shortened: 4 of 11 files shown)


KristofferC:
If there is a worry that users would have to reinstall their julia installations, we can fall back to trying to find them in `.julia/juliaup`.

cc @davidanthoff

StefanKarpinski:
Would it make sense to default to `~/.julia/juliaup` instead? I guess the downside of that is if someone deletes their `~/.julia` they also lose all their juliaup installed julia versions, but that could also be seen as a benefit since it means uninstalling julia only requires deleting `~/.julia` as opposed to `~/.julia` and `~/.juliaup`.

KristofferC:
I don't think the julia configuration folder should be related at all to the Julia installation itself. I want to be able to nuke `.julia` and still be able to start julia (just like I can when I install julia in some other way).,

guilhermebodin:
Hi, 

I am trying to install juliaup in a cloud environment and I don't have permission from my IT managers to install julia/juliaup or anything on the homedir. I have a very simple sh to install julia and jill in a tmp folder that follows my IT rules and the script does not currently work with juliaup. The reason is that although it is possible to choose the juliaup installation when I enter the CLI it is not possible when I run with `-y` flag. This PR would most probably solve all of my issues.

These are the variables that I must use in Jill (I could put download and install in the same place)
```
export JULIA_DOWNLOAD="/tmp/julia/download"
export JULIA_INSTALL="/tmp/julia/install"
export JULIA_DEPOT_PATH="/tmp/julia/.julia"
```

As I mentioned it is currently working with jill but I would really like to change it to work with juliaup.

KristofferC:
As a concrete example, I want to run https://github.com/maleadt/creduce_julia which modifies depot path and all of my sudden my `julia` command doesn't work anymore.

davidanthoff:
Sorry that I sat so long on this, busy with other stuff. Here is my take on this.

I entirely understand where @KristofferC is coming from, I originally had pretty similar thoughts. I think there is an alternative view, though, that is also reasonable that in the end informed the current design.

In that view the "brand" `juliaup` really fades away, i.e. this whole Juliaup machinery just becomes what it means to have Julia installed on a system. Essentially the definition of "I have Julia installed" then is "I have a `julialauncher` (also symlinked as `julia`) and `juliaup` on my `PATH`". In that view, the Julia binaries are just more stuff that one needs to run a given piece of Julia code, similar to how one needs certain packages and Julia artifacts. And then it seems fairly logical to say that we already store all the stuff one needs to run a given piece of Julia code in the depot. All this current design does is add yet another type of dependency that one needs to run specific Julia code to the depot.

In my mind, in the abstract, both of these designs are fairly reasonable approaches, and it is kind of hard to decide between without looking at some more specific scenarios. Here are the ones I've been thinking about:

- In general, my hope is that we can get to a point where like 80% of users never hear the term `juliaup`, don't know what it is and don't have to interact with that concept at all. On Windows we have that at install time (you install "Julia" from the Windows Store, not "Juliaup"), and I hope that we can eventually also make `sudo apt install julia` and `brew install julia` do the same thing (i.e. install `julialauncher` and `juliaup`). I then also hope that we can do regular version updates interactively at Julia launch. So, for a normal user, instead of printing a message "There is a new version of Julia, run `juliaup up`" we would print "There is a new version of Julia, do you want to update (Y/N)?" I think for a large number of users that would essentially eliminate the need to be familiar with `juliaup` entirely. In that world, I think it would _not_ be good to put a `~/.juliaup` folder on every user system, I think that is more confusing than not.
- I think in general it is good to not spread files all over user systems, I think the current Julia design where you know that essentially everything Julia related is in one place is clean and simple to understand. If we move the Juliaup binaries out of the depot, we will by default have a basic Julia install spread out over three locations on many user systems: the system location where `julialauncher` and `juliaup` get installed (unless one uses the standalone installer), the location where the Julia binaries end up `~/.juliaup` and the Julia depot `~/.julia`.
- I think there is some logic that everything that all the dependencies that are recorded in a manifest go into the same location, i.e. the depot. We haven't implemented that yet, but Julia versions are recorded in a manifest, and once Juliaup picks uses that information I think that would also be nice.
- I think the current design works well for the vast majority of users. I think the only feedback we got on it is from power-users that want to do some advanced stuff that most regular users probably will never touch. But we _never_ heard anything back from a regular user that they found the current design confusing, or not working or anything like that. To me that is a really, really important point.

I _do_ think that the scenarios that motivated @KristofferC originally are valid and that we need to find a better solution for those. I think my current thinking would be to try to do that in a way that makes it easy for power-users, but leave the current implementation in place for the rest of the user-base.

One way to do that would simply be that if no `JULIAUP_HOME` is defined, everything stays the way it is. But if `JULIAUP_HOME` is defined, then it just completely overrides the depot story, i.e. at that point the content of `JULIA_DEPOT_PATH` is completely ignored and it switches to the implementation in this PR.

One major benefit of that would also be that it is completely backwards compatible with existing installations.

StefanKarpinski:
I think that viewpoint makes sense, @davidanthoff. It might also be handy to have a way to copy a juliaup installation elsewhere or at least docs on how to do it.

KristofferC:
I'm honestly surprised this is getting push back at all. I'll try say the same thing in a couple of ways to try get my point across again.

- A julia installer that breaks correctly written julia tools and packages is not a good julia installer.
- The recommended Julia installer should for example not swap out the julia binary under your feet when you run various correctly written tools. This currently happens in some cases and if you don't notice it you run your tool with the wrong julia version.
- The default settings of a Julia installer should be so that programs written in Julia work correctly.
- Juliaup right now is buggy, programs and tools running with a juliaup julia does not work the same as those running with a normal installation. 

Now, addressing the post above. 

> It uses one folder less
This is obviously true since it uses an existing folder. But this folder is the julia depot and that *already has a meaning* and you cannot just repurpose an existing folder for something new that breaks existing uses. I mean, I honestly don't know what more to say. Is one folder less preferable over a properly functioning julia install? Am I going crazy?

> I think there is some logic that everything that all the dependencies that are recorded in a manifest go into the same location, i.e. the depot. 

The julia installation cannot go into the depot because you should be able to run your julia with a cleared depot.

>  But we never heard anything back from a regular user that they found the current design confusing, or not working or anything like that. To me that is a really, really important point.

Most users in general don't hit bugs. That doesn't mean they shouldn't be fixed. And doing the proposed file structure here would in no way be any more confusing to a regular user since they don't interact with the paths here at all. So I'm not sure what weight this argument holds.

> One way to do that would simply be that if no JULIAUP_HOME is defined, everything stays the way it is.

Why would the default behavior be the one where important tools don't work? For your average user, the installation location doesn't matter (unless they happen to run a package/tool that assumes that DEPOT_PATH works like it should). For the advanced user it is annoying when you install juliaup and forget to set the "i want my julia to work correctly" flag and get incorrect behavior.

davidanthoff:
> As a concrete example, I want to run https://github.com/maleadt/creduce_julia which modifies depot path and all of my sudden my julia command doesn't work anymore (or rather it installs 1.8 instead of using my julia).

I now looked into that one. I think what is happening there is that the tool added a hack to work around a bug in Juliaup that has since been fixed. I think if that Juliaup specific hack is removed from `creduce_julia`, then things should work just fine with the existing Juliaup because the way `creduce_julia` uses layered depots strikes me as entirely compatible with the Juliaup use of depots. I opened a PR [here](https://github.com/maleadt/creduce_julia/pull/5), would be great if someone could give that a try and report back.

@KristofferC you wrote that multiple existing tools are broken with the current design, could you point me to the specifics? I'd like to understand those scenarios better.

DilumAluthge:
I think previously I agreed with the "Juliaup should store data inside `~/.julia/juliaup`" approach. But now I'm firmly in the "Juliaup should install into `~/.juliaup`" camp, for the reasons listed by Kristoffer above

DilumAluthge:
> layered depots

Layered depots don't really solve the problem. It's very common for me to do stuff like `JULIA_DEPOT_PATH=$(mktemp -d) julia`. I can't use layered depots, because the whole point of what I'm trying to test is that I'm in a completely fresh depot environment.

davidanthoff:
@DilumAluthge would some of the ideas in https://github.com/JuliaLang/juliaup/issues/257#issuecomment-1537235627 solve your scenario? In particular `JULIAUP_DEPOT_PATH`?

DilumAluthge:
Not really, because then Pkg can see that other depot, but I only want Pkg to see the single temp depot.

davidanthoff:
But the `JULIAUP_DEPOT_PATH` idea mentioned in my link would address that, right?

KristofferC:
> @KristofferC you wrote that multiple existing tools are broken with the current design, could you point me to the specifics? I'd like to understand those scenarios better.

The simplest example is enough:

```julia
❯ julia -e 'print(VERSION)'
1.9.0-rc3
❯ JULIA_DEPOT_PATH=/tmp/testing/ julia -e 'VERSION'
Installing Julia 1.8.5+0.x64.linux.gnu
1.8.5
```

Changing `DEPOT_PATH` is not meant to change your julia installation and any tool relying on changing the depot path is vulnerable to this.

Concretely, I just hit this again running https://github.com/timholy/JuliaCon2022_Precompilation/blob/teh/jan_2023/analysis/cache_external/execute.sh trying to reproduce some data for the 1.9 release highlight.

```
~/JuliaTests/JuliaCon2022_Precompilation/analysis/cache_external teh/jan_2023*
❯ ./execute.sh                                                        
Installing Julia 1.8.5+0.x64.linux.gnu
Error: The Julia launcher failed to determine the command for the `1.9` channel.

Caused by:
    Invalid channel specified
Error: The Julia launcher failed to determine the command for the `1.8` channel.

Caused by:
    Invalid channel specified
Error: The Julia launcher failed to determine the command for the `1.7` channel.

Caused by:
    Invalid channel specified
```

You can see, the script even has comments for working around the bug by *reinstalling* the relevant julia versions in the new depot.

davidanthoff:
I should say, I'm not particularly attached to the current use of `JULIA_DEPOT_PATH`. The thing that I _really_ would like to keep is that by default the Juliaup storage is inside the default depot, i.e. inside `~/.julia`.

So maybe another way to handle this is to just replace all instances of `JULIA_DEPOT_PATH` in Julia with `JULIAUP_DEPOT_PATH` and call it a day? I haven't thought through all of the implications, but I think that would presumably address most of the concerns about the current implementation, while at the same time allowing us to keep the default storage at `~/.julia/juliaup`?

KristofferC:
It is better but I still don't like the "hijacking" of someone elses folder. It's mixing up responsibilities between julia settings and julia itself.  But it would probably fix it to the degree that I wouldn't care because I don't nuke my `.julia` very often (which other people seem to do though)

staticfloat:
I would be happy to just have it always store into `.julia`.  There are plenty of other packages that are already littering stuff into that folder, sadly.

davidanthoff:
So, one general problem is that of backwards compatibility. Essentially Juliaup follows a model where we can never break any backwards compatibility because it continuously updates in the background on user systems with no ability to block an update... If we make either @KristofferC's or my suggested change, we _are_ going to break anyone who relies on the current behavior of Juliaup to pick up the value of `JULIA_DEPOT_PATH`...

I think the only way to really avoid that would be something along the lines of
- if `JULIAUP_DEPOT_PATH` does not exist, keep the current behavior,
- if `JULIAUP_DEPOT_PATH` does exist, Juliaup ignores `JULIA_DEPOT_PATH`.

It would mean that folks that want to do the kind of thing that @KristofferC is concerned about would have to manually set `JULIAUP_DEPOT_PATH` to `:` or empty or something like that to then get Juliaup to ignore `JULIA_DEPOT_PATH`.

So, I don't know... I am reluctant to set a precedent to make a breaking change, especially given that we have said that Juliaup is production ready for a while now. On the other hand, maybe this one, last time, really exceptionally situation?

staticfloat:
I'm not particularly worried about backwards compatibility in this case, because I think backwards compatibility is most useful when changing a behavior that users could be using, but I'm having a hard time thinking up a workflow where the current behavior is desirable.  There are legitimate reasons to maintain multiple Julia depots, but they're almost exclusively around layering multiple depots, rather than having multiple separate depots.  Can you think of a workflow where it makes sense for the global `julia` version to change/disappear when you change `JULIA_DEPOT_PATH`?  If not, I think of this more as a bugfix than as a breaking change, and I'm against the idea of keeping the current behavior by default.

I think a `JULIAUP_DEPOT_PATH` (or perhaps `JULIAUP_PATH`, since I argue that `juliaup` has nothing to do with Julia depots, we just happen to store it inside of `~/.julia`) makes a lot of sense.

If you _really_ want to allow users to keep the current behavior, I think supporting a special syntax of `JULIAUP_DEPOT_PATH="\${JULIA_DEPOT_PATH}"` could make sense, where if `juliaup` notices that the setting of its search path is precisely the string `${JULIA_DEPOT_PATH}` or `$JULIA_DEPOT_PATH`, it defaults to reading in the value of that environment variable instead.  But again, I am in search of a usecase where this actually makes sense, as anyone who wants to programmatically switch `julia` versions should instead be using the `julia +foo` interface to manage multiple side-by-side versions of Julia, rather than swapping out the installed versions underneath `juliaup`, IMO.

PR #613, opened 2023-04-10
WIP: Ureq to reqwest

uncomfyhalomacro:
Hello. This is my first time to contribute to this project. As I don't have access with any MacOS machines, I cannot test this change on MacOS so I just added back ureq from this commit 88a98a0eb9f364c67aa1ea3058820d04523a2cbd . For now, I am not sure if this will work on Linux (for the most part I tested it and it worked well on my openSUSE machine).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,7 +55,7 @@ jobs:
           - label: x86_64-unknown-linux-musl
             target: x86_64-unknown-linux-musl
             os: ubuntu
-            features: selfupdate native-tls-vendored
+            features: selfupdate
             rustflags: -C target-feature=+crt-static
           - label: i686-pc-windows-msvc-windowsstore
             target: i686-pc-windows-msvc
@@ -75,7 +75,7 @@ jobs:
           - label: i686-unknown-linux-musl
             target: i686-unknown-linux-musl
             os: ubuntu
-            features: selfupdate native-tls-vendored
+            features: selfupdate
             rustflags: -C target-feature=+crt-static
           - label: aarch64-unknown-linux-gnu
             target: aarch64-unknown-linux-gnu
@@ -85,7 +85,7 @@ jobs:
           - label: aarch64-unknown-linux-musl
             target: aarch64-unknown-linux-musl
             os: ubuntu
-            features: selfupdate native-tls-vendored
+            features: selfupdate
             rustflags: -C target-feature=+crt-static
           - label: aarch64-apple-darwin
             target: aarch64-apple-darwin
@@ -96,8 +96,6 @@ jobs:
     - uses: actions/checkout@v3
     - if: ${{ contains(matrix.target, '-musl') }}
       run: sudo apt-get install musl-tools
-    - if: ${{ matrix.os == 'ubuntu' }}
-      run: sudo apt-get install libssl-dev pkg-config
     - uses: actions-rs/toolchain@v1
       with:
         toolchain: stable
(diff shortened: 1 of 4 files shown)


uncomfyhalomacro:
Hmm openssl build errors?

uncomfyhalomacro:
Commit https://github.com/JuliaLang/juliaup/pull/613/commits/11405a1e2482fa435a69faf4c8ae0d3513db80d2 changes to only check non-windows to test if it builds in macos correctly. however tests are cancelled because of a build failure.

uncomfyhalomacro:
seems i just need to disable features of reqwest.

Cargo.toml
```diff
@@ -45,12 +45,11 @@ indoc = "1.0.7"
...
-[target.'cfg(target_os = "macos")'.dependencies]
-ureq = { version = "2", features = ["native-tls"] }
-native-tls = "0.2.10"
+[target.'cfg(not(any(target_os = "windows", target_env = "musl")))'.dependencies]
+reqwest = { version = "0.11.16",  default-features = false, features = ["blocking", "native-tls"] }
 
-[target.'cfg(all(not(target_os = "macos"),not(windows)))'.dependencies]
-ureq = { version = "2", features = ["native-certs"] }
+[target.'cfg(all(target_os = "linux", target_env = "musl"))'.dependencies]
+reqwest = { version = "0.11.16",  default-features = false, features = ["blocking", "native-tls-vendored"] }
```
davidanthoff, line 52: I think you want the entire block for `reqwest` look like this:
```
[target.'cfg(target_os = "macos")'.dependencies]
reqwest = { version = "0.11", default-features = false, features = ["blocking", "native-tls"] }

[target.'cfg(all(not(target_os = "macos"),not(windows)))'.dependencies]
reqwest = { version = "0.11", default-features = false, features = ["blocking", "rustls-tls-native-roots"] }
```

I think that would be equivalent to how we handle this right now in terms of using certificates etc. I think you then also don't need to worry about installing `libssl` etc.

.github/workflows/test.yml
davidanthoff, line 1: I think you can remove all changes in this file if you adopt my suggestion below from `Cargo.toml`.

src/operations.rs
```diff
@@ -48,7 +48,7 @@ where
 }
 
 #[cfg(not(windows))]
-fn get_proxy(url: &str) -> Option<Result<ureq::Proxy>> {
+fn get_proxy(url: &str) -> Option<Result<reqwest::Proxy>> {
```
davidanthoff, line 51: Do we even need this, or does `reqwest` handle proxies fully automatically?

davidanthoff (commented):
Left a couple of comments. Let me know if you want to address them, otherwise I might also just push to this branch?

Cargo.toml
davidanthoff: I just pushed a commit that fixes this.

.github/workflows/test.yml
davidanthoff: I also did this.

src/operations.rs
davidanthoff: I removed all the manual proxy handling, my understanding is that reqwest does this transparently in the background.

uncomfyhalomacro:
> Left a couple of comments. Let me know if you want to address them, otherwise I might also just push to this branch?

Hello. thanks for reviewing and merging my PR. I didn't notice as I got busy as well. I haven't simplified it as I was just trying to not change the logic at first.

davidanthoff:
> EDIT: Seems most of the code changes are still yours. oooof

Well, I just deleted stuff ;) Your initial work was super helpful, much appreciated!

Issue #10, opened 2020-07-14
Start specific Julia version if it is specified in Project.toml/Manifest.toml


davidanthoff:
This is probably a bit speculative at this point, but it would be fairly simply to look at a Project.toml/Manifest.toml if one is specified via the `--project` command line arguments and see whether these specify a specific Julia version. If they do, run things in that specific version.

Of course, right now there is no official way to specify a Julia version in a Manifest.toml. For Project.toml, the situation is a bit different: at least for packages one can specify a version in the `compat` section. The mybinder integration for example looks for a `julia` entry in the compat section to decide what Julia version to use. We could do something similar, for example if there was an entry `julia = "=1.3.1"` in the Project.toml, we could start Julia 1.3.1.

But maybe this should also just wait until there is an official way to record Julia versions in Manifest.toml files.

davidanthoff:
The entire handling of this from `rustup` is interesting: https://rust-lang.github.io/rustup/overrides.html. We might want to add the same options here as well.

ericphanson:
> Of course, right now there is no official way to specify a Julia version in a Manifest.toml

Just to say, Pkg in Julia 1.7+ already does this! See https://github.com/JuliaLang/Pkg.jl/pull/2561 and followups.

davidanthoff:
Ah, nice!

I think the big question is how we would want this to actually behave. One option would be that we always try to use the exact Julia version that is recorded in the manifest, but maybe that is a bit too drastic, as one probably would want to normally at least use the closest minor matching version, or something like that... Not really sure what the right design here is :)

johnnychen94:
When this is supported, I would imagine adding some alias

```
alias julia='juliaup'
alias julia-1='juliaup launch 1'
```

and everything just works smarter.

davidanthoff:
No one is working on this, as far as I know. I thought about it, but that is it :)

In my mind, the override hierarchy should probably look like this, highest priority first:
1. A channel selector used on the command-line, such as `julia +beta` (implemented).
2. The `JULIAUP_CHANNEL` environment variable (not implemented).
3. A directory override that is stored in the central `juliaup.json` config file in the Julia depot (not implemented).
4. The `.juliaup.toml` file (not implemented, could also come later).
5. The Julia version stored in the `Manifest.toml` (not implemented).
6. The default Juliaup channel, stored in the `Juliaup.json` config file in the Julia depot (implemented).

[...]

PR #632, opened 2023-05-08
Add JULIAUP_CHANNEL env variable support

davidanthoff:
Implements https://github.com/JuliaLang/juliaup/issues/10#issuecomment-1518253416.

Fixes https://github.com/JuliaLang/juliaup/issues/115.

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -9,9 +9,9 @@ use std::path::Path;
 use std::path::PathBuf;
 
 #[derive(thiserror::Error, Debug)]
-pub enum JuliaupInvalidChannel {
-    #[error("Invalid channel specified")]
-    FromCmdLine(),
+#[error("{msg}")]
+pub struct UserError {
+    msg: String    
 }
 
 fn get_juliaup_path() -> Result<PathBuf> {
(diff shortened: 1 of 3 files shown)


src/bin/julialauncher.rs
```diff
@@ -211,18 +216,36 @@ fn run_app() -> Result<i32> {
...
 
         if let Some(stripped) = first_arg.strip_prefix('+') {
-            julia_channel_to_use = Some(stripped.to_string());
-            julia_version_from_cmd_line = true;
+            channel_from_cmd_line = Some(stripped.to_string());
+        }
+    }
+
+    let julia_channel_to_use: Option<String>;
+    let juliaup_channel_source: JuliaupChannelSource;
```
kdheepak, line 231: It's probably slightly more idiomatic to write the code this way:

```rust
    let julia_channel_to_use: Option<String>;
    let juliaup_channel_source: JuliaupChannelSource;

    // First check whether a channel is specified at the command line
    if channel_from_cmd_line.is_some() {
        julia_channel_to_use = channel_from_cmd_line;
        juliaup_channel_source = JuliaupChannelSource::CmdLine;
    } else {
        // Second check the JULIAUP_CHANNEL env var
        match std::env::var("JULIAUP_CHANNEL") {
            Ok(val) => {
                julia_channel_to_use = Some(val);
                juliaup_channel_source = JuliaupChannelSource::EnvVar;
            }
            Err(_) => {
                // Third use the default channel from the config file
                julia_channel_to_use = config_file.data.default.clone();
                juliaup_channel_source = JuliaupChannelSource::Default;
            }
        }
    }

    let julia_channel_to_use = julia_channel_to_use.ok_or_else(|| {
        anyhow!("The Julia launcher failed to figure out which juliaup channel to use.")
    })?;
```

And write it like this instead:

```rust
    let (julia_channel_to_use, juliaup_channel_source) = if let Some(channel) = channel_from_cmd_line {
        (channel, JuliaupChannelSource::CmdLine)
    } else if let Ok(channel) = std::env::var("JULIAUP_CHANNEL") {
        (channel, JuliaupChannelSource::EnvVar)
    } else if let Some(channel) = config_file.data.default.clone() {
        (channel, JuliaupChannelSource::Default)
    } else {
        return Err(anyhow!("The Julia launcher failed to figure out which juliaup channel to use."));
    };
```

PR #633, opened 2023-05-09
Add support for directory overrides

davidanthoff:
This adds an `override` command to `juliaup`:

- `juliaup override set lts`: Sets an override for the current directory to the `lts` channel.
- `juliaup override unset`: Removes an override for the current directory.
- `juliaup override status`: Shows all overrides.

There are some more options, see the command line help for those.


DilumAluthge:
How does this work? Does it store a file in the given directory? If so, can we document the name of that filev

PR #636, opened 2023-05-10
MSI Installer

davidanthoff:
Fallback option where MSIX doesn't work (Windows Server primarily).


davidanthoff:
Merging right away so that I can work on the release workflow integration.

PR #639, opened 2023-05-10
Host crate on Cargo: tag `julia` as binary name for `julialauncher`

### LINKED ISSUE #618: Host crate on Cargo as well
Following https://github.com/JuliaLang/juliaup/issues/615#issuecomment-1513926408, it would be great to be able to install using `cargo install juliaup`.  The rationale is that many systems (Linux distributions, in particular) have working Rust installations, and this would provide a convenient, hash-verified way to install on those platforms.

MilesCranmer:
This creates `julia` as another executable for `julialauncher` in the cargo bin folder. Thus allowing you to install `juliaup` from crates.io and have a working `julia` command.

Fixes #618. Note that `julialauncher` will still appear in the `bin` folder.

You can test this out by running:

```
cargo install --git https://github.com/MilesCranmer/juliaup.git --branch rust-pkg
```
and verifying that `juliaup` and `julia` are installed in the cargo binary folder.

@davidanthoff feel free to push to this PR


MilesCranmer:
@davidanthoff it might be worth adding https://github.com/marketplace/actions/publish-crates to automatically publish new updates to the `juliaup` crate

MilesCranmer:
If some of the deployments need to avoid installing the `julia` binary, you can use the `--bin` and `--bins` commands to specify which binaries are installed. See here: https://doc.rust-lang.org/cargo/commands/cargo-install.html#install-options. @davidanthoff should I add this as an option anywhere?

davidanthoff:
Great!

The whole CI workflow is very, very complicated, so probably easiest if I push to this branch to get that into working shape. Thanks so much for the start, though!

uncomfyhalomacro:
Just a quick question. Wouldn't it be better to just rename `julialauncher.rs` to `julia.rs` or am I missing something here? I feel like having two binaries with the same purpose are wasted duplicates. Maybe it's just me.

davidanthoff:
> Just a quick question. Wouldn't it be better to just rename `julialauncher.rs` to `julia.rs` or am I missing something here? I feel like having two binaries with the same purpose are wasted duplicates.

For the binaries that we ship via any of the other installers the current design has the benefit that one can distinguish between the launcher process and an actual Julia process in things like the task manager. But I think with this PR here we can keep the Juliaup `julia` binary really strictly to the cargo install method, and leave all the other deployments on `julialauncher` and `juliaup`.

davidanthoff:
All done now: https://crates.io/crates/juliaup/. New package versions should be automatically pushed whenever we release a build to the release channel. Thanks, @MilesCranmer!

PR #674, opened 2023-06-30
bitmaps for msi installer

### LINKED ISSUE #671: Custom bitmaps in MSI installer
We should customize the UI in the new MSI installer to match the general Julia branding. What needs to be done is to replace the two bmp files in https://github.com/JuliaLang/juliaup/tree/main/deploy/msi/Bitmaps with versions that have our branding.

Important: DON'T change the resolution of the files, they need to have exactly this size, also need to be bmp files. Also, leave the parts that are white at the moment white, other UI elements will be placed there. Here are two screenshots how these bitmaps look in the actual UI:

image

image

For anyone who wants to tackle this, just open a PR where those two files are modified :)

alecloudenback:
closes #671 
tiling taken from julialang.org (originally @cormullion?)

Github won't let me upload `.bmp`s for the preview here (though the commit is `.bmp`). I also have the Affinity Photo files that were the basis for these but I figured you didn't want them in this repo?

digbmp

bannerbmp


davidanthoff:
They look great!

Two small things: I think the dlgbmp.bmp has a different resolution now? Width seems to be 491 now but was (and should be) 493. And I think the background might have changed to gray, but ideally should be white?

alecloudenback:
Fixed the width. I also double checked: both new and original backgrounds are RGB(255,255,255)

davidanthoff:
Awesome, thanks!

PR #701, opened 2023-07-26
MacOS `uname -m` can lie due to Rosetta shenanigans

StefanKarpinski:
This uses sysctl instead, which is more trustworthy.


DilumAluthge:
Can you add a brief comment in the code that explains that we're doing this because of problems with Homebrew `uname` and Rosetta?

Issue #327, opened 2022-07-17
Directly use libjulia


davidanthoff:
I'm not suggesting we do this anytime soon, but eventually I think the `julialauncher` binary here could take the place of https://github.com/JuliaLang/julia/tree/master/cli? That way we would speed up startup times.

CC @staticfloat

staticfloat:
I don't think the `cli` piece adds much startup time.  I would suggest you measure it before you try to re-implement it, it does a lot of stuff (and the stuff it does changes from version to version as we add more features)

vchuravy:
It might be interesting implementing the cli piece in rust, especially if we are going to parse some TOML files to figure out which libraries to load.

fingolfin:
How much potential speed gain are we talking that would make it worth to have the extra complexity (and possibility for subtle bugs due to difference between the two implementation)?

In a quick experiment, the overhead seems neglible:
```
tutulla:~$ hyperfine -r 25 "julia +1.10 -e 'exit()'"
Benchmark 1: julia +1.10 -e 'exit()'
  Time (mean ± σ):     103.5 ms ±   2.4 ms    [User: 161.1 ms, System: 478.4 ms]
  Range (min … max):   101.8 ms … 113.9 ms    25 runs

```
versus
```
tutulla:~$ hyperfine -r 25 "$HOME/.julia/juliaup/julia-1.10.0+0.x64.linux.gnu/bin/julia -e 'exit()'"
Benchmark 1: /users/cip/alggeom/horn/.julia/juliaup/julia-1.10.0+0.x64.linux.gnu/bin/julia -e 'exit()'
  Time (mean ± σ):     101.6 ms ±   0.9 ms    [User: 165.7 ms, System: 474.4 ms]
  Range (min … max):    99.6 ms … 102.7 ms    25 runs
```

Maybe my test is a bad one?

davidanthoff:
I think the benefit might be a bit better on Windows? My understanding is that the overhead on Windows to launch a process is larger than on Linux/Mac?

PR #739, opened 2023-10-26
Truncate exit code to u8, even if the real code doesn't fit in u8.

### LINKED ISSUE #737: Julia crashes when using exit(-1)
minimal example:
```
exit(-1)
```

crash report toml:
```
name = "Juliaup launcher"
operating_system = "Windows 10.0.22621 (Windows 11 Professional) [64-bit]"
crate_version = "1.11.22"
explanation = """
Panic occurred in file 'src/bin/julialauncher.rs' at line 369
"""
cause = "called `Result::unwrap()` on an `Err` value: TryFromIntError(())"
method = "Panic"
backtrace = """

   0: 0x7ff7da484b21 - <unresolved>
   1: 0x7ff7da4723ef - <unresolved>
   2: 0x7ff7da4724c6 - <unresolved>
   3: 0x7ff7da48a8b7 - <unresolved>
   4: 0x7ff7da5c1c3c - <unresolved>
   5: 0x7ffab12c257d - BaseThreadInitThunk
   6: 0x7ffab23aaa78 - RtlUserThreadStart"""
```

installed julia using juliaup from windows store, versioninfo():
```
Julia Version 1.9.3
Commit bed2cd540a (2023-08-24 14:43 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 16 × AMD Ryzen 7 5800H with Radeon Graphics
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-14.0.6 (ORCJIT, znver3)
  Threads: 16 on 16 virtual cores
Environment:
  JULIA_NUM_THREADS = 16
```

jakobnissen:
This prevents JuliaLauncher from panicking if a user exits with an error code which doesn't fit in u8.
This is already the behaviour of Julia on Linux, where `ccall(:jl_exit)` can be called with Int32(-1), yet returns error code 255.

Closes #737

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -366,7 +366,9 @@ fn main() -> Result<std::process::ExitCode> {
         }
     }
 
-    let client_status: u8 = client_status?.try_into().unwrap();
+    // We truncate to the lower 8 bits, even though a Windows exit code is i32.
+    // this matches Julia's own behaviour on Unix
+    let client_status: u8 = client_status? as u8;
 
     Ok(std::process::ExitCode::from(client_status))
 }


StefanKarpinski:
Can we please change things around so that the Julia process is the parent and the updater is the child? The. This would be unnecessary as would all the logic for handling the termination of the Julia child process.

davidanthoff:
I have an alternative fix for the underlying problem in https://github.com/JuliaLang/juliaup/pull/750. @jakobnissen what do you think of that?

@StefanKarpinski I don't understand your suggestion :) The updater is not in the picture here at all, and is never the parent process. This here is the `julialauncher`, which is what starts when you type in `julia`. It is in charge of a) figuring out which Julia version to start, and b) will periodically launch the updater as a child process in the background.

StefanKarpinski:
https://github.com/JuliaLang/juliaup/issues/552:

> I think we should change julialauncher to use execve() (or its moral equivalent, wherever possible) instead of fork() when launching the julia sub-process. Because this does not launch a child process but instead causes the current process to be replaced by the new process

This would mean, among other benefits listed there, that `julialauncher` doesn't need to propagate the exit code because it becomes the Julia process which then terminates as it normally would. Having the real julia process be the child of the `julialauncher` process is just less good in basically every way.

davidanthoff:
Yeah, I'm just quite nervous about the interaction of `execve` and Rust packages, there seems to be a lot of potential for pain. And on Windows that won't ever work, and I'm also not keen to have such a massive difference there...

My sense right now is that the only way to really have a one-process solution that is architecturally similar on the various platforms would be some variant of https://github.com/JuliaLang/juliaup/issues/327. Which also seems anything but trivial...

Issue #151, opened 2021-09-27
shell completion


Roger-luo:
I'm not sure if the CLI generator can generate the shell completion directly in Rust (if there is one that would be better), we have shell completion auto generated via [Comonicon.jl](https://github.com/comonicon/Comonicon.jl) for IonCLI, I find this is quite convenient for terminal based people, so in case someone wants a script to start with for juliaup I wrap up a quick one here (for ZSH), I didn't open a PR because I'm not sure where would be best place to put it, but I think it would be nice if the shell completion can also be installed automatically while installing `juliaup`

In principle this shell completion can be improve to complete more specific things such as julia versions (at least in ZSH, I'm not sure about other shells), but I think this could be a start point.

```zsh
#compdef _juliaup juliaup
function _juliaup() {
    # These are set by _arguments
    local context state state_descr line
    typeset -A opt_args
    
    _arguments -C \
        '(- 1 *)'{-h,--help}'[Prints help information]' \
        '(- 1 *)'{-V,--version}'[Prints version information]' \
        "1: :((add\:'Add a specific Julia version or channel to your system'
                default\:'Set the default Julia version'
                gc\:'Garbage collect uninstalled Julia versions'
                help\:'Prints this message or the help of the given subcommand(s)'
                link\:'Link an existing Julia binary to a custom channel name'
                remove\:'Remove a Julia version from your system'
                status\:'Show all installed Julia versions'
                update\:'Update all or a specific channel to the latest Julia version'))" \
        "*:: :->args" \
}
```

davidanthoff:
Just a note to self, there is a crate that somehow helps creating these for clap, we should try to use that when we get to this.

PR #753, opened 2023-11-15
generate completions

BoostCookie:
EDIT:
This is done now. It solves https://github.com/JuliaLang/juliaup/issues/151. It works the same way as it does in rustup. https://rust-lang.github.io/rustup/installation/index.html

OLD MESSAGE:
> This is some starting effort that can eventually solve https://github.com/JuliaLang/juliaup/issues/151.
> I don't know how the build files are being packaged for the different distributions and operating systems. I'm asking for feedback in this regard.
> 
> What happens now:
> Every time the package is build the completions for the different shells are generated and put into the build directory `target/release/build/juliaup-<somehexstring>/out/`.
> 
> What is left to do:
> Add the completions to the packages so that they are available after installing juliaup.

### DIFF
--- a/src/bin/juliaup.rs
+++ b/src/bin/juliaup.rs
@@ -2,6 +2,7 @@ use anyhow::{Context, Result};
 use clap::Parser;
 use juliaup::cli::{ConfigSubCmd, Juliaup, OverrideSubCmd, SelfSubCmd};
 use juliaup::command_api::run_command_api;
+use juliaup::command_completions::run_command_completions;
 #[cfg(not(windows))]
 use juliaup::command_config_symlinks::run_command_config_symlinks;
 use juliaup::command_config_versionsdbupdate::run_command_config_versionsdbupdate;
@@ -111,5 +112,6 @@ fn main() -> Result<()> {
             #[cfg(not(feature = "selfupdate"))]
             SelfSubCmd::Uninstall {} => run_command_selfuninstall_unavailable(),
         },
+        Juliaup::Completions { shell } => run_command_completions(&shell),
     }
 }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -246,6 +246,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "clap_complete"
+version = "4.4.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "97aeaa95557bd02f23fbb662f981670c3d20c5a26e69f7354b28f57092437fcd"
+dependencies = [
+ "clap",
+]
+
+[[package]]
 name = "clap_derive"
 version = "4.4.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -968,6 +977,7 @@ dependencies = [
  "built",
  "chrono",
  "clap",
+ "clap_complete",
  "cli-table",
  "cluFlock",
  "console",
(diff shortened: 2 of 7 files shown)


davidanthoff:
I'm not really familiar with how completion files work. Do we need to drop them into a specific location on user systems? Or do they need to be run? Or something else?

BoostCookie:
> I'm not really familiar with how completion files work. Do we need to drop them into a specific location on user systems? Or do they need to be run? Or something else?

Packages usually install their zsh completions to `/usr/share/zsh/site-functions/`. If you are an admin that installs stuff manually you should use `/usr/local/` instead of `/usr/` of course.
Similarly bash completions go into `/usr/share/bash-completion/completions/`.

So OS repositories (Arch User Repository, openSUSE Tumbleweed, Solus) should definitely package the completions into these directories.

For the "recommended" install method
```
curl -fsSL https://install.julialang.org | sh
```
we are only the local user. We could put the completions into `$JULIA_DEPOT_PATH` and then do something similar like here https://github.com/JuliaLang/juliaup/blob/a9d67559c42ddf87598dc80c808674d2381fa623/src/operations.rs#L790 where we add the binary to the `$PATH`. We can also add our completions directory to the specific shell completions.

Though personally I'm not a fan in fishing around in people's config files. I think we should do it the way rustup does it.
https://rust-lang.github.io/rustup/installation/index.html
With a command
```
$ juliaup completions bash > ~/.local/share/bash-completion/completions/juliaup
```
for bash, for example.

BoostCookie:
I have now followed the rustup way, and added a command that generates the completions.

src/bin/juliaup.rs
```diff
@@ -27,128 +29,6 @@ use juliaup::{
 };
 use log::info;
 
-#[derive(Parser)]
```
fingolfin, line 30: Disclaim: I am not a maintainer of this repository, so feel free to ignore me. This advice is just based on my experience maintaining other projects. If this PR was one on my repository, I'd be reluctant to review and merge it as it is somewhat opaque.

Specifically: Do you *need* to move these definitions to a new file? If not, I would recommend against doing it, at least in this PR, as it makes it that much harder to review it, as it is not obvious if anything (and what) was changed in this code block other than moving it.


If the code must be split out, I would recommend doing it in two commits: the first commit just moves the code to a new file, but without otherwise modifying it. The second commit than does the actual work and modifications.

BoostCookie: Thanks for your feedback.
Unfortunately I think separating the files is necessary, otherwise the `command_completions.rs` cannot access the `enum Juliaup`.

I've now separated the commits like you recommended.

fingolfin:
@BoostCookie unfortunately it seems this now as conflicts that must be resolved?

BoostCookie:
Is there even any interest in this pull request?
@davidanthoff

StefanKarpinski:
Yes, this is great functionality. @davidanthoff is quite busy, but please don't be discouraged. We'll get this merged.

davidanthoff:
We had more changes to the clap structures on `main`, and I'm a bit worried that those changes might get lost in the simultaneous move of the code to a new file in this PR. So, I just created a new PR off `main` that _just_ moves the clap structs to a new file, like in this PR. I'm going to merge that right away, and then we can be entirely sure that we don't lose anything in the merge story.

@BoostCookie could you rebase this PR on top of `main` once I've merged https://github.com/JuliaLang/juliaup/pull/787?

Awesome PR, thanks!

davidanthoff (approved):

PR #767, opened 2023-12-10
make julialauncher have saner error messages for channels and versions

### LINKED ISSUE #766: "ERROR: Invalid Juliaup channel `1.0` at command line."
This channel isn't really "invalid", I just hadn't installed it yet. I think a simple fix would be "Invalid or uninstalled Julia channel". A fancier fix would be to lookup if it is just not yet installed, and if running interactively, prompt to do so, given we are on an error path anyway and can afford the lookup time.

uncomfyhalomacro:
closes #766

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -165,23 +165,23 @@ fn get_julia_path_from_channel(
             .ok_or_else(|| match juliaup_channel_source {
                 JuliaupChannelSource::CmdLine => {
                     if versions_db.available_channels.contains_key(channel) {
-                        UserError { msg: format!("`{}` is not installed. Please run `juliaup add {}` to install channel or version", channel, channel) }
+                        UserError { msg: format!("`{}` is not installed. Please run `juliaup add {}` to install channel or version.", channel, channel) }
                     } else {
-                        UserError { msg: format!("ERROR: Invalid Juliaup channel `{}`. Please run `juliaup list` to get a list of valid channels and versions",  channel) }
+                        UserError { msg: format!("ERROR: Invalid Juliaup channel `{}`. Please run `juliaup list` to get a list of valid channels and versions.",  channel) }
                     }
                 }.into(),
                 JuliaupChannelSource::EnvVar=> {
                     if versions_db.available_channels.contains_key(channel) {
-                        UserError { msg: format!("`{}` for environment variable JULIAUP_CHANNEL is not installed. Please run `juliaup add {}` to install channel or version", channel, channel) }
+                        UserError { msg: format!("`{}` for environment variable JULIAUP_CHANNEL is not installed. Please run `juliaup add {}` to install channel or version.", channel, channel) }
                     } else {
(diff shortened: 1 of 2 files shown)


uncomfyhalomacro:
\cc @davidanthoff ready for review. if youre free :D

tests/channel_selection.rs
```diff
@@ -99,7 +99,7 @@ fn channel_selection() {
         .env("JULIAUP_DEPOT_PATH", depot_dir.path())
         .assert()
         .failure()
-        .stderr("ERROR: Invalid Juliaup channel `1.8.6` at command line.\n");
+        .stderr("ERROR: Invalid Juliaup channel `1.8.6`. Please run `juliaup list` to check a valid channel or version\n");
```
fingolfin, line 102: This reads a bit weird to me: "to check a valid channel or version".

Maybe "to find a valid channel or version" or "to get a list of valid channels and versions" ?

uncomfyhalomacro:
very late to you and contributors but happy new year! 🎉

@fingolfin thanks for the code review. I applied your suggestions. 🫂

src/bin/julialauncher.rs
```diff
@@ -159,13 +159,40 @@ fn get_julia_path_from_channel(
     juliaupconfig_path: &Path,
     juliaup_channel_source: JuliaupChannelSource,
 ) -> Result<(PathBuf, Vec<String>)> {
+    let paths = get_paths().with_context(|| "Trying to load all global paths.")?;
+    let versiondb_data =
+        load_versions_db(&paths).with_context(|| "command failed to load versions db.")?;
```
davidanthoff, line 164: I don't think we need to load it again, the function has an argument `versions_db` that can be used?

src/bin/julialauncher.rs
```diff
@@ -159,13 +159,40 @@ fn get_julia_path_from_channel(
...
             .installed_channels
             .get(channel)
             .ok_or_else(|| match juliaup_channel_source {
-                JuliaupChannelSource::CmdLine => UserError { msg: format!("ERROR: Invalid Juliaup channel `{}` at command line.", channel).to_string() }.into(),
-                JuliaupChannelSource::EnvVar => UserError { msg: format!("ERROR: Invalid Juliaup channel `{}` in environment variable JULIAUP_CHANNEL.", channel).to_string() }.into(),
-                JuliaupChannelSource::Override => UserError { msg: format!("ERROR: Invalid Juliaup channel `{}` in directory override.", channel).to_string() }.into(),
+                JuliaupChannelSource::CmdLine => {
+                    if versiondb_data.available_channels.contains_key(channel) {
+                        UserError { msg: format!("`{}` is not installed. Please run `juliaup add {}` to install channel or version", channel, channel) }
+                    } else if versiondb_data.available_versions.contains_key(channel) {
```
davidanthoff, line 172: I don't think we need to check in the `availabe_versions`, users can't access those versions in any, I think just looking at `available_channels` is enough. Same applies to the same `if` blocks below.

davidanthoff (changes requested):
Nice! Just a few small change requests.

davidanthoff:
Ok, one (very!) minor last change request: could you add a period at the end of each of the error messages? That is the convention I have used throughout the project.

davidanthoff (approved):
Awesome, thanks!

PR #768, opened 2023-12-14
Init script: Added possiblity for overriding JULIAUP_VERSION

stemann:
This PR should make it possible to customize the juliaup version used, e.g., to lock the juliaup version used for reproducibility in a Dockerfile.


davidanthoff:
Do we really want this? My original idea for Juliaup was really that it would always be on the latest version for everyone, and that reproducability would really only come in for Julia itself. We also don't really publish a list of available versions anywhere, or make any guarantees that old versions will hang around.

I can see that this might be useful if we ever make a breaking change in Juliaup...

DilumAluthge:
This PR has not been updated in over one year.

@stemann Are you still interested in working on this PR?

PR #769, opened 2023-12-15
improvement: handling of selfuninstall

### LINKED ISSUE #752: Provide better UI for `self uninstall` on platforms that don't actually support it
Right now `juliaup self uninstall` is an error as "unknown command" on platforms where Juliaup is installed via some third party method (brew, cargo etc). Instead of just pretending that the command doesn't exist, we could either output a more useful error message, or maybe on some platforms even initiate the uninstall (similar to Windows, where `juliaup self update` triggers a Windows Store update for Juliaup).

uncomfyhalomacro:
We added the feature `distro-default` to let packagers configure the disablement of `selfupdate`. The build.rs file is adapted to raise a compiler error if both features are configured together.

Closes #752

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -90,6 +90,7 @@ predicates = "3.0.3"
 
 [features]
 selfupdate = []
+distro-default = []
 windowsstore = []
 windowsappinstaller = []
 dummy = []
--- a/build.rs
+++ b/build.rs
@@ -62,4 +62,3 @@ fn main() -> Result<()> {
 
 #[cfg(all(feature = "selfupdate", feature = "distro-default"))]
 compile_error!("Compilation is not allowed. You can't compile with feature `selfupdate` and `distro-default` enabled together.");
-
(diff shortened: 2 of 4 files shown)


uncomfyhalomacro:
Sample running commands:
1. The first command is compiling juliaup with both features turned on. It should raise a compilation error.
2. The second command is running `juliaup self uninstall` built with `-F distro-default`.

image

davidanthoff:
Do we really need a new feature here? Can't we just have some clear logic to detect when we need to output this based on the existing features?

uncomfyhalomacro:
let me see if i can still find a way to not add a new feature. it's hard since some of the options are tied with `selfupdate`

davidanthoff:
I just took another look, I _think_ it could just be `#[cfg(not(feature = "selfupdate"))]`, right?

uncomfyhalomacro:
yeah im going to try that. i dont remember why but i feel like i made a mess somewhere for using cfg(not(...))

uncomfyhalomacro:
cc @davidanthoff new PR at https://github.com/JuliaLang/juliaup/pull/786. :D

PR #782, opened 2024-01-06
Fix usage of drop

fingolfin:
Fixes #778

PR #783, opened 2024-01-06
Replace unmaintained action-rs actions

fingolfin:
The action-rs repositories were archived in October 2023. They use deprecated GH Action feature which will stop working at some point.

Replace `actions-rs/toolchain` by `actions-rust-lang/setup-rust-toolchain` which is different but similar enough for our purposes. It lacks the `default` argument but as far as I can tell it does not matter?

As a nice enhancement, the new action also sets up "Problem Matchers" which scan all output for `cargo` and `rustfmt` error messages which then are highlighted in the GitHub web interface.

That also means that as a replacement for `actions-rs/cargo` one can usually just call `cargo` directly. Well, except when `use-cross` is enabled... One could probably also call `cross` directly, but I decided to play it safe and just use `clechasseur/rs-cargo` instead, which is an improved fork of the original `actions-rs/cargo`.

Requires #782 -- so e.g. close and reopen this PR after #782 was merged to (hopefully) make it pass

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -191,13 +191,12 @@ jobs:
       with:
         platform: x86
         version: 12.2.0
-    - uses: actions-rs/toolchain@v1
+    - uses: actions-rust-lang/setup-rust-toolchain@v1
       with:
         toolchain: ${{matrix.toolchain}}
-        default: true
         target: ${{matrix.target}}
     - name: Release build
-      uses: actions-rs/cargo@v1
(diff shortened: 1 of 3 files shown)


davidanthoff:
> It lacks the `default` argument but as far as I can tell it does not matter?

Yes, I think [this](https://github.com/actions-rust-lang/setup-rust-toolchain/blob/d5fcac478a8270ade8a005df33017ae06891af79/action.yml#L141) means that if we install a toolchain with this action it always is made the default.

PR #786, opened 2024-01-12
improvement: new message for `juliaup self uninstall command` if feature `selfupdate` is disabled

### LINKED ISSUE #752: Provide better UI for `self uninstall` on platforms that don't actually support it

uncomfyhalomacro:
supersedes #769 and fixes #752 

---

Off-topic but I may have found a bug. Not sure if the location of the binaries are hard-coded to the julia home folder to cause this. it should be `target/debug/<juliaup binaries>`

```
Do you really want to uninstall Julia? yes
Removing background self update task. Failed.
Removing startup self update configuration. Failed.
Removing PATH modifications in startup scripts. Failed.
Removing symlinks. Failed.
Deleting Juliaup home folder "/home/uncomfy/.julia/juliaup". Success.
Deleting julia symlink "/home/uncomfy/development/projects/juliaup/target/bin/julia". Failed.
Deleting julialauncher binary "/home/uncomfy/development/projects/juliaup/target/bin/julialauncher". Failed.
Deleting juliaup binary "/home/uncomfy/development/projects/juliaup/target/bin/juliaup". Failed.
Error: No such file or directory (os error 2)
```

### DIFF
--- a/src/bin/juliaup.rs
+++ b/src/bin/juliaup.rs
@@ -23,8 +23,15 @@ use juliaup::{
     command_config_backgroundselfupdate::run_command_config_backgroundselfupdate,
     command_config_modifypath::run_command_config_modifypath,
     command_config_startupselfupdate::run_command_config_startupselfupdate,
-    command_selfchannel::run_command_selfchannel, command_selfuninstall::run_command_selfuninstall,
+    command_selfchannel::run_command_selfchannel,
 };
+
+#[cfg(feature = "selfupdate")]
+use juliaup::command_selfuninstall::run_command_selfuninstall;
+
+#[cfg(not(feature = "selfupdate"))]
+use juliaup::command_selfuninstall::run_command_selfuninstall_unavailable;
+
 use log::info;
 
 #[derive(Parser)]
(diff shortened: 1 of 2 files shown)


src/bin/juliaup.rs
```diff
@@ -109,6 +116,11 @@ enum SelfSubCmd {
     #[cfg(feature = "selfupdate")]
     /// Uninstall this version of juliaup from the system
     Uninstall {},
+    #[cfg(not(feature = "selfupdate"))]
+    /// If `selfupdate` feature is not enabled, this
+    /// configures the Uninstall to just print an error message
+    /// that it is **unavailable**.
```
davidanthoff, line 122: I believe the text after the tripple `/` will be output as the help text in the compiled version, so it should address end-users, not devs looking at the source code :) So maybe just `Uninstall this version of juliaup from the system (UNAVAILABLE)` or something like that?

uncomfyhalomacro: Oh i didnt know the comment is used for the help text. Resolved

davidanthoff:
Great, thanks!

PR #791, opened 2024-01-14
Apply suggestions made by `cargo clippy --fix --lib` and `cargo clippy --fix` and some more

fingolfin:


build.rs
```diff
@@ -24,7 +24,7 @@ fn main() -> Result<()> {
         .join(format!("versiondb-{}.json", target_platform));
 
     let version_db_path = out_path.join("versionsdb.json");
-    std::fs::copy(&db_path, &version_db_path).unwrap();
+    std::fs::copy(&db_path, version_db_path).unwrap();
```
davidanthoff, line 27: Do we understand why it is suggesting this change here? That seems odd to me...

src/bin/julialauncher.rs
```diff
@@ -225,7 +225,7 @@ fn get_julia_path_from_channel(
                         juliaupconfig_path.display()
                     )
                 })?;
-            return Ok((absolute_path.into_path_buf(), Vec::new()));
+            Ok((absolute_path.into_path_buf(), Vec::new()))
```
davidanthoff, line 228: Hm, also not keen on this one here, to me the `return` style is easier to read and understand?

src/operations.rs
```diff
@@ -71,12 +71,13 @@ pub fn download_extract_sans_parent(
             .progress_chars("=> "),
     );
 
-    let foo = pb.wrap_read(response);
+    let response_with_pb = pb.wrap_read(response);
 
-    let tar = GzDecoder::new(foo);
+    let tar = GzDecoder::new(response_with_pb);
```
davidanthoff, line 76: That is pretty cool, did it do this automatically?

davidanthoff (commented):
Hm, so some of these changes I don't understand the rational for. Why does it want to pass ownership into all these functions when we call them?

src/operations.rs
fingolfin: No that was me in the second commit ;-)

davidanthoff:
Probably ok to close this for now? If not, please reopen!

PR #798, opened 2024-01-19
Some minor cleanup

fingolfin:
This is a small subset of the changes in #791 in particular it includes all the "manual" changes I made there. I hope this is less controversial and can just be merged.

(While here, may I ask if we could get a new juliaup release soon? I am eager to try out the new forking code, and verify that it fixes `ctrl-Z` for me ;-) )

### DIFF
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -71,12 +71,13 @@ pub fn download_extract_sans_parent(
             .progress_chars("=> "),
     );
 
-    let foo = pb.wrap_read(response);
+    let response_with_pb = pb.wrap_read(response);
 
-    let tar = GzDecoder::new(foo);
+    let tar = GzDecoder::new(response_with_pb);
     let archive = Archive::new(tar);
     unpack_sans_parent(archive, target_path, levels_to_skip)
         .with_context(|| format!("Failed to extract downloaded file from url `{}`.", url))?;
+
     Ok(())
 }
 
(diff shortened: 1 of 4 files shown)


src/operations.rs
```diff
@@ -154,13 +155,11 @@ pub fn download_extract_sans_parent(
...
-    let foo = pb.wrap_read(DataReaderWrap(reader));
-
-    let tar = GzDecoder::new(foo);
+    let response_with_pb = pb.wrap_read(DataReaderWrap(reader));
 
+    let tar = GzDecoder::new(response_with_pb);
     let archive = Archive::new(tar);
-
-    unpack_sans_parent(archive, &target_path, levels_to_skip)
+    unpack_sans_parent(archive, target_path, levels_to_skip)
```
LilithHafner, line 162: I don't see any particular reason to transfer ownership here if a reference works just as well, but still happy to merge this part of the diff if you think it's better for stylistic reasons.

LilithHafner (approved):
LGTM, My one comment is optional, feel free to accept or decline.

I would merge, but I'm not sure of a repo-wide merge-block is in effect @davidanthoff

src/operations.rs
davidanthoff: I agree, I actually prefer a ref here. But also, it really doesn't matter very much, so I'll merge this as soon as the merge window opens again.

fingolfin: Just to explain: I merely wanted the two copies of code calling `unpack_sans_parent` to match as closely as possible, so that future edits don't have to wonder "why are these different?". Since I had to decide with two options, I used what `cargo clippy` considers as "best practice".

PR #828, opened 2024-02-09
document -p

gdkrmr:
I was using `juliaup` to install `julia` in a docker container (install as root in `/opt/juliaup` to be used by different user that is mapped into the container at deployment time) and had to download and look at the shellscript to learn about the `-p` option. I think this option should be documented in the `README.md`.


davidanthoff (approved):
Thanks!

PR #836, opened 2024-02-14
Short channel matching for julia command

rakeshksr:
Closes #830
This pull request introduces the feature of short channel matching for the Julia command. Users can now launch alpha, beta, LTS, release, and nightly versions by specifying their respective short names: "a" for alpha, "b" for beta, "l" for LTS, "r" for release, and "n" for nightly.

PR #838, opened 2024-02-14
Alternate short channel matching support

### LINKED ISSUE #830: Support short channel matching
On julia you can do `--start=no` to control `--startup-file=no` because there's no other arg that matches `--start`

It would be nice if juliaup did the same for channel names

So `julia +n` for nightly. `julia +r` for release. etc.

christiangnrd:
Closes #836 and closes #830.

If the channel is the prefix for one and only one installed channel, use that, otherwise old behaviour. This behaves more like the julia command line arguments

Should work with any installed channel including linked channels instead of a hard-coded few.

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -168,40 +168,71 @@ enum JuliaupChannelSource {
     Default,
 }
 
+fn only<T, I>(mut iter: I) -> Option<T>
+where
+    I: Iterator<Item = T>,
+{
+    let first = iter.next();
+    if iter.next().is_some() {
+        // There are multiple elements
+        return None;
+    }
+    first
+}
+
 fn get_julia_path_from_channel(
     versions_db: &JuliaupVersionDB,
     config_data: &JuliaupConfig,
-    channel: &str,
+    requested_channel: &str,
     juliaupconfig_path: &Path,
     juliaup_channel_source: JuliaupChannelSource,
 ) -> Result<(PathBuf, Vec<String>)> {
-    let channel_valid = is_valid_channel(versions_db, &channel.to_string())?;
+    let expanded_channel = if requested_channel
+        .chars()
+        .next()
+        .map(char::is_numeric)
+        .unwrap_or(false)
+    {
+        requested_channel
+    } else {
+        match only(
+            config_data
+                .installed_channels
+                .keys()
+                .filter(|&item| item.starts_with(requested_channel)),
+        ) {
+            Some(val) => val,
+            None => requested_channel,
+        }
+    };
+
+    let channel_valid = is_valid_channel(versions_db, &expanded_channel.to_string())?;
     let channel_info = config_data
             .installed_channels
-            .get(channel)
+            .get(expanded_channel)
             .ok_or_else(|| match juliaup_channel_source {
                 JuliaupChannelSource::CmdLine => {
                     if channel_valid {
-                        UserError { msg: format!("`{}` is not installed. Please run `juliaup add {}` to install channel or version.", channel, channel) }
+                        UserError { msg: format!("`{}` is not installed. Please run `juliaup add {}` to install channel or version.", expanded_channel, expanded_channel) }
                     } else {
-                        UserError { msg: format!("Invalid Juliaup channel `{}`. Please run `juliaup list` to get a list of valid channels and versions.",  channel) }
+                        UserError { msg: format!("Invalid Juliaup channel `{}`. Please run `juliaup list` to get a list of valid channels and versions.",  expanded_channel) }
                     }
                 }.into(),
                 JuliaupChannelSource::EnvVar=> {
                     if channel_valid {
-                        UserError { msg: format!("`{}` from environment variable JULIAUP_CHANNEL is not installed. Please run `juliaup add {}` to install channel or version.", channel, channel) }
+                        UserError { msg: format!("`{}` from environment variable JULIAUP_CHANNEL is not installed. Please run `juliaup add {}` to install channel or version.", expanded_channel, expanded_channel) }
                     } else {
-                        UserError { msg: format!("Invalid Juliaup channel `{}` from environment variable JULIAUP_CHANNEL. Please run `juliaup list` to get a list of valid channels and versions.",  channel) }
+                        UserError { msg: format!("Invalid Juliaup channel `{}` from environment variable JULIAUP_CHANNEL. Please run `juliaup list` to get a list of valid channels and versions.",  expanded_channel) }
                     }
                 }.into(),
                 JuliaupChannelSource::Override=> {
                     if channel_valid {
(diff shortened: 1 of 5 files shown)


LilithHafner:
I don't like this because I could have Julia 1.6.2 as the only installed version and then run `julia +1` and get that version. Similarly, I could have Julia 1.10 installed and run `julia +1.1` and get version 1.10 instead.

The reasons this short matching works okay in Julia (and some other programs') command matching is that
- Their options are intelligently named to avoid misleading prefixing
- Their options are words or phrases that have a more diverse prefix set than version numbers

Whereas in this use case,
- Juliaup's set of version options varies depending on the installation (if we are filtering to installed channels)
- Juliaup's option names are already very concise so the benefit is less.

christiangnrd:
I've updated to not match on version numbers. The approach feels hacky but also I don't know if it needs t be more complicated. 

Unfortunately the new tests end up downloading 2 extra Julia versions so that might take up a lot more time. Maybe I should test with channels linked to already downloaded versions?

src/bin/julialauncher.rs
```diff
@@ -166,32 +166,49 @@ fn get_julia_path_from_channel(
     juliaupconfig_path: &Path,
     juliaup_channel_source: JuliaupChannelSource,
 ) -> Result<(PathBuf, Vec<String>)> {
+    let potential_matches: Vec<_> = config_data
+        .installed_channels
+        .keys()
+        .filter(|&item| {
+            item.starts_with(channel) && !item.starts_with("0") && !item.starts_with("1")
```
LilithHafner, line 173: This is more general:
```suggestion
            item.starts_with(channel) && !item.chars().next().map(char::is_numeric).unwrap_or(false)
```

src/bin/julialauncher.rs
```diff
@@ -166,32 +166,49 @@ fn get_julia_path_from_channel(
...
+        .cloned()
+        .collect();
+    let actual_channel = if potential_matches.len() == 1 {
+        potential_matches
+            .first()
+            .expect("length should be 1")
+            .as_str()
+    } else {
+        channel
+    };
```
LilithHafner, line 184: A more efficient solution (though probably negligibly so)

```suggestion
    let actual_channel = if channel
        .chars()
        .next()
        .map(char::is_numeric)
        .unwrap_or(false)
    {
        channel
    } else {
        match only(
            config_data
                .installed_channels
                .keys()
                .filter(|&item| item.starts_with(channel)),
        ) {
            Some(val) => val,
            None => channel,
        }
    };
```

With a helper function

```rs
fn only<T, I>(mut iter: I) -> Option<T>
where
    I: Iterator<Item = T>,
{
    let first = iter.next();
    if iter.next().is_some() {
        // There are multiple elements
        return None;
    }
    first
}
```

Mostly this is a stylistic choice.

src/bin/julialauncher.rs
```diff
@@ -166,32 +166,49 @@ fn get_julia_path_from_channel(
...
+
     let channel_info = config_data
             .installed_channels
-            .get(channel)
+            .get(actual_channel)
```
LilithHafner, line 188: `channel`/`actual_channel` is not a good pair of variable names, imo.

`requested_channel`/`expanded_channel` could be a better choice.

tests/channel_selection.rs
```diff
@@ -125,4 +125,63 @@ fn channel_selection() {
         .assert()
         .failure()
         .stderr("ERROR: Invalid Juliaup channel `1.8.6`. Please run `juliaup list` to get a list of valid channels and versions.\n");
+
+    // Now testing short channel matching
+    // At this point, installed channels are: 1.6.7, 1.7.3, 1.8.5
+
+    // Test that incomplete number matching does not autocomplete:
+    // https://github.com/JuliaLang/juliaup/pull/838#issuecomment-2206640506)
```
LilithHafner, line 133:
```suggestion
    // https://github.com/JuliaLang/juliaup/pull/838#issuecomment-2206640506
```

tests/channel_selection.rs
```diff
@@ -125,4 +125,63 @@ fn channel_selection() {
...
+    Command::cargo_bin("julia")
+        .unwrap()
+        .arg("+1.8")
+        .arg("-e")
+        .arg("print(VERSION)")
+        .env("JULIA_DEPOT_PATH", depot_dir.path())
+        .env("JULIAUP_DEPOT_PATH", depot_dir.path())
+        .env("JULIAUP_CHANNEL", "1.7.3")
```
LilithHafner, line 141:
```suggestion
```
I don't think we should set this env variable.

tests/channel_selection.rs
```diff
@@ -125,4 +125,63 @@ fn channel_selection() {
...
+    // Now testing short channel matching
+    // At this point, installed channels are: 1.6.7, 1.7.3, 1.8.5
+
+    // Test that incomplete number matching does not autocomplete:
+    // https://github.com/JuliaLang/juliaup/pull/838#issuecomment-2206640506)
+    Command::cargo_bin("julia")
+        .unwrap()
+        .arg("+1.8")
+        .arg("-e")
+        .arg("print(VERSION)")
```
LilithHafner, line 138: These are also unnecessary

LilithHafner (commented):
This works; I left a few comments, but fundamentally it's a design choice between this and #836. Do we want `+r` to always be `release`, or to be either `rc` or `release` depending on which channels the use has installed. I lean toward the simplicity of #836.

tests/channel_selection.rs
christiangnrd: This might be a stupid question but how does the test not get stuck in the REPL without the "-e"?

christiangnrd:
I've addressed your feedback and implemented all your suggestions.

Regarding this PR or #836, I wrote this because I sometimes write more than just the first letter, and I want this feature to work with linked channels. Also, for someone like me who sometimes tries out PRs but rarely more than one at a time, once #903 is a thing, it'll be nice to be able to call `julia +pr` instead of having to remember the pull request number.

tests/channel_selection.rs
LilithHafner: Good point; that is necessary (or -v or something)

christiangnrd: I was asking because the tests pass without them and I was genuinely wondering what made it work. I'll add -v to be safe because though.

LilithHafner:
@rakeshksr @christiangnrd and espeically @davidanthoff, could we discuss which option we prefer (#838 vs #836) both are imo ready to merge but we need to pick which.

christiangnrd:
My justification for preferring this approach are in my previous comment. To add to that, I always have a `dev` channel linked to my local branch of Julia, I feel like it would be quite unintuitive to have `+b` work for beta, and `+l` work for lts, but `+d` error out for dev.

davidanthoff:
I think I've finally am getting clarity what I'm thinking about this, see https://github.com/JuliaLang/juliaup/issues/830.

Issue #879, opened 2024-03-27
`juliainstaller` up fails when `juliaup` fails to execute


t-bltg:
Consider the following wrapper when `$HOME/bin` is added to the `PATH`:
```bash
$ cat $HOME/bin/juliaup
#!/usr/bin/env bash
exec __some_invalid_path__/.juliaup/bin/juliaup "$@"
$ chmod +x $HOME/bin/juliaup
$ juliaup
$HOME/bin/juliaup: line 2: __some_invalid_path__/.juliaup/bin/juliaup: No such file or directory
$ echo $?
127
```

Then, when invoking the following command to install `juliaup`
```bash
$ curl -fsSL https://install.julialang.org | sh -s -- -y
```

, the function [is_juliaup_installed](https://github.com/JuliaLang/juliaup/blob/main/src/bin/juliainstaller.rs#L102-L113) returns success (which is a false positive, since the wrapper tries to execute an invalid command, error 127) leading to the following message):
```
It seems that Juliaup is already installed on this system. Please remove the previous installation of Juliaup before you try to install a new version.
```

PR #880, opened 2024-03-27
fix `is_juliaup_installed` for invalid path

t-bltg:
This PR:

- fixes https://github.com/JuliaLang/juliaup/issues/879;
- prints the Juliaup configuration file, if existing;
- ~~changes the `--add_to_path` description (`juliainstaller --help`) to match the command line argument `--add-to-path` (underscore -> dash)~~.

### DIFF
--- a/src/bin/juliainstaller.rs
+++ b/src/bin/juliainstaller.rs
@@ -132,7 +132,7 @@ struct Juliainstaller {
     #[clap(short = 'p', long = "path")]
     alternate_path: Option<String>,
     /// Control adding the Juliaup dir to the PATH
-    /// Use Option<bool> and default_value="yes" to force --add-to-path=[yes|no|0|1] instead of flag
+    /// Use Option<bool> and default_value="yes" to force --add_to_path=[yes|no|0|1] instead of flag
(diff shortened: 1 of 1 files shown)


t-bltg:
@davidanthoff, bump ?

StefanKarpinski (commented):
I am not a Rust programmer but this looks reasonable to me. If no feedback in a week, I will merge.

davidanthoff:
Hm, I'm not sure this is a good idea. If the user has a script `juliaup` on the `PATH` then things are really broken already, and I'm almost thinking we should just abort install at that point. What would be good is a better error message, for example something like "Your system has a broken `juliaup` command on the `PATH`, please first remove that" or something along those lines.

Could you move the unrelated typo fix to a different PR? Thanks!

t-bltg:
> If the user has a script juliaup on the PATH then things are really broken already, and I'm almost thinking we should just abort install at that point

Exactly. Currently, `juliaup` returns a false positive as exposed in https://github.com/JuliaLang/juliaup/issues/879, and is thus a bug. This PR corrects this behaviour.

t-bltg:
> Could you move the unrelated typo fix to a different PR?

Done in https://github.com/JuliaLang/juliaup/pull/916.

davidanthoff (approved):

Issue #844, opened 2024-02-20
A path to XDG compliance


tecosaur:
In a recent Slack conversation, I offered to help answer questions about the XDG Base Directories spec. and the `~/.local/bin` directory, as I've previously looked into this extensively over the development of [BaseDirs.jl](https://tecosaur.github.io/BaseDirs.jl/stable/others/).

There are a number of problems that result from needing to modify shell initialisation scripts to extend `PATH` on Linux. The conversation on Slack started with one, somebody needing to give `juliaup` write permission to their shell config in order to successfully install. Others are noted in [this comment of mine](https://github.com/JuliaLang/Pkg.jl/pull/3772#issuecomment-1917004210) in a Pkg.jl PR.

@davidanthoff mentioned some of the reservations he had with the idea as a lack of clarity, and a seeming need for uncomfortable heuristics. To this, I produced the following response on how the `~/.local/bin` directory could be well-handled, and am copying it here (at David's request) for posterity.

-----

Completely get the heuristics making you uncomfortable. With Mac, everything they do define is well-defined, they just leave a bunch of blindspots, and those blind spots end up being a bit like the wild west. Linux is where I'm most familiar by far though. I'm familiar with the issue you mention regarding `~/.local/bin` only being listed on PATH when it already exists. This is usually controlled by `/etc/profile`, which often includes something like this:

```
PATH=/usr/local/bin:/usr/bin:/bin
if test "$HOME" != "/" ; then
for dir in $HOME/bin/$CPU $HOME/bin $HOME/.local/bin/$CPU $HOME/.local/bin ; do
    test -d $dir && PATH=$dir:$PATH
done
fi
```

In this case, the way to confirm that `~/.local/bin`* will be on the `PATH` (short of ugly heuristics that rely on reading the content of profile files) would be to:

- If `~/.local/bin` does not exist and is not on `PATH`
- Create `~/.local/bin` (this is in accordance with the XDG spec's instructions on what to do when a given XDG_*_HOME does not exist)
- Spawn a new (login*) instance of `$SHELL`, and fetch its `PATH` value
- Test to see `$HOME/.local/bin` is on the `PATH` from the previous step
- The vast majority of the time, at this point you will have identified that `~/.local/bin` is indeed on the `PATH`.
[...]

PR #886, opened 2024-04-05
canonicalize the current_exe result when self updating, so that it works with symlinking.

### LINKED ISSUE #885: `juliaup` doesn't correctly resolve symlinks when self updating on macOS
Hello,

installing `juliaup` through the `sh` script, and then symlinking the `juliaup` binary breaks correct autoreferentiality of the executable. `let my_own_path = std::env::current_exe()` does not follow symlinks (at least on macOS).

running `juliaup self update` incurs in
```
Error: Failed to update versions db.

Caused by:
    0: `run_command_update_version_db` command failed to load configuration db.
    1: Failed to open juliaup config file.
    2: No such file or directory (os error 2)
```
running `juliaup self uninstall` will fail to correctly remove the `paths.juliaupselfhome` folder since it will point to two parent folders with respect to the symlink location instead of the true executable

ghyatzo:
Fixes #885,

It centralises the resolution of the self path into a single function. Consolidated all the locations which queried the self path to call that function for better maintainability and consistency.

this change could also help towards #844, giving better support to symlinks and therefore custom locations

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -4,6 +4,7 @@ use itertools::Itertools;
 use juliaup::config_file::{load_config_db, JuliaupConfig, JuliaupConfigChannel};
 use juliaup::global_paths::get_paths;
 use juliaup::jsonstructs_versionsdb::JuliaupVersionDB;
+use juliaup::utils::get_juliaup_path;
 use juliaup::versions_file::load_versions_db;
 #[cfg(not(windows))]
 use nix::{
@@ -20,20 +21,6 @@ use std::path::PathBuf;
 #[error("{msg}")]
 pub struct UserError {
     msg: String,
-}
-
-fn get_juliaup_path() -> Result<PathBuf> {
-    let my_own_path = std::env::current_exe()
-        .with_context(|| "std::env::current_exe() did not find its own path.")?
-        .canonicalize()
-        .with_context(|| "Failed to canonicalize the path to the Julia launcher.")?;
-
-    let juliaup_path = my_own_path
-        .parent()
-        .unwrap() // unwrap OK here because this can't happen
-        .join(format!("juliaup{}", std::env::consts::EXE_SUFFIX));
-
-    Ok(juliaup_path)
 }
 
 fn do_initial_setup(juliaupconfig_path: &Path) -> Result<()> {
--- a/src/command_selfupdate.rs
+++ b/src/command_selfupdate.rs
@@ -71,19 +71,12 @@ pub fn run_command_selfupdate(paths: &GlobalPaths) -> Result<()> {
                 )
             })?;
 
-        let my_own_path = std::env::current_exe()
-            .with_context(|| "Could not determine the path of the running exe.")?;
-
-        let my_own_folder = my_own_path
-            .parent()
-            .ok_or_else(|| anyhow!("Could not determine parent."))?;
-
         eprintln!(
             "Found new version {} on channel {}.",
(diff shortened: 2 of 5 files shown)


davidanthoff (changes requested):
I like it in general, but maybe the right strategy here would be to just store the path to the bin and the parent folder in the `GlobalPaths` structure and then use that everywhere? That seems more consistent with the current design.

ghyatzo:
An Issue I encounter with that approach is that the `juliaupselfbin` which would be the `parent` folder to `get_juliaup_path()`, is locked behind the `selfupdate` feature, and so, not available in all places where access to the self path is required (i.e.: in the `install_version` function in the `operations.rs` file).

Another issue is that that would require passing the `GlobalPaths` object to many functions that as of now do not require it and would need to get their signature updated:
- `install_background_selfupdate(interval: i64)`
- `get_bin_dir()`
- `do_initial_setup(juliaupconfig_path: &Path)`
- `run_versiondb_update(
    config_file: &juliaup::config_file::JuliaupReadonlyConfigFile,
)`
- `run_selfupdate(config_file: &juliaup::config_file::JuliaupReadonlyConfigFile)`

are all functions that would need to change signature for them to use the `GlobalPaths` object.

Now that I dig a bit more in depth, I noticed also some inconsistencies on how the program refers to its own bin folder:
there is in use the function`get_bin_dir()` (from #165) for some symlink operations, which searches an env variable `JULIAUP_BIN_DIR` or defaults to the parent of the binary if and only if the bin dir is within the user home folder.
But in the `GlobalPaths` object we refer to the bin folder only as the parent of the executable (it was also like that before my changes).

It would very much be convenient to standardise this, so that there is only one "authority" that tells us which is the bin folder.
From my point of view there are two possibilities:
- let the binary (`current_exe()`) decide where is the bin folder, but that would conflict with existing `JULIAUP_BIN_DIR` and therefore would require either rethinking that pull request. Moreover, now that `JULIAUP_DEPOT_PATH` is a thing, we would have that `juliaupselfhome != juliauphome` which I have trouble imagining a scenario where that could be beneficial.
- let the bin dir (either `JULIAUP_BIN_DIR` or a default location) decide where to find the executable. This in my opinion would instead create some possible conflicts with the `JULIAUP_DEPOT_PATH` which would need to remain in sync ot it might create inconsistencies.
- let the `JULIAUP_DEPOT_PATH` (or a default path if it isn't set) be the only authority where we can find the bin dir and therefore the binary. This means getting rid of both `JULIAUP_BIN_DIR` and `current_exe()` calls.

Personally I think that the most robust option is using `JULIAUP_DEPOT_PATH` as the only authority if juliaup get's installed in a non default location. It makes little sense to have both `JULIAUP_DEPOT_PATH` ***and*** `JULIAUP_BIN_DIR`, and at that point we could also get rid of all `current_exe()` calls.
then
- `juliaupselfhome = $JULIAUP_DEPOT_PATH`
- `juliaupselfbin = juliaupselfhome\bin`
- `my_own_path = juliaupselfbin\juliaup`

But it could be breaking for some (probably only those using `JULIAUP_BIN_DIR`, which isn't even documented).

Unless I am missing something foundamental, I don't really see why we should ever have `juliauphome != juliaupselfhome` (which would require keeping the reference to the `current_exe`)

ghyatzo:
@davidanthoff bump? any news regarding my comment above and/or your version in #915 ?
Shall I close this?

ghyatzo:
Refer to #915

PR #889, opened 2024-04-06
Prevent downgrading on old version in mirror

### LINKED ISSUE #250: Warn if server version is less than installed version.
I am in China. Usually, I need to use mirror to update juliaup. Some mirrors fail to update for a long time resulting my juliaup downgrading when I want to upgrade.

I think it is a good idea to warn users if the latest version from server is less than installed version so that users know the present mirror server is not updating.

The default behavior should be not doing anything after warning since the command is `juliaup self update` rather than something like `juliaup self sync version`.

ghyatzo:
should fix #250


davidanthoff (changes requested):
I think this would also disable the ability to switch from a prerelease channel to the release channel, where we _want_ a downgrade to happen. So I don't think we can merge this as is...

ghyatzo:
I completely missed that possibility, sorry.
I tried to reason about it in terms of detecting a channel change, but as far as I could tell there is no way currently to do it.
Instead I opted for simplifying it, and just warn and ask the user for input when we detect a downgrade.

I have ready an incomplete patch for it, but there is another issue. What should be the default behaviour when the self update is run non interactlively. Should we default to not downgrade at all? Or install whatever it is found on the channel?

I'd be more towards ignoring any downgrades, to avoid surprises.

ghyatzo:
I just opted for pushing a complete working version, so that if you agree it's ready.

DilumAluthge:
This PR has not been updated in over one year.

@ghyatzo Are you still interested in working on this PR?

ghyatzo:
The PR was ready for review at the time. I can rebase this on master and see where it goes from there

DilumAluthge:
Yeah, that would be great. Once you've rebased, let me know, and I can try to find a reviewer.

DilumAluthge:
@IanButterworth or @davidanthoff - Could you re-review this, and see if the previously requested changes have been made?

PR #915, opened 2024-04-30
Centralize where paths are discovered

### LINKED ISSUE #885: `juliaup` doesn't correctly resolve symlinks when self updating on macOS

davidanthoff:
Alternative to #886.

@ghyatzo I started to look into it, and then got carried away ;) I _think_ this should be correct, but I'm also a bit nervous as any error in any of this would really be bad... If you could take a careful look if this all looks right to you that would be amazing.

The general idea now is that `juliaupselfexecfolder` is the folder where the `juliaup` binary is located and `juliaupselfexec` is the path to the `juliaup` binary itself. And all paths should be canonical.

There is one thing I don't fully understand, I'll leave a comment at that location.

### DIFF
--- a/src/global_paths.rs
+++ b/src/global_paths.rs
@@ -12,8 +12,8 @@ pub struct GlobalPaths {
     pub juliaupselfhome: PathBuf,
     #[cfg(feature = "selfupdate")]
     pub juliaupselfconfig: PathBuf,
-    #[cfg(feature = "selfupdate")]
-    pub juliaupselfbin: PathBuf,
+    pub juliaupselfexecfolder: PathBuf,
+    pub juliaupselfexec: PathBuf,
 }
 
 fn get_juliaup_home_path() -> Result<PathBuf> {
@@ -54,17 +54,22 @@ fn get_default_juliaup_home_path() -> Result<PathBuf> {
 }
 
 pub fn get_paths() -> Result<GlobalPaths> {
+    use anyhow::Context;
+
     let juliauphome = get_juliaup_home_path()?;
 
-    #[cfg(feature = "selfupdate")]
     let my_own_path = std::env::current_exe()
-        .with_context(|| "Could not determine the path of the running exe.")?;
+        .with_context(|| "std::env::current_exe() did not find its own path.")?
+        .canonicalize()
+        .with_context(|| "Failed to canonicalize the path to the Julia launcher.")?;
 
-    #[cfg(feature = "selfupdate")]
-    let juliaupselfbin = my_own_path
+    let juliaupselfexecfolder = my_own_path
         .parent()
         .ok_or_else(|| anyhow!("Could not determine parent."))?
         .to_path_buf();
+
+    let juliaupselfexec = juliaupselfexecfolder
+        .join(format!("juliaup{}", std::env::consts::EXE_SUFFIX));
 
     let juliaupconfig = juliauphome.join("juliaup.json");
 
@@ -92,7 +97,7 @@ pub fn get_paths() -> Result<GlobalPaths> {
         juliaupselfhome,
         #[cfg(feature = "selfupdate")]
         juliaupselfconfig,
-        #[cfg(feature = "selfupdate")]
-        juliaupselfbin,
+        juliaupselfexecfolder,
+        juliaupselfexec,
     })
 }
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -292,10 +292,8 @@ pub fn install_version(
     // TODO At some point we could put this behind a conditional compile, we know
     // that we don't ship a bundled version for some platforms.
     let full_version_string_of_bundled_version = get_bundled_julia_version();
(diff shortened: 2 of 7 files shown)


src/global_paths.rs
```diff
@@ -54,18 +54,23 @@ fn get_default_juliaup_home_path() -> Result<PathBuf> {
...
 
-    #[cfg(feature = "selfupdate")]
-    let juliaupselfbin = my_own_path
+    let juliaupselfexecfolder = my_own_path
         .parent()
         .ok_or_else(|| anyhow!("Could not determine parent."))?
         .to_path_buf();
 
+    let juliaupselfexec = juliaupselfexecfolder
+        .join(format!("juliaup{}", std::env::consts::EXE_SUFFIX));
```
davidanthoff, line 72: So I don't really understand why we are doing this, i.e. why we start with the path to the binary, then take the parent, and then reconstruct the path to the binary here. BUT, that is how the code was before, and I'm nervous about changing it... And this might have been related to deal with some symbolic links stuff that I don't fully understand?

ghyatzo: Honestly I struggle to see the point of it as well. The only scenario I can think of for the reason of that, is that in windows `current_exe` doesn't return (maybe it didn't in the past?) the `.exe` suffix.
Dealing with symbolic links should be handled by `.canonicalize()` now anyway.

ghyatzo:
Nice to have this alternative!
I took a look, it all seems good, and actually clears up a lot of the code, which is nice.
I didn't want to touch the various features flags, or signatures, but happy that you did it, it feels better to have a single storage.

first a minor comment: probably still having the logic to get the exec path in a separate function can be beneficial for maintainability.

On another note, I am really uneasy about the `get_bin_dir()` function and the `JULIAUP_BIN_DIR` environmental variable.
I commented about it exensively in my PR.
That function `get_bin_dir` really feels out of place, and the `JULIAUP_BIN_DIR` is only used there as far as I can tell.

I would also really know what you think about the proposal of just instead using `JULIAUP_DEPOT_PATH` as the real "authority" and everything else can follow from that. (check the comment in #886)!

ghyatzo:
Any news regarding this? or comments in #886?

src/operations.rs
```diff
@@ -575,8 +573,8 @@ fn _remove_symlink(symlink_path: &Path) -> Result<()> {
     Ok(())
 }
 
-pub fn remove_symlink(symlink_name: &String) -> Result<()> {
-    let symlink_path = get_bin_dir()
+pub fn remove_symlink(symlink_name: &String, paths: &GlobalPaths) -> Result<()> {
+    let symlink_path = get_bin_dir(paths)
```
ghyatzo, line 577: See comment (second half): https://github.com/JuliaLang/juliaup/pull/886#issuecomment-2086510625
In short: I don't get why the `get_bin_dir` uses an undocumented environment variable `JULIAUP_BIN_DIR`.
If we want to depend on `current_exe()` then that makes `JULIAUP_BIN_DIR` a potential source of headaches, since almost by definition, the path obtained by `current_exe()` *is* the actual `JULIAUP_BIN_DIR` (unless I am not seeing the use case for it)

If the `JULIAUP_BIN_DIR` was exclusively introduced (in #165 ) as a kitchen sink for various versions' symlinks then I guess it's not such a big deal. But I must admit I still find it confusing, especially since it is not documented and so it seems like a left over from a past approach?

ghyatzo (approved):
I might have just now realised I was marked as reviewer on this PR :P
Added a comment reporting some doubts I had from my own PR about this topic.
But other than that I like this a lot.
Sorry for the long wait!

ghyatzo (approved):

ghyatzo:
What's the status of this PR? :)

StefanKarpinski:
Seems stalled out. @davidanthoff, would you like @ghyatzo to pick it up and continue it maybe?

fingolfin:
Ping @davidanthoff

ghyatzo:
New year new bump.

Shall I just make a new PR based on this and my previous one?

DilumAluthge:
@IanButterworth

IanButterworth:
I'll give it a week for response before trying to take over.

ghyatzo:
I'm a bit burdened with other things at the moment. Haven't forgot.
If things don't change in the mean time I'll take over as soon as I have some bandwidth

PR #872, opened 2024-03-19
FreeBSD support

ararslan:
This doesn't yet pass tests because the tarball fails to extract and I don't understand why. The URL it's using is correct and I'm able to download and unpack manually.

PR #948, opened 2024-06-05
Include FreeBSD when generating versiondb

ararslan:
This splits out the versiondb part of #872. This ensures that any updates to versiondb that happen while #872 is in development don't require re-syncing, and it will mean that tests will pass locally. (Though the test that failed locally that was related to this mysteriously passed on CI in #872.)

For some reason, versions.json lists the FreeBSD triplet with the suffix 11.1, which does not reflect the version of FreeBSD used for building and has not been supported for a long time, by us or upstream. In fact, the current FreeBSD binaries won't work on anything before 13.2. But in order to use versions.json without some special-casing and/or refactoring, we can just retain the weird suffix.

### DIFF
--- a/scripts/versiondb/updateversiondb.jl
+++ b/scripts/versiondb/updateversiondb.jl
@@ -29,6 +29,8 @@ function triplet2channel(triplet)
         "x86"
     elseif triplet=="aarch64-linux-gnu"
         "aarch64"
+    elseif triplet=="x86_64-unknown-freebsd11.1"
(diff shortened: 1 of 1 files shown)


ararslan:
I verified locally that the changes made here work, but I didn't include the output as part of this PR. I figured that we can rerun the versiondb update after this is merged so that everything is in a fully consistent state.

davidanthoff (approved):

Issue #942, opened 2024-06-01
Fails when /tmp is not executable


LilithHafner:
I recently failed to install Julia on a server with this error:
```julia
$ curl -fsSL https://install.julialang.org | sh

info: downloading installer
Cannot execute /tmp/tmp.aI66r38EKw/juliainstaller.
Please install the files to a location where you can execute binaries.
```
I don't see how to work around this issue.

davidanthoff:
The relevant line is https://github.com/JuliaLang/juliaup/blob/a351f4bb9c1b09c4b6070f91465b441aa63ca935/deploy/shellscript/juliaup-init.sh#L83.

I'm not entirely sure what the best workaround is... A command line option to manually specify a folder?

LilithHafner:
1. It is safe to assume that the place Julia will eventually be installed to is executable (if it's not then the install will be unusable) so that could be a good backup, as long as juliaup cleans up after itseld and that's not to automagical for your design philosophy
2. Manually specifying a tmp folder would be easier than downloading the executable and `scp`ing it in; that would be a nice workaround if 1 isn't viable

davidanthoff:
> the place Julia will eventually be installed to

The tricky thing is that we don't know that location when the shell script runs. The shell script downloads the actual installer, and the actual installer then prompts the user to pick a location (if they go for a custom location)...

So I think we probably just need the option to manually specify a folder for this.

davidanthoff:
Ha, so closing this as a duplicate :) Maybe we just need to document the `TMP` var trick, seems to me that is a good enough solution?

Issue #450, opened 2022-11-22
Cannot execute juliainstaller


aplavin:
First time trying juliaup, attempted to install in on a server (with gentoo). The result is:
```zsh
╰─$ curl -fsSL https://install.julialang.org | sh


info: downloading installer
Cannot execute /tmp/tmp.WgttSEbBvc/juliainstaller.
Please install the files to a location where you can execute binaries.
```
The server policy is to mount the `/tmp` filesystem as noexec, so makes sense that it couldn't execute that file.

I tried copying `/tmp/tmp.WgttSEbBvc/juliainstaller` to another location to run it, but the file doesn't seem to even exist.

What else can I do to install juliaup?

davidanthoff:
Maybe we need to add a command line argument for the shell script that allows one to specify a temporary location where the `juliainstaller` should be put? In a case like yours one could then simply specify an alternative location that is different than `/tmp`. Would that work?

fingolfin:
Issue #673 is related: there `curl` can't download to `/tmp`.

Issue #673, opened 2023-06-29
Install failing on Ubuntu (via standard curl command )


psombe:
I'm attempting to install `juliaup` via the standard command `curl -fsSL https://install.julialang.org | sh`. However, the script fails with the following output.
```
info: downloading installer
curl: (23) Failure writing output to destination
juliaup: command failed: downloader https://julialang-s3.julialang.org/juliaup/bin/juliainstaller-1.8.16-x86_64-unknown-linux-musl /tmp/tmp.8mk7iXLQ50/juliainstaller x86_64-unknown-linux-musl
```
It is unable to detect that my laptop has GNU Lib C installed. My laptop details are listed below.
```
(core-py) psombe@psombe-x1:~/Downloads$ ldd /bin/ls
	linux-vdso.so.1 (0x00007ffe2e7d1000)
	libselinux.so.1 => /lib/x86_64-linux-gnu/libselinux.so.1 (0x00007f1ad0b26000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f1ad0800000)
	libpcre2-8.so.0 => /lib/x86_64-linux-gnu/libpcre2-8.so.0 (0x00007f1ad0a8f000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f1ad0b8d000)
(core-py) psombe@psombe-x1:~/Downloads$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.2 LTS
Release:	22.04
Codename:	jammy

```

davidanthoff:
> It is unable to detect that my laptop has GNU Lib C installed.

That shouldn't matter, we now always install the `musl` version of Juliaup.

To me this just looks like some download/permission problem? Do you get the same error if you try it again?

psombe:
Yes. I get the same error.

davidanthoff:
Can you download `https://julialang-s3.julialang.org/juliaup/bin/juliainstaller-1.8.16-x86_64-unknown-linux-musl` to a location in `/tmp` with `curl`? That seems to not work here..

psombe:
I can download the file when I specify the file as `chk.txt` say. However, the script is mangling the output and failing to download. For example, I see a space in `/tmp/tmp.8mk7iXLQ50/juliainstaller x86_64-unknown-linux-musl` which maybe should not exist. I also don't see a `--create-dirs` option in the `curl` command.

[...]

PR #952, opened 2024-06-08
Add workaround as error hint to tmp-not-executable branch

LilithHafner:
I changed "Install the files to a location..." to "Use a tmp location..." because "the files" being installed are, according to the user, julia and maybe juliaup.

I used a hint instead of a fallback because the hint is less magical and almost as helpful. And because it's easier to implement. I'd be open to this being an automatic fallback when `tmp` is not executable, but then we'd need to handle cleanup logic as well.

If this PR merges I might make a followup to hint for https://github.com/JuliaLang/juliaup/issues/673, but tbh that feels like a system bug: curl should be able to write to the default tmpdir.

Fixes https://github.com/JuliaLang/juliaup/issues/450
Fixes https://github.com/JuliaLang/juliaup/issues/942

### DIFF
--- a/deploy/shellscript/juliaup-init.sh
+++ b/deploy/shellscript/juliaup-init.sh
@@ -136,7 +136,10 @@ main() {
     if [ ! -x "$_file" ]; then
         printf '%s\n' "Cannot execute $_file." 1>&2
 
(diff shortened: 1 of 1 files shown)


davidanthoff (approved):
Great!

PR #978, opened 2024-07-03
Make completions subcommand more user-friendly

Andriamanitra:
This small change makes the completions subcommand slightly more user-friendly by using strongly typed `Shell` for the completions subcommand argument.

## completions with invalid shell
```console
$ juliaup completions fishy  # before
Error: 'fishy' is not a supported shell.

$ ./target/release/juliaup completions fishy  # after
error: invalid value 'fishy' for '<SHELL>'
  [possible values: bash, elvish, fish, powershell, zsh]

  tip: a similar value exists: 'fish'

For more information, try '--help'.
```

## completions --help:
```console
$ juliaup completions --help  # before
Generate tab-completion scripts for your shell

Usage: juliaup completions <SHELL>

Arguments:
  <SHELL>  

Options:
  -h, --help  Print help

$ ./target/release/juliaup completions --help  # after
Generate tab-completion scripts for your shell

Usage: juliaup completions <SHELL>

Arguments:
  <SHELL>  [possible values: bash, elvish, fish, powershell, zsh]

Options:
  -h, --help  Print help
```

### DIFF
--- a/src/bin/juliaup.rs
+++ b/src/bin/juliaup.rs
@@ -110,6 +110,6 @@ fn main() -> Result<()> {
             #[cfg(not(feature = "selfupdate"))]
             SelfSubCmd::Uninstall {} => run_command_selfuninstall_unavailable(),
         },
-        Juliaup::Completions { shell } => run_command_completions(&shell),
+        Juliaup::Completions { shell } => run_command_completions(shell),
     }
 }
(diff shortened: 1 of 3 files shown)


LilithHafner (approved):
Thanks! This is also a better use of the typesystem. Shells should be represented as `Shell`s, not Strings, esp. if they are being converted to `Shell`s later.

PR #980, opened 2024-07-04
Support PR Channels

### LINKED ISSUE #903: Support pr channels
It would be lovely to have `julia +pr54152` download and run the binaries built from the pull request # 54152 (if binaries are available). If the build failed or the PR is old enough that the binaries have been discarded or my system is not compatible with any of the CI systems then I wouldn't expect this to be possible, but in most cases I care about none of those are the case. It would make interactive testing of a PR much nicer to e.g. verify that purported performance improvements are real.

I recall hearing on slack that with the new nightly support this shouldn't be too hard, but I can't find that conversation now.

LilithHafner:
Implements & closes #903 

- **support pr channels**
- **add docs**
- **add tests**

I tested a failed PR install, but not a successful one because no PR will be permanently available to test against.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1019,6 +1019,7 @@ dependencies = [
  "normpath",
  "path-absolutize",
  "predicates",
+ "regex",
  "reqwest",
  "semver",
  "serde",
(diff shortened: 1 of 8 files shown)


davidanthoff (approved):
Awesome!

PR #994, opened 2024-07-15
Make nightly update message quieter

### LINKED ISSUE #991: `nightly` channel can be updated message is too verbose
When launching `julia +nightly`, this message is almost always displayed:

```
A new version of Julia for the `nightly` channel is available. Run:

  juliaup update

to install the latest Julia for the `nightly` channel.
```

I think this message should be shorter because
- Folks using `nightly` are typically more experienced and need less detailed prompting
- nightly is updateable several times per day so this message will show more often than not unless folks update a couple of times a day.
- there is no way to opt-out
- It's typically fine to be several versions behind on the nightly channel [added in edit]

I propose

```
A new `nightly` version is available. Install with `juliaup update`.
```

LilithHafner:
Fixes #991

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -235,7 +235,7 @@ fn get_julia_path_from_channel(
             version: _,
         } => {
             if local_etag != server_etag {
-                if channel == "nightly" {
+                if channel.starts_with("nightly") {
                     // Nightly is updateable several times per day so this message will show
                     // more often than not unless folks update a couple of times a day.
                     // Also, folks using nightly are typically more experienced and need
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -235,14 +235,24 @@ fn get_julia_path_from_channel(
             version: _,
         } => {
             if local_etag != server_etag {
-                eprintln!(
-                    "A new version of Julia for the `{}` channel is available. Run:",
-                    channel
-                );
-                eprintln!();
-                eprintln!("  juliaup update");
-                eprintln!();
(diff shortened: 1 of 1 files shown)


christiangnrd:
From #991:
> I think this message should be shorter because
> 
> * Folks using `nightly` are typically more experienced and need less detailed prompting
> * nightly is updateable several times per day so this message will show more often than not unless folks update a couple of times a day.
> * there is no way to opt-out

I feel like those points also apply to PR channels, should this new shorter message also show for those?

src/bin/julialauncher.rs
```diff
@@ -235,14 +235,24 @@ fn get_julia_path_from_channel(
...
             if local_etag != server_etag {
-                eprintln!(
-                    "A new version of Julia for the `{}` channel is available. Run:",
-                    channel
-                );
-                eprintln!();
-                eprintln!("  juliaup update");
-                eprintln!();
-                eprintln!("to install the latest Julia for the `{}` channel.", channel);
+                if channel == "nightly" {
```
davidanthoff, line 238: Should also check for the arch versions, i.e. `nightly~x64` etc.

LilithHafner:
You're quite right that those three points apply to PRs.

However, I see a fourth point that does not apply to PR branches:

- It's typically fine to be several versions behind on the nightly channel

If I am reviewing a PR and testing an out of date version of that PR, I want a large warning/hint.

christiangnrd:
@LilithHafner I hadn't considered that at the time but I agree. PR channel requests should keep the large hint.

davidanthoff (approved):

PR #1002, opened 2024-07-25
Support overriding an override

### LINKED ISSUE #987: Allow overriding an override
I just discovered overrides, which work great, though I have one suggestion on setting overrides.

Currently if you set an override on a path that already has an override, you get an error and the override does not change:

```
❯ juliaup override set 1.10.4

❯ juliaup override set 1.10.4
Error: '\\?\D:\repo\ribasim\Ribasim' path already has an override configured.
```

This means that the only way to reliably set an override is this, because `unset` luckily doesn't error if it is already `unset`:

```
juliaup override unset
juliaup override set 1.10.4
```

I feel like if a user asks for a valid override, that should be done, rather than erroring out.

visr:
Fixes #987.

Now we get a stderr message on `override add`, which indicates if an override is changed, new, or already exists.

If the channel is not installed, it gives the same error as julialauncher.

Please check carefully since I'm new to Rust. Especially not sure if the `iter_mut` pattern is idiomatic.

### DIFF
--- a/src/command_override.rs
+++ b/src/command_override.rs
@@ -66,7 +66,11 @@ pub fn run_command_override_set(
         .with_context(|| "`override set` command failed to load configuration data.")?;
 
     if !config_file.data.installed_channels.contains_key(&channel) {
-        bail!("'{}' channel does not exist.", &channel);
+        bail!(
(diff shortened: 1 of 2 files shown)


visr:
Bump :)

PR #1009, opened 2024-08-10
Check for existing install location

### LINKED ISSUE #773: Symlink error when link/file exist
I already had the file `.juliaup/bin/julia` which resulted in the following error during installation:
```
Error: failed to create symlink `/Users/fredrik/.juliaup/bin/julia`.

Caused by:
    File exists (os error 17)
```

It seems like this case should be handled better. In my case the file that existed was already the correct symlink too.

davidanthoff:
Fixes #773.

### DIFF
--- a/src/bin/juliainstaller.rs
+++ b/src/bin/juliainstaller.rs
@@ -379,6 +379,34 @@ pub fn main() -> Result<()> {
         }
     }
 
+    if install_choices.install_location.exists() {
+        println!("You are trying to install Juliaup into the folder");
+        println!("`{}`,", install_choices.install_location);
+        println!("but that folder already exists.");
+
+        if args.disable_confirmation_prompt {
+            println!();
+            println!(
+                "Please remove the folder or use interactive mode."
+            );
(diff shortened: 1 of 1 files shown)


src/bin/juliainstaller.rs
```diff
@@ -379,6 +379,35 @@ pub fn main() -> Result<()> {
...
+            let continue_with_setup = Confirm::with_theme(theme.as_ref())
+                .with_prompt("Do you want to continue with the installation and overwrite the existing Juliaup installation folder?")
+                .default(true)
+                .interact_opt()?;
+
+            if !continue_with_setup.unwrap_or(false) {
+                return Ok(());
+            }
+
+            std::fs::remove_dir_all(install_choices.install_location)
```
davidanthoff, line 404: Or maybe this is too drastic? A user might have picked a folder where other (legit) stuff is, and then we delete it all. Probably not good... Maybe we just output an error message saying: "select a folder that doesn't exist" or something like that?

Issue #985, opened 2024-07-08
Installation attempt crash report


suzycheung:
I encountered the following error when attempting to install Julia with the command `curl -fsSL https://install.julialang.org | sh`

This is from the toml file:

```
"name" = "Juliainstaller"
"operating_system" = "Mac OS 14.5.0 [64-bit]"
"crate_version" = "1.14.9"
"explanation" = """
Panic occurred in file 'src/operations.rs' at line 1021
"""
"cause" = """
called `Result::unwrap()` on an `Err` value: Failed to open file /Users/suzy/.bash_profile.

Caused by:
    Permission denied (os error 13)"""
"method" = "Panic"
"backtrace" = """

   0: 0x104583d94 - core::result::unwrap_failed::h8e3b933261dd7fec
   1: 0x104447990 - juliaup::command_config_modifypath::run_command_config_modifypath::hca63d35b576be3c9
   2: 0x10435758c - juliainstaller::main::hb0bee2b2a4d9e158
   3: 0x104343268 - std::sys_common::backtrace::__rust_begin_short_backtrace::h2b1ac18042ef3658
   4: 0x10435bc4c - _main"""
```

mortenpi:
Duplicate of https://github.com/JuliaLang/juliaup/issues/920 I think.

Issue #981, opened 2024-07-05
Crash report


zhongpc:
`Juliainstaller had a problem and crashed. To help us diagnose the problem you can send us a crash report.
`

Here is the `toml` file

```
"name" = "Juliainstaller"
"operating_system" = "Mac OS 14.3.0 [64-bit]"
"crate_version" = "1.14.9"
"explanation" = """
Panic occurred in file 'src/operations.rs' at line 1021
"""
"cause" = """
called `Result::unwrap()` on an `Err` value: Failed to open file /Users/xxxx/.zshrc.

Caused by:
    Permission denied (os error 13)"""
"method" = "Panic"
"backtrace" = """

   0: 0x10104bd94 - core::result::unwrap_failed::h8e3b933261dd7fec
   1: 0x100f0f990 - juliaup::command_config_modifypath::run_command_config_modifypath::hca63d35b576be3c9
   2: 0x100e1f58c - juliainstaller::main::hb0bee2b2a4d9e158
   3: 0x100e0b268 - std::sys_common::backtrace::__rust_begin_short_backtrace::h2b1ac18042ef3658
   4: 0x100e23c4c - _main"""
```

Issue #938, opened 2024-05-29
Juliainstaller Crash Report


Rex-jay:
Hello everyone, I tried installing Julia on my Mac, here is the content of the error report;

`"name" = "Juliainstaller"
"operating_system" = "Mac OS 11.6.1 [64-bit]"
"crate_version" = "1.14.8"
"explanation" = """
Panic occurred in file 'src/operations.rs' at line 1021
"""
"cause" = """
called `Result::unwrap()` on an `Err` value: Failed to open file /Users/mac/.bash_profile.

Caused by:
    Permission denied (os error 13)"""
"method" = "Panic"
"backtrace" = """

   0: 0x1016e05f5 - core::result::unwrap_failed::h1d71c5bccf791789
   1: 0x101555b22 - juliaup::command_config_modifypath::run_command_config_modifypath::hbb65cb019f0e0f09
   2: 0x10142721b - juliainstaller::main::h581b6dbe2667c5d0
   3: 0x10140f586 - std::sys_common::backtrace::__rust_begin_short_backtrace::h78688f301d1253ab
   4: 0x10142d7c3 - _main"""
`

Issue #920, opened 2024-05-09
Fails when startup scripts are inaccessible (even if asked not to modify them)


goerz:
While preparing setup instructions for a tutorial, I was trying to install `juliaup` on a fresh macOS installation (inside an [UTM virtual machine](https://mac.getutm.app), using the default/automatic installation image provided by Apple for a minimal, unconfigured install). The VM is running Snonoma 14.4.1.

In the standard Terminal app, I tried to install `juliaup` with the recommended `curl -fsSL https://install.julialang.org | sh`, keeping all the default options. The installation failed ("Juliainstaller had a problem and crashed"). The report file boils down to

```
Failed to open file /Users/goerz/.bash_profile

Caused by:
    Permission denied (os error 13)
```

This appears to be related to some kind of macOS security feature, as I get `permission denied` if I try to manually run `touch .bash_profile` in the terminal. Similarly, for other configuration files like `~/.bashrc` and `~/.zshrc`, although `touch random_file` works. Strangely, I haven't been able to find anything on Google about this issue. The only workaround I've found is to run

```
sudo touch .bashrc .bash_profile .zshrc
sudo chown goerz:staff .bashrc .bashr_profile .zshrc
```

to create the files. Once they exist, the `juliaup` installation runs through without any issues.

In principle, I would expect the `juliaup` installer to work around any such macOS permission issues (potentially even by prompting for a sudo password).

Strangely, the crash during the `juliaup` installation also occurs if I customize the installation to decline "Do you want to add the Julia binaries to your PATH by manipulating various shell startup scripts?".

Even if `juliaup` cannot work around the permissions issue, that seems like bug! I would expect not to run into any kind of issue at all related to the startup files if I don't ask the installer to modify the startup file.

LilithHafner:
Applications in mac (even Terminal) often need explicit user permission to modify "files and folders". It's hard to install software without modifying files.

See https://support.apple.com/guide/mac-help/control-access-to-files-and-folders-on-mac-mchld5a35146/mac for a mediocre explanation.

goerz:
[...]

PR #1014, opened 2024-08-19
Check file permissions before installing

davidanthoff:
Fixes https://github.com/JuliaLang/juliaup/issues/985, https://github.com/JuliaLang/juliaup/issues/981, https://github.com/JuliaLang/juliaup/issues/938, https://github.com/JuliaLang/juliaup/issues/920, https://github.com/JuliaLang/juliaup/issues/918, https://github.com/JuliaLang/juliaup/issues/891, https://github.com/JuliaLang/juliaup/issues/874, https://github.com/JuliaLang/juliaup/issues/864, https://github.com/JuliaLang/juliaup/issues/826, https://github.com/JuliaLang/juliaup/issues/796,
https://github.com/JuliaLang/juliaup/issues/624.

### DIFF
--- a/src/bin/juliainstaller.rs
+++ b/src/bin/juliainstaller.rs
@@ -388,6 +388,41 @@ pub fn main() -> Result<()> {
         return Ok(());
     }
 
+    if install_choices.modifypath {
+        // Now check whether we have the necessary permissions for all the files
+        // we want to modify.
+
+        let paths = find_shell_scripts_to_be_modified(true)?;
+
+        let failed_paths: Vec<PathBuf> = [];
+
+        for cur_path in paths {
+            let mut file_result = std::fs::OpenOptions::new()
+                .read(true)
+                .write(true)
+                .open(cur_path);
+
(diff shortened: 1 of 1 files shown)


src/bin/juliainstaller.rs
```diff
@@ -388,6 +388,43 @@ pub fn main() -> Result<()> {
...
+        for cur_path in paths {
+            let file_result = std::fs::OpenOptions::new()
+                .read(true)
+                .write(true)
+                .open(&cur_path);
+
+            if file_result.is_err() {
+                failed_paths.push(cur_path.clone());
+            }
+        }
```
kdheepak, line 408: This will fail if the file doesn't exist at that location.

davidanthoff: If you look at how we construct the `paths` vector, we look for existing files and then only run this for loop for files that we found on disc. Admittedly, there is a race condition here that someone could delete the file between our check and this code here, but I think that is ok, it will just error and then they can try again.

PR #1028, opened 2024-09-02
Ensure Julia terminates when launcher terminates on Windows

davidanthoff:
This essentially follows the logic in libuv https://github.com/libuv/libuv/blob/5ff1fc724f7f53d921599dbe18e6f96b298233f1/src/win/process.c#L69.

The idea is that when the launcher process is killed, we want to make sure the actual Julia process also is terminated.

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -5,18 +5,18 @@ use juliaup::config_file::{load_config_db, JuliaupConfig, JuliaupConfigChannel};
 use juliaup::global_paths::get_paths;
 use juliaup::jsonstructs_versionsdb::JuliaupVersionDB;
 use juliaup::versions_file::load_versions_db;
-use normpath::PathExt;
-#[cfg(not(windows))]
-use std::os::unix::process::CommandExt;
-use std::path::Path;
-use std::path::PathBuf;
 #[cfg(not(windows))]
 use nix::{
     sys::wait::{waitpid, WaitStatus},
     unistd::{fork, ForkResult},
 };
+use normpath::PathExt;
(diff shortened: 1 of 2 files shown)


davidanthoff:
@vtjnash any chance you could take a look at this? Does this look reasonable to you?

src/bin/julialauncher.rs
```diff
@@ -430,11 +437,51 @@ fn run_app() -> Result<i32> {
...
+        let mut job_attr: windows::Win32::Security::SECURITY_ATTRIBUTES =
+            windows::Win32::Security::SECURITY_ATTRIBUTES::default();
+        let mut job_info: windows::Win32::System::JobObjects::JOBOBJECT_EXTENDED_LIMIT_INFORMATION =
+            windows::Win32::System::JobObjects::JOBOBJECT_EXTENDED_LIMIT_INFORMATION::default();
+
+        job_attr.bInheritHandle = false.into();
+        job_info.BasicLimitInformation.LimitFlags =
+            windows::Win32::System::JobObjects::JOB_OBJECT_LIMIT_BREAKAWAY_OK
+                | windows::Win32::System::JobObjects::JOB_OBJECT_LIMIT_SILENT_BREAKAWAY_OK
+                | windows::Win32::System::JobObjects::JOB_OBJECT_LIMIT_DIE_ON_UNHANDLED_EXCEPTION
```
vtjnash, line 449: This is a bug in libuv v1.x for compatibility reasons, but is fixed in libuv master
```suggestion
```

vtjnash:
okay, seems pretty reasonable

PR #1030, opened 2024-09-03
Prune linked versions that no longer exist with `--prune-linked` option.

### LINKED ISSUE #965: `juliaup gc` should prune linked versions that no longer exist
If I have a `juliaup` configuration with lots of linked julia versions it can be unwieldy to clean them up.  Uf those versions of Julia have been deleted from my harddrive, I think it makes sense that `juliaup gc` should remove the links too (or at least provide an option to do so, such as `--prune-linked`).

christiangnrd:
Depends on #1029 
Fixes #965


davidanthoff (approved):
Nice!

Maybe you could add a test? But I'll merge this now and you could add a test in a follow up PR.

christiangnrd:
I've created an issue so I don't forget.

PR #1031, opened 2024-09-03
Notify user when trying to launch nonexistent linked channel

### LINKED ISSUE #979: Nonexistent linked channels silently fail
When adding a new channel that links to nothing, no error is thrown, and when launching that nonexistent channel, again, no error is thrown.
```sh
$ juliaup link mychannel does_not_exist
$ echo $?
0
$ julia +mychannel
$ echo $?
0
$
```

christiangnrd:
Only warn when adding nonexistent julia install, as I find it useful to be able to add the link before I have the julia binary.

Fixes #979

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -5,6 +5,7 @@ use itertools::Itertools;
 use juliaup::config_file::{load_config_db, JuliaupConfig, JuliaupConfigChannel};
 use juliaup::global_paths::get_paths;
 use juliaup::jsonstructs_versionsdb::JuliaupVersionDB;
+use juliaup::utils::is_valid_julia_path;
 use juliaup::versions_file::load_versions_db;
 #[cfg(not(windows))]
 use nix::{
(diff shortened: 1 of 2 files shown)


src/bin/julialauncher.rs
```diff
@@ -393,9 +395,16 @@ fn run_app() -> Result<i32> {
             }
 
             // replace the current process
-            std::process::Command::new(julia_path)
-                .args(&new_args)
-                .exec();
+            if is_valid_julia_path(&julia_path) {
```
davidanthoff, line 398: Hm, this will essentially add a complete Julia launch into the path before we _actually_ start Julia, right? I think that would add too much overhead.

I would assume that we can instead just handle the error if the `exec` command doesn't succeed?

davidanthoff (changes requested):
I think the check on launch we can do in some other way, but the check on link seems good as is.

src/bin/julialauncher.rs
christiangnrd: Your suggestion seems to work!

davidanthoff (approved):

PR #1037, opened 2024-09-06
Add `up` alias to self update command

christiangnrd:
I can't remember where I saw this mentioned but very easy.


davidanthoff:
> Unless they were defined that way to have different docstrings, in which case I can undo those changes.

Yes, that was the idea :)

christiangnrd:
Fixed

davidanthoff (approved):

PR #1041, opened 2024-09-06
feat: Add logging with tracing

kdheepak:
This PR adds the setup for logging using the `tracing` module. 

After `init_logging` is called, a log file will be created in one of these locations (per XDG standard):

image

or in a folder in the environment variable `JULIAUP_DATA_HOME` or in the current directory. 

`tracing::info!` or `tracing::debug!` can be used to log into the log file.

Additionally, `RUST_LOG` environment variable can be used to control the logs for the various modules or dependencies.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -173,7 +173,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40723b8fb387abc38f4f4a37c09073622e41dd12327033091ef8950659e6dc0c"
 dependencies = [
  "memchr",
- "regex-automata",
+ "regex-automata 0.4.7",
  "serde",
 ]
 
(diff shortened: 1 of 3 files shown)


davidanthoff:
If we were to use those locations, then on Windows it should go into the `AppData\Local` not `AppData\Roaming` folder.

BUT, given that Julia generally has not adopted XDG at this point and has a standard log file location of `.julia/logs` we should just also put them there (ideally by checking the value of `JULIAUP_DEPOT_PATH`).

The other question I have is how does this handle concurrency? Generally, we need a solution where multiple processes all write logs at the same time. This here doesn't seem to handle that scenario?

And finally, this only adds the infrastructure, but never actually calls the init stuff, right? I think in general we probably want no logging by default but only opt into it via some configuration or something like that?

tecosaur:
I can't say I really see a downside of on-by-default logging, it could be nice to have a record of what versions are installed/uninstalled.

How I wish Julia itself actually paid attention to platform standards instead of fighting them...

davidanthoff:
Well, the _really_ Windows native solution would be to use the platform logic infrastructure instead of log files. I'm pretty sure `tracing` has support for that. But at the end of the day I think it is easier if we just stick with the Julia convention here.

PR #1059, opened 2024-10-05
Launch the manifest-specified Julia version

### LINKED ISSUE #10: Start specific Julia version if it is specified in Project.toml/Manifest.toml

tecosaur:
Closes #10.

Auto-installing would be the next step, but let's do that another day :slightly_smiling_face:.

Code review would be great, I haven't written rust in years :wink:.

Things needed before we can merge:
- [ ] Unit tests for command line parsing functions
- [ ] Feature tests for the entire new functionality

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -180,10 +180,10 @@ fn get_julia_path_from_channel(
 ) -> Result<(PathBuf, Vec<String>)> {
     if let JuliaupChannelSource::Manifest { version } = launch_parameters {
         let version_string = versions_db.available_channels.get(version)
-            .ok_or_else(|| anyhow!("The project you are trying to launch uses Julia {}, but no such Julia version exists. Please make sure you are using a valid Julia manifest file.", version))?;
+            .ok_or_else(|| UserError {msg: format!("The project you are trying to launch uses Julia {}, but no such Julia version exists. Please make sure you are using a valid Julia manifest file.", version) } )?;
 
         let version_config = config_data.installed_versions.get(&version_string.version)
-            .ok_or_else(|| anyhow!("The project you are trying to launch uses Julia {}, but you do not have that version installed. You can install it by running `juliaup add {}`.", version, version))?;
+            .ok_or_else(|| UserError {msg: format!("The project you are trying to launch uses Julia {}, but you do not have that version installed. You can install it by running `juliaup add {}`.", version, version) } )?;
 
         let absolute_path = juliaupconfig_path
             .parent()
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -13,12 +13,14 @@ use nix::{
     unistd::{fork, ForkResult},
 };
 use normpath::PathExt;
+use semver::Version;
 #[cfg(not(windows))]
 use std::os::unix::process::CommandExt;
 #[cfg(windows)]
 use std::os::windows::io::{AsRawHandle, RawHandle};
 use std::path::Path;
 use std::path::PathBuf;
+use toml::Value;
 #[cfg(windows)]
 use windows::Win32::System::{
     JobObjects::{AssignProcessToJobObject, SetInformationJobObject},
@@ -161,10 +163,12 @@ fn check_channel_uptodate(
     Ok(())
 }
 
+#[derive(PartialEq, Eq)]
 enum JuliaupChannelSource {
     CmdLine,
     EnvVar,
     Override,
+    Manifest,
     Default,
 }
 
@@ -176,6 +180,12 @@ fn get_julia_path_from_channel(
     juliaup_channel_source: JuliaupChannelSource,
 ) -> Result<(PathBuf, Vec<String>)> {
     let channel_valid = is_valid_channel(versions_db, &channel.to_string())?;
+    if juliaup_channel_source == JuliaupChannelSource::Manifest {
+        let path =
+            get_julia_path_for_version(config_data, juliaupconfig_path, &Version::parse(channel)?)?;
+        return Ok((path, Vec::new()));
+    }
+
     let channel_info = config_data
             .installed_channels
             .get(channel)
@@ -208,6 +218,8 @@ fn get_julia_path_from_channel(
                     }
                 }.into(),
                 JuliaupChannelSource::Default => UserError {msg: format!("The Juliaup configuration is in an inconsistent state, the currently configured default channel `{}` is not installed.", channel) }
+                JuliaupChannelSource::Manifest => unreachable!(),
+                JuliaupChannelSource::Default => anyhow!("The Juliaup configuration is in an inconsistent state, the currently configured default channel `{}` is not installed.", channel)
             })?;
 
     match channel_info {
@@ -308,6 +320,189 @@ fn get_override_channel(
     }
 }
 
+fn get_program_file(args: &Vec<String>) -> Option<(usize, &String)> {
+    let mut program_file: Option<(usize, &String)> = None;
+    let no_arg_short_switches = ['v', 'h', 'i', 'q'];
+    let no_arg_long_switches = [
+        "--version",
+        "--help",
+        "--help-hidden",
+        "--interactive",
+        "--quiet",
+        // Hidden options
+        "--lisp",
+        "--image-codegen",
+        "--rr-detach",
+        "--strip-metadata",
+        "--strip-ir",
+        "--permalloc-pkgimg",
+        "--heap-size-hint",
+        "--trim",
+    ];
+    let mut skip_next = false;
+    for (i, arg) in args.iter().skip(1).enumerate() {
+        if skip_next {
+            skip_next = false;
+        } else if arg == "--" {
+            if i + 1 < args.len() {
+                program_file = Some((i + 1, args.get(i + 1).unwrap()));
+            }
+            break;
+        } else if arg.starts_with("--") {
+            if !no_arg_long_switches.contains(&arg.as_str()) && !arg.contains('=') {
+                skip_next = true;
+            }
+        } else if arg.starts_with("-") {
+            let arg: Vec<char> = arg.chars().skip(1).collect();
+            if arg.iter().all(|&c| no_arg_short_switches.contains(&c)) {
+                continue;
+            }
+            for (j, &c) in arg.iter().enumerate() {
+                if no_arg_short_switches.contains(&c) {
+                    continue;
+                } else if j < arg.len() - 1 {
+                    break;
+                } else {
+                    // `j == arg.len() - 1`
+                    skip_next = true;
+                }
+            }
+        } else {
+            program_file = Some((i, arg));
+            break;
+        }
+    }
+    return program_file;
+}
+
+fn get_project(args: &Vec<String>) -> Option<PathBuf> {
+    let program_file = get_program_file(args);
+    let recognised_proj_flags: [&str; 4] = ["--project", "--projec", "--proje", "--proj"];
+    let mut project_arg: Option<String> = None;
+    for arg in args
+        .iter()
+        .take(program_file.map_or(args.len(), |(i, _)| i))
+    {
+        if arg.starts_with("--proj") {
+            let mut parts = arg.splitn(2, '=');
+            if recognised_proj_flags.contains(&parts.next().unwrap_or("")) {
+                project_arg = Some(parts.next().unwrap_or("@.").to_string());
+            }
+        }
+    }
+    let project = if project_arg.is_some() {
+        project_arg.unwrap()
+    } else if let Ok(val) = std::env::var("JULIA_PROJECT") {
+        val
+    } else {
+        return None;
+    };
+    if project == "@" {
+        return None;
+    } else if project == "@." || project == "" {
+        let mut path = PathBuf::from(std::env::current_dir().unwrap());
+        while !path.join("Project.toml").exists() && !path.join("JuliaProject.toml").exists() {
+            if !path.pop() {
+                return None;
+            }
+        }
+        return Some(path);
+    } else if project == "@script" {
+        if let Some((_, file)) = program_file {
+            let mut path = PathBuf::from(file);
+            path.pop();
+            while !path.join("Project.toml").exists() && !path.join("JuliaProject.toml").exists() {
+                if !path.pop() {
+                    return None;
+                }
+            }
+            return Some(path);
+        } else {
+            return None;
+        }
+    } else if project.starts_with('@') {
+        let depot = match std::env::var("JULIA_DEPOT_PATH") {
+            Ok(val) => match val.split(':').next() {
+                Some(p) => PathBuf::from(p),
+                None => dirs::home_dir().unwrap().join(".julia"),
+            },
+            _ => dirs::home_dir().unwrap().join(".julia"),
+        };
+        let path = depot.join("environments").join(&project[1..]);
+        if path.exists() {
+            return Some(path);
+        } else {
+            return None;
+        }
+    } else {
+        return Some(PathBuf::from(project));
+    }
+}
+
+fn julia_version_from_manifest(path: PathBuf) -> Option<Version> {
+    let manifest = if path.join("JuliaManifest.toml").exists() {
+        path.join("JuliaManifest.toml")
+    } else if path.join("Manifest.toml").exists() {
+        path.join("Manifest.toml")
+    } else {
+        return None;
+    };
+    let content = std::fs::read_to_string(manifest)
+        .ok()?
+        .parse::<Value>()
+        .ok()?;
+    if let Some(manifest_format) = content.get("manifest_format") {
+        if manifest_format.as_str()?.starts_with("2.") {
+            if let Some(julia_version) = content.get("julia_version") {
+                return julia_version.as_str().and_then(|v| Version::parse(v).ok());
+            }
+        }
+    }
+    return None;
+}
+
+fn get_julia_path_for_version(
+    config_data: &JuliaupConfig,
+    juliaupconfig_path: &Path,
+    version: &Version,
+) -> Result<PathBuf> {
+    let mut best_match: Option<(&String, Version)> = None;
+    for (installed_version_str, path) in &config_data.installed_versions {
+        if let Ok(installed_semver) = Version::parse(installed_version_str) {
+            if installed_semver.major != version.major || installed_semver.minor != version.minor {
+                continue;
+            }
+            if let Some((_, ref best_version)) = best_match {
+                if installed_semver > *best_version {
+                    best_match = Some((&path.path, installed_semver));
+                }
+            } else {
+                best_match = Some((&path.path, installed_semver));
+            }
+        }
+    }
+    if let Some((path, _)) = best_match {
+        let absolute_path = juliaupconfig_path
+            .parent()
+            .unwrap()
+            .join(path)
+            .join("bin")
+            .join(format!("julia{}", std::env::consts::EXE_SUFFIX))
+            .normalize()
+            .with_context(|| {
+                format!(
+                    "Failed to normalize path for Julia binary, starting from `{}`.",
+                    juliaupconfig_path.display()
+                )
+            })?;
+        return Ok(absolute_path.into_path_buf());
+    } else {
+        return Err(anyhow!(
+            "No installed version of Julia matches the requested version."
+        ));
+    }
+}
+
 fn run_app() -> Result<i32> {
     if std::io::stdout().is_terminal() {
         // Set console title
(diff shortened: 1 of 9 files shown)


src/bin/julialauncher.rs
```diff
@@ -300,6 +302,181 @@ fn get_override_channel(
...
+    for (_, arg) in args.iter().enumerate() {
+        if arg.starts_with("--project=") {
+            project_arg = Some(arg["--project=".len()..].to_string());
+            // Note: You'd think this might work, but it's not actually supported.
+            // } else if arg == "--project" && i + 1 < args.len() {
+            //     project_arg = Some(args[i + 1].clone());
+        } else if arg == "--" {
+            break;
+        }
+    }
```
davidanthoff, line 316: This loop still needs to be modified so that it doesn't pick up things like `julia main.jl --project=@.`.

src/bin/julialauncher.rs
```diff
@@ -300,6 +302,181 @@ fn get_override_channel(
...
+    } else if path.join("JuliaManifest.toml").exists() {
+        path.join("JuliaManifest.toml")
+    } else {
+        return None;
+    };
+    let file = std::fs::File::open(manifest).ok()?;
+    let reader = std::io::BufReader::new(file);
+    // This is a somewhat bootleg way to parse the manifest,
+    // but since we know the format it should be fine.
+    for line in reader.lines() {
```
davidanthoff, line 429: We should use a proper TOML parser here, I'm sure there is one for Rust.

src/bin/julialauncher.rs
```diff
@@ -300,6 +302,181 @@ fn get_override_channel(
...
+            }
+            if let Some((_, ref best_version)) = best_match {
+                if installed_semver > *best_version {
+                    best_match = Some((&path.path, installed_semver));
+                }
+            } else {
+                best_match = Some((&path.path, installed_semver));
+            }
+        }
+    }
```
davidanthoff, line 457: I think we should broadly use a different strategy here. Two problems:
1. I don't think we should just auto-handle patch updates. If I get a repo that records it uses Julia 1.10.1, and I only have 1.10.2 installed, I would want it to _not_ automatically use that. There might have been a really good reason that the Julia version was recorded as 1.10.1. We should think about lots of options to alert me to what is going on and make it easy to opt-in, but the default IMO should be what was recorded in the manifest, just how we do it for packages.
2. This doesn't handle platforms well. We can always have multiple versions installed for the same version that are only differentiated by platform (say `1.10.1~x64` and `1.10.1~x86`).

Here is how I would handle this instead:
1. Read the Julia version from the manifest
2. Look in the version db for the channel that matches that version exactly, and extract the exact Julia version from that (this makes for a 1:1 mapping from Julia semver to _one_ thing that can show up in `installed_versions`
3. Check whether that version exists in `installed_versions`, and then either use that or error (or in the future offer to install that channel).

src/bin/julialauncher.rs
```diff
@@ -344,19 +531,23 @@ fn run_app() -> Result<i32> {
...
-        &config_file.data,
-        &julia_channel_to_use,
-        &paths.juliaupconfig,
-        juliaup_channel_source,
-    )
-    .with_context(|| {
-        format!(
-            "The Julia launcher failed to determine the command for the `{}` channel.",
-            julia_channel_to_use
+    let (julia_path, julia_args) = if let Some(path) = manifest_derived_julia_path {
```
davidanthoff, line 534: We probably should think the priority list here through a bit more... So with this PR here it is
1. Command line switch
2. Manifest
3. `JULIAUP_CHANNEL` env var
4. Folder override
5. Default

That might be the right choice, but we should just make sure that we think it is :)

davidanthoff (changes requested):
Very cool!

I'll open a PR against this branch here later today with a config option so that this can be behind a feature flag for a while.

Looking at all the command line parsing, I'm wondering whether we should factor that out a bit more and make it more general than mix it into the rest of the logic as it is right now. In particular because we will _also_ want to know soon whether `julia` is running in interactive setting or not, which will require more understanding of command line flags.

src/bin/julialauncher.rs
```diff
@@ -300,6 +302,181 @@ fn get_override_channel(
...
+}
+
+fn julia_version_from_manifest(path: PathBuf) -> Option<Version> {
+    let manifest = if path.join("Manifest.toml").exists() {
+        path.join("Manifest.toml")
+    } else if path.join("JuliaManifest.toml").exists() {
+        path.join("JuliaManifest.toml")
+    } else {
+        return None;
+    };
```
christiangnrd, line 424: `JuliaManifest.toml` should take priority over `Manifest.toml` as it does in Julia

src/bin/julialauncher.rs
christiangnrd: How should it handle a situation where an environment has multiple versioned manifests?

For example, in my Metal.jl dev environment, I have `Manifest.toml`, `Manifest-v1.11.toml`, and `Manifest-v1.12.toml`.

Maybe if there's a manifest that matches the next-highest priority setting, select that?

For example, if I start Julia with 
```bash
$ JULIAUP_CHANNEL=rc julia --project=.
```
in an environment where my default channel is `release`, and that I have `Manifest.toml`, `Manifest-v1.11.toml`, and `Manifest-v1.12.toml` in that folder, it would use `Manifest-v1.11.toml` since the next priority is `JULIAUP_CHANNEL`, and it's set to `rc`.

Calling 
```bash
$ julia --project=.
```
would launch v1.10 because that's the default channel.

I'm not exactly sure this should interact with the Julia Manifest searching priority order (listed in my other review comment), but it's worth figuring out.

src/bin/julialauncher.rs
```diff
@@ -300,6 +302,181 @@ fn get_override_channel(
...
+    let mut project_arg: Option<String> = None;
+    for (_, arg) in args.iter().enumerate() {
+        if arg.starts_with("--project=") {
+            project_arg = Some(arg["--project=".len()..].to_string());
+            // Note: You'd think this might work, but it's not actually supported.
+            // } else if arg == "--project" && i + 1 < args.len() {
+            //     project_arg = Some(args[i + 1].clone());
+        } else if arg == "--" {
+            break;
+        }
```
christiangnrd, line 315: This doesn't seem to handle my default way of launching Julia, which is `julia --proj`.

src/bin/julialauncher.rs
davidanthoff: Responding to my own post here, I think the order should be this:
1. Command line switch
2. `JULIAUP_CHANNEL` env var
3. Folder override
4. Manifest
5. Default

@christiangnrd I think it essentially doesn't, i.e. the Juliaup launcher will just always completely ignore versioned manifest files. If there is a `Manifest.toml`, then it will (by default) launch the Julia version in there, and then that Julia version will use that manifest. If one launches with say a command line flag `+1.11`, _and_ there is a 1.11 specific manifest, then Julia itself will of course use that manifest, but the code here in Juliaup doesn't know about this, nor does it need.

At some level this would all be much easier if one could just select a manifest file by name (and they could be freely named)... I think the ideal design would be `julia +manifestfilename` really... Say we had `julia +mymanifest.juliamanifest.toml` convention or something like that...But that would require changes in Julia base, and I'm not sure whether one could come up with a good structure there.

src/bin/julialauncher.rs
davidanthoff: Yep, but we don't need to look for the version specific manifest files.

davidanthoff:
> I'll open a PR against this branch here later today with a config option so that this can be behind a feature flag for a while.

I just added this to the PR here via a direct push.

src/bin/julialauncher.rs
tecosaur: Ah yes, thanks. Done.

src/bin/julialauncher.rs
tecosaur: We can, I just don't think it's needed. Since the manifest has a rather prescribed format, a manifest that doesn't work with this will cause issues with Julia anyway.

src/bin/julialauncher.rs
tecosaur: My handling of patch versions was informed by a conversation I was sitting in on in Triage a while back. I may be misremembering, but I recall discussion on how auto-selection of Julia version should interact with patch versions, and the conclusion was that the latest patch version should be used.

There are two aspects to this that I see:
- Whether you actually want bugs from an earlier Julia version
- Proliferation of installed Julia versions

In my mind, both of these point towards wanting to use the latest available patch version. The sole exception in my mind is when you *want* to reproduce bugs in the Base Julia, but I'm not sure that should be the guiding case for this design.

Having the same Julia version compiled for multiple platforms installed is an annoying edge-case...

src/bin/julialauncher.rs
tecosaur: Done.

src/bin/julialauncher.rs
tecosaur: Ta, I now handle `--proj`, `--proje`, `--projec`, and `--project`.

src/bin/julialauncher.rs
davidanthoff: In my mind the entire point of a manifest is to _exactly_ pin down dependencies, so that things just run if one instantiates it. I see no reason why Julia itself should be handled differently here from any other package, after all we are not also suggesting that the package loader can just load newer patch versions of packages, right? I really want to get to a point where if I receive a manifest, I can be pretty sure that the person that created this manifest ran as close as possible the same versions of everything that will run on my machine if I instantiate it.

This would also lead to really weird UX problems. Say I download a replication archive from a paper that records Julia 1.10.0 in the manifest. That replication code works around a bug in 1.10.0. The bug has since been fixed in 1.10.1, but the replication code actually depends on the presence of the bug, so it won't run correctly under 1.10.1. As a user of that code, what am I supposed to do now? Uninstall 1.10.1 entirely from my system to force Juliaup to actually run with the version that is recorded in the manifest? I will now also get a different version of Julia when running depending on (completely unrelated) channels I have installed, that also seems really weird.

The worry about binary bloat in terms of installed Julia versions of course is real. But I think we should try to fix that via UX. For example, the launcher could print a notice when it starts a project, and the compat section in the `Project.toml` suggests that a newer version of Julia could be used for that project. It could even offer an interactive option "hey, you are running with a project that is using an outdated Julia version" or something like that.

(As a side note, I just checked, my `.julia` folder is at 27GB. If we want to fight binary bloat, there seems some other villain in town ;)

tecosaur: Exact replication has occurred to me, and that is somewhat convincing.

However, in most projects I have manifests for, I want to use the newest patch version available. I don't think it should take as much effort to upgrade the patch version as the minor version, and I wouldn't see exact reproducibility at the expense of more involved patch version upgrading as a clear win.

Re: binary bloat, yes. I think we can manage that better via UX (I actually have a particular idea on this: I think juliaup should auto-install and GC them after some number of days of not being used, leaving manually added channels alone), but *also* installing less is good IMO (also from the UX perspective of waiting for more Julia versions to be installed less often).

tecosaur:
> Looking at all the command line parsing, I'm wondering whether we should factor that out a bit more and make it more general than mix it into the rest of the logic as it is right now. In particular because we will also want to know soon whether julia is running in interactive setting or not, which will require more understanding of command line flags.

If you take a look at the revised code, you should see the current `get_project` contains less general code.

Out of curiosity, why would we want to know whether it's being run interactively or not?

davidanthoff:
> Out of curiosity, why would we want to know whether it's being run interactively or not?

I'd generally like to replace the current prompts that say "Run `juliaup update` to update" with just a "Do you want to update (y/n)?" interactive prompt, but we must make sure to only show these interactive prompts when Julia itself is run in interactive mode.

For this feature here that would also be relevant: when run interactively, we could prompt for "This project requires Julia version X, do you want to install it (y/n)?", but in non-interactive mode we should just error (and ideally also offer a command line argument that triggers automatic installation).

src/bin/julialauncher.rs
davidanthoff: In my mind replication is such a central and important goal in scientific work that we should really support that well. Scientific code is, after all, an important target for Julia. And for that scenario, any of these fully auto-update mechanisms are really not good.

We should probably have configuration for some of this. We could have a config option `auto_update_manifests`. The default could be unconfigured. When the launcher is started with a manifest with an outdated Julia version, it could show a prompt `The Julia version in this project is outdated, do you want update it (u), run with the old version (n), or always update Julia versions in this case (a), never update Julia versions in this case (x)?`

In all of these cases we'll have to do a lot of checking, though, or probably will even have to run the resolver. At the end of the day, any of the packages listed in the manifest might actually exclude the newer Julia version via a compat section in its `Project.toml`, and we'll have to respect that. I think that is one of the main reasons that in my mind we should only ever run with a different Julia version if the package manager logic has signed off on that combination. But checking all these compats is not really something we can do on the hot-path of the launcher... 

>  I think juliaup should auto-install and GC them after some number of days of not being used, leaving manually added channels alone

Yes, agreed, I had thought about a similar idea. I think we should keep track of user projects in the config file, and then having that project tracked in the config file is the reason why that specific Julia version is not GCed. So essentially we would _not_ install a channel to instantiate a project, but instead just install that version as a dependency of the project.

src/bin/julialauncher.rs
davidanthoff: It just is very, very hacky, and I would much prefer to do the right thing here and use a tested library. https://docs.rs/toml/0.8.19/toml/ exists and seems easy enough.

tecosaur:
Ah yep, I can see the use of that. I'd be happy to PR that too later.

src/bin/julialauncher.rs
tecosaur: I don't know about all those "very"s :stuck_out_tongue:, with how nice and simple TOML is. There are exactly two ways this could produce an incorrect result:
1. A manifest with no initial `julia_version` key, and then a `julia_version` table key that's non-indented. However, (1) Pkg doesn't use arbitrary keys, and (2) Julia's TOML writer always indents nested table keys.
2. A multi-line string that contains a line starting with `julia_version = value`.

In other words, the only manifests that this will not work with are manifests that _aren't Julia manifests_, and we can't produce correct results then anyway.

tecosaur: Eh, I can't be bothered to discuss this any further. The `toml` crate is now used.

src/bin/julialauncher.rs
tecosaur: This is now the order used.

src/bin/julialauncher.rs
tecosaur: I'm very much on board with having exact replication be first-class.

I also think that most of the time, when a patch release comes out fixing bugs, that you want to use that.

So, ideally what I think we want to do here is to "have our cake and eat it too" by having both replication and patch-upgrades be easy. I'm not sure how best to do so.

Perhaps a prompt or a central list of per-project policies is the best way to do things?

davidanthoff: Yeah, I like the prompt idea best. So, I think essentially I would use the logic I outlined above to decide which Julia to launch, but augment it with logic that decides whether to show a "This project could be updated to a new Julia version" prompt.

I would propose that we implement that for this PR, and then the next step (but could be a new PR) could be that we detect interactive run scenarios, and in those show a prompt that offers to update the project.

> a central list of per-project policies

We might be able to avoid that if we lean on the `julia` entry in the `[compat]` section of `Project.toml`. Essentially the launcher could only show this update prompt for a project if that actually makes sense given the `julia` compat entry in the `Project.toml`. As a project author I could then completely prevent the prompt from showing by using `julia = "=1.10.3"` for example. Or if you want patch to be prompted about patch version updates, use `julia = "~1.10"`.

tecosaur: Using the `[compat]` of the `Project.toml` for project/compatability intent of Julia seems like the obvious thing with the current tools we have. I like it!

That seems to strike a good balance of concerns to me, I'm happy to go ahead with this :+1:

lgoettgens:
Maybe a stupid question: Starting with julia 1.11, I can have multiple manifest files in one folder. Which one does this feature here choose, if I have a Manifest.toml with 1.10.5, Manifest-v1.11.toml with 1.11.0 and Manifest-v1.12.toml with some 1.12-DEV version?

tecosaur:
If you don't specify the Julia version any other way, this will specifically look at the `Manifest.toml`.

lgoettgens:
> If you don't specify the Julia version any other way, this will specifically look at the `Manifest.toml`.

I must admit that this may be a reason for me to stop using `juliaup` in the future. If I set a default channel in juliaup and use the latest language features, I would still like to get the default channel. Please add a config option to opt-out/opt-in into this feature!

Image the world in a year or so (let's say after the release of 1.13). Everyone will then have `Manifest-v1.11.toml`, `Manifest-v1.12.toml`, `Manifest-v1.13.toml`, `Manifest-v1.14.toml` (for the upcoming rc), and `Manifest-v1.15.toml` (for nightly), but if one forgets to delete the old `Manifest.toml` from nowadays, one would still get 1.10 per this feature.

tecosaur:
If you want to use the latest Julia in a project, you can just update the manifest to the newer Julia version, or only have versioned manifests... :shrug:

davidanthoff:
In my mind the versioned manifest files will be more of a niche thing, at least that is how I imagine things will play out once this feature here lands.

I _do_ think we should have good UI that makes it super, super easy to update a project once a new Julia version comes out. I generally imagine that if you launch Julia with a project that the launcher will check the `julia` compat section in the `Project.toml`, cross-check that with the Julia version in the manifest, and if this project could be updated to a newer Julia version, prompt the user whether they want to update the project before the actual Julia launches. We'll need various ways to store answers etc, but roughly that is how I imagine this will work out.

tecosaur:
Indeed, I don't think David or I see this as the "final destination" so much as a massive first step to more conveniently reproducible Julia environments.

src/bin/julialauncher.rs
davidanthoff: I pushed a commit that implements my original proposal of how to decide which binary to launch.

DilumAluthge:
> In my mind the versioned manifest files will be more of a niche thing, at least that is how I imagine things will play out once this feature here lands.

I'm not sure I follow. Why would versioned manifest files be a niche thing? If you want to check your manifest files into source control, and you also want your project to support multiple Julia versions, then versioned manifest files are the only way to reliably accomplish that, right? And this seems like a pretty common use case to me.

The General registry has been using versioned manifests for our CI for a long time now. It's an important component of how we maintain CI stability on the registry. See https://github.com/JuliaRegistries/General/tree/master/.ci, where we have our `Manifest.1.$minor.toml` files.

davidanthoff:
> If you want to check your manifest files into source control, and you also want your project to support multiple Julia versions, then versioned manifest files are the only way to reliably accomplish that, right? And this seems like a pretty common use case to me.

Is it common, though? I've always thought that we generally discourage folks from checking in manifests as part of packages (which really in my mind are the main use-case where one is trying to support multiple Julia versions at the same time). But at least in my world of research, where we _do_ check-in manifests for say replication code for a paper etc, we really have zero need to support multiple Julia versions at the same time. On the contrary, we really _don't_ want to do that, we just want to have projects that pin down everything, so that anyone trying to run them gets as close as possible to using exactly the same pieces of software that was originally used, including the Julia version.

> The General registry has been using versioned manifests for our CI for a long time now.

Ok, but that strikes me as the definition of a niche use-case ;) For the LS we also do something similar, but like how many projects really fall into that kind of Julia-infrastructure category?

Having said that, even if I'm completely wrong about this, I don't really think it changes how we would want to design this feature here?

davidanthoff:
I started adding tests. They are by no means complete, i.e. we really need to come up with a much more thorough feature test suite.

BUT, I think the tests already uncovered a bug, i.e. it seems to me that the launcher at the moment isn't handing `--project` correctly. I left a comment at the relevant test. @tecosaur could you take a look?

davidanthoff:
Work on this continues now in https://github.com/JuliaLang/juliaup/pull/1087

davidanthoff:
And now it continues in https://github.com/JuliaLang/juliaup/pull/1095.

PR #1061, opened 2024-10-06
Update the LTS version to 1.10

davidanthoff:
To be merged when it is actually made official. Then needs to be followed by running the regular version db update GitHub action etc.


KristofferC:
It's been made official now.

PR #1066, opened 2024-10-08
Update progress bar style to match 1.12 progress bar

christiangnrd:
Couldn't get the smoother version without causing some issues with the colour printing.

I moved the ':' out of the prefix as it was too blue for me but I can revert that if people don't like it.

### DIFF
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -105,12 +105,12 @@ pub fn download_extract_sans_parent(
         None => ProgressBar::new_spinner(),
     };
 
-    pb.set_prefix("  Downloading:");
+    pb.set_prefix("  Downloading");
     pb.set_style(
         ProgressStyle::default_bar()
-            .template("{prefix:.cyan.bold} [{bar}] {bytes}/{total_bytes} eta: {eta}")
+            .template("{prefix:.cyan.bold}: {bar:.cyan/white} {bytes}/{total_bytes} eta: {eta}")
             .unwrap()
-            .progress_chars("=> "),
+            .progress_chars("━╸━"),
     );
 
     let last_modified = match response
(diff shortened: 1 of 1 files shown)


tecosaur:
It would be good to use both the left and right half-bar characters.

christiangnrd:
I agree, but I couldn't find how to have the right half-bar character use the alternate colour instead of the main colour. 

image

If there's an easy fix suggested, I'll integrate it with this PR. Otherwise, I suggest we merge this until a more involved solution is implemented.

christiangnrd:
> It would be good to use both the left and right half-bar characters.

christiangnrd:
Didn't mean to close sorry for the spam

tecosaur:
We're using a progress bar library? The only non-trivial component in implementing a progress bar ourselves is the ETA.

christiangnrd:
IMO this should be merged as-is for now as it's still an improvement, and I'll open an issue to improve it so that I can remember to return to it in an unknown amount of time if someone else hasn't gotten to it first.

davidanthoff:
Hm, I have to admit I find that visually way more difficult to distinguish than the existing progress bar... Everything is super small (why is this like two pixels high, only?), the difference in color between "finished" and "in progress" (blue and black) is hard to see... All the default options shown here https://crates.io/crates/indicatif (the progress bar package we use) are _way_ easier on the eye to me...

tecosaur:
Even if it's not visually great to some eyes (I suspect this is a little font-dependent too), I do think there's a fair bit of value in consistency with Pkg.

davidanthoff:
Let's hold this for now, the earliest we should merge this is when 1.12 is actually released.

KristofferC:
I think the bar can be smoothened a bit similar to how it was done in https://github.com/JuliaLang/Pkg.jl/pull/4038.

IanButterworth:
Did that here https://github.com/JuliaLang/juliaup/pull/1287

IanButterworth:
Happy to add the commit here if you agree.

christiangnrd:
Thank you! I don't know when I would have had time to smooth it out

> Happy to add the commit here if you agree.

#1287 includes my commits plus yours so I'll just close this one and the code can be finalized over there

Issue #1000, opened 2024-07-22
Warn user and get opt-in for +pr mode


IanButterworth:
On MacOS it's not possible to use the `+pr...` mode because the binaries are not codesigned, given that they are non-merged PRs.

We should do something this on MacOS
```
% juliaup add +pr51811
Installing Julia pr51811-macos-aarch64
As this is a non-merged PR it has not been codesigned. Would you like to locally codesign this to make it run? Note that it is your responsibility to review for security issues with the PR.
Locally codesign? y/[n]:
```

And perhaps on platforms that don't codesign check before installing the files
```
% juliaup add +pr51811
This is a PR that may not have been merged. Note that it is your responsibility to review for security issues with the PR.
Continue installation? y/[n]:
```

---

In testing this I signed with:

sign_binaries.sh
```
#!/bin/bash

# Function to sign a binary
sign_binary() {
    local binary="$1"
    echo "Signing $binary"
    codesign --sign - "$binary"
}

# Export the function for use with find
export -f sign_binary

# Find and sign all binaries in the current directory and its subdirectories
find . -type f -perm +111 -exec bash -c 'sign_binary "$0"' {} \;

echo "All binaries signed."
```

But note that stdlibs will re-precompile because codesigning changes their mtime/hash (on buildkite we update the .ji cache file headers after codesigning)

Issue #1068, opened 2024-10-10
Juliaup Crash Report


LilithHafner:
[report-e6e73013-157e-4c21-bee7-638b11117bb5.toml.txt](https://github.com/user-attachments/files/17320563/report-e6e73013-157e-4c21-bee7-638b11117bb5.toml.txt)

fredrikekre:
Got the same, pasting it here to help search:
```
"name" = "Juliaup"
"operating_system" = "Mac OS 14.6.1 [64-bit]"
"crate_version" = "1.17.4"
"explanation" = """
Panic occurred in file 'src/command_update.rs' at line 83
"""
"cause" = "assertion failed: !version.is_empty()"
"method" = "Panic"
"backtrace" = """

   0: 0x10491fc44 - core::panicking::panic::h61ea408fdd25f03d
   1: 0x1047da4d4 - juliaup::command_update::update_channel::h3737618e1bbe561d
   2: 0x1046b51b0 - juliaup::main::hcafec2366675f9f8
   3: 0x1046a9224 - std::sys_common::backtrace::__rust_begin_short_backtrace::h6f69fd9b04946a92
   4: 0x1046b87fc - _main"""
```

davidanthoff:
Can you post the content of your `~/.julia/juliaup/juliaup.json` file?

LilithHafner:
[juliaup.json](https://github.com/user-attachments/files/17351866/juliaup.json)

davidanthoff:
And you are still getting the same error? I'm confused... The stack trace indicates that we have a direct download channel that has an empty `version` element, but at least in @LilithHafner's config that doesn't seem to be the case...

christiangnrd:
This happened to me on macos after trying to install a pr channel because juliaup is not able to launch julia to get a version.

fredrikekre:
That's probably it then because I have tried adding a PR.

LilithHafner:
This happened to me again when trying to update:

```
x@x ~ % juliaup update
Installing Julia 1.11.1+0.aarch64.apple.darwin14
Checking standard library notarization............................................................done.
Well, this is embarrassing.

Juliaup had a problem and crashed. To help us diagnose the problem you can send us a crash report.

[...]

PR #1084, opened 2024-10-25
Prevent compiler warning without panicking

christiangnrd:
My attempt at resolving #1068.

The source of the panic seems to be an assertion that was added purely to remove a compiler warning, so I turned it into a warning because I don't think an empty version field should prevent updating to a latest PR.

TODO: If people decide this is an acceptable solution, I'll improve the warning message with potential causes.

One cause is when an installed channel has an empty version field due to not being able to get the version number from a non-codesigned executable on macOS https://github.com/JuliaLang/juliaup/issues/1068#issuecomment-2409141076. It could also be that a particular PR was merged https://github.com/JuliaLang/juliaup/issues/1068#issuecomment-2435179919.

### DIFF
--- a/src/command_update.rs
+++ b/src/command_update.rs
@@ -82,10 +82,11 @@ fn update_channel(
             server_etag,
             version,
         } => {
-            // We only do this so that we use `version` on both Windows and Linux to prevent a compiler warning/error
-            assert!(!version.is_empty());
-
             if local_etag != server_etag {
+                // We only do this so that we use `version` on both Windows and Linux to prevent a compiler warning/error
+                if version.is_empty() {
+                    eprintln!("Channel {} version is empty, something may be wrong.", channel);
+                }
                 eprintln!("{} channel {}", style("Updating").green().bold(), channel);
 
                 let channel_data =
--- a/src/command_update.rs
+++ b/src/command_update.rs
@@ -85,7 +85,10 @@ fn update_channel(
             if local_etag != server_etag {
                 // We only do this so that we use `version` on both Windows and Linux to prevent a compiler warning/error
                 if version.is_empty() {
-                    eprintln!("Channel {} version is empty, something may be wrong.", channel);
+                    eprintln!(
+                        "Channel {} version is empty, something may be wrong.",
+                        channel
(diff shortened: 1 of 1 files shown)


IanButterworth:
> not being able to get the version number from a non-codesigned executable on macOS

Good call. Although I tried adding a `pr` channel, julia didn't load because of this.
I feel we should make that clearer to the user somehow.

davidanthoff:
So if the underlying issue at the moment is that pr channels just don't work, then maybe we should (in the very short run) disable that feature on MacOS, and then try to properly fix https://github.com/JuliaLang/juliaup/issues/1000.

But I think we should not just start to ignore these empty versions at this place here, instead we should probably just abort installation of a PR channel if we can't launch the executable to extract the version in the first place.

src/command_update.rs
```diff
@@ -82,10 +82,14 @@ fn update_channel(
...
             version,
         } => {
-            // We only do this so that we use `version` on both Windows and Linux to prevent a compiler warning/error
-            assert!(!version.is_empty());
-
             if local_etag != server_etag {
+                // We only do this so that we use `version` on both Windows and Linux to prevent a compiler warning/error
+                if version.is_empty() {
+                    eprintln!(
+                        "Channel {} version is empty, something may be wrong.",
```
christiangnrd, line 89:
```suggestion
                        "Channel {} version is empty, you may need to manually codesign this channel if you trust the contents of this pull request.",
```

christiangnrd:
> So if the underlying issue at the moment is that pr channels just don't work, then maybe we should (in the very short run) disable that feature on MacOS, and then try to properly fix #1000.
> 
> But I think we should not just start to ignore these empty versions at this place here, instead we should probably just abort installation of a PR channel if we can't launch the executable to extract the version in the first place.

While I agree #1086 is a much better solution. I think that this PR would be a better very-short-term solution than disabling the feature for mac users because it prevents a crash when trying to update, and still allows to download/update a PR branch. Users can then manually codesign the downloaded PR if they want.

davidanthoff (approved):
You convinced me :)

PR #1086, opened 2024-10-26
Code sign pr binaries on macOS

davidanthoff:
Fixes https://github.com/JuliaLang/juliaup/issues/1000.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1032,6 +1032,7 @@ dependencies = [
  "tempfile",
  "thiserror",
  "url",
+ "walkdir",
  "windows",
  "winres",
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,7 @@ windows = { version = "0.58.0", features = ["Win32_Foundation", "Win32_UI_Shell"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 reqwest = { version = "0.12", default-features = false, features = ["blocking", "native-tls", "socks"] }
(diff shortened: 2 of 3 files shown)


davidanthoff:
@IanButterworth could you give this a try? And maybe you know what we need to check for with the return value from `mode()` to really just target the executable files?

IanButterworth:
As demonstrated in https://github.com/JuliaLang/juliaup/issues/1000#issue-2423024139 locally codesigning causes all stdlib pkgimages to be invalidated, meaning the user will hit precompilation for stdlibs, which will go into the user depot, which I believe creates non-ideal cache loading state. If we do want to do this, then we'd need to also employ the post-codesign fix that julia CD does by updating the hash in the .ji files too.

https://github.com/JuliaCI/julia-buildkite/blob/32243903bbc9e282da801d6160e6ee1009e6d0a4/utilities/upload_julia.sh#L42

davidanthoff:
Oh gee, this all feels very brittle... But ok. Probably best if we re-implement that hash-update functionality in Rust? It doesn't look too complicated... I think we can't assume that we have a functioning Julia available at the point in time when we need to make this change.

I should say, I'll probably leave this to someone else to finish :)

IanButterworth:
FWIW, I'm not that someone else. To re-write that .jl header update in rust, we'd need to keep rust up to date with the .ji format.. which is purely internal, so prone to change.

IanButterworth:
This was implemented in https://github.com/JuliaLang/juliaup/pull/1277

PR #1095, opened 2024-10-29
Launch the manifest-specified Julia version v3

davidanthoff:
Take 3, because I accidentally merged take 2 prematurely, then had to undo that, and now we should be back to "we can finish this PR here and then merge".


tecosaur:
I don't think it's realistic that I'll get around to writing the missing tests, let alone resolving the merge conflicts any time soon.

Issue #1096, opened 2024-11-02
`juliaup update` is initially slow to check versions db


IanButterworth:
I've been looking into why `juliaup update` hangs for a while before doing any substantial downloads, even on very fast connections.

I think the issue is that this file may be served via a CDN that needs its caching settings tweaked to be less lazy
https://julialang-s3.julialang.org/juliaup/RELEASECHANNELDBVERSION

It's a very small file, so doesn't make any sense that it takes 7s to download.. so I believe the time is the CDN refreshing a local cache.

--- 

From a profile during that quiet ~10s it can be seen that `update_version_db` is spending 7s waiting on `reqwest`.

![Image](https://github.com/user-attachments/assets/af5baf36-ccb4-49b6-99a2-e1b9d5f43d7a)

cc. @giordano @davidanthoff

davidanthoff:
Yes, that makes a ton of sense: the idea is that this file is small so that it can be checked really quickly (clearly not working...). The file is _also_ configured to _not_ be cached by the CDN so that updates are picked up by clients quickly.

I'm not sure what the best strategy is, probably really depends on the CDN, right?

IanButterworth:
Actually, I caught the slowdown with timing printing added and it seems to be something else.

This is with https://github.com/JuliaLang/juliaup/compare/main...ib/timing

Note I set the db patch version back one, just to exercise all code paths, but that is only hit after the problematic slow one.

```
% ./target/release/juliaup update
Checking for new Julia versions
Online db version `https://julialang-s3.julialang.org/juliaup/RELEASECHANNELDBVERSION`: Version { major: 1, minor: 0, patch: 55 }
Request to 'https://julialangnightlies-s3.julialang.org/bin/macos/aarch64/julia-latest-macos-aarch64.tar.gz' took 5.979535584s
Direct download etags: [("nightly", "\"2eeb5e6ccc5363b5180feb502f96d919-27\"")]
Bundled db version: Version { major: 1, minor: 0, patch: 54 }
Updating channel nightly
```
Then immediately after
```
ian@Ians-MacBook-Pro juliaup % ./target/release/juliaup update
Checking for new Julia versions
Online db version `https://julialang-s3.julialang.org/juliaup/RELEASECHANNELDBVERSION`: Version { major: 1, minor: 0, patch: 55 }
Request to 'https://julialangnightlies-s3.julialang.org/bin/macos/aarch64/julia-latest-macos-aarch64.tar.gz' took 112.53425ms
[...]

PR #1097, opened 2024-11-02
If nightly or pr check is going slowly inform user

IanButterworth:
I just happened to hit this in the wild:

Screenshot 2024-11-03 at 9 53 05 AM

Helps explain https://github.com/JuliaLang/juliaup/issues/1096 to the user

### DIFF
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -32,11 +32,15 @@ use std::{
     io::{BufReader, Read, Seek, Write},
     path::{Path, PathBuf},
 };
+use std::sync::mpsc::channel;
+use std::thread;
+use std::time::Duration;
 #[cfg(not(target_os = "freebsd"))]
 use tar::Archive;
 use tempfile::Builder;
 use tempfile::TempPath;
 use url::Url;
+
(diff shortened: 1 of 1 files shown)


IanButterworth:
weird.. the formatter suggested removing imports that broke the code..

IanButterworth:
Bump

davidanthoff:
I guess we can merge this for now, but somehow this makes me think that we maybe should just adopt `tokio` for the `juliaup` binary and then do this using async features... For example, the Windows download API is already completely async, so it really shouldn't be necessary to use extra threads to do something like that, and `reqwest` also has a fully async API, so same story.

Plus, we already have some places where we for example start a number of parallel downloads via async, but those things could presumably also be simplified if we had a proper async runtime.

IanButterworth:
Merge?

IanButterworth:
Bump for having this released.

IanButterworth:
@davidanthoff I will look into making a new release at the end of the week. Let me know if I shouldn't. Thanks.

Issue #343, opened 2022-07-29
Alternative to Windows Store for Windows


NicholasWMRitchie:
My work computer is forbidden by policy to access the Windows Store.  Could there be an alternative source for JuliaUp for Windows as there is for Linux and Mac?

davidanthoff:
@NicholasWMRitchie Could you try a new install option that avoids the store? Here is how:
- Click on https://install.julialang.org/Julia.appinstaller
- Launch the file that gets downloaded (just "open" it or double click on it in explorer)

Be warned, the version that you are getting there is very much a dev version, and you probably will have to uninstall this and then re-install a more robust version down the road. Right now I'm just trying to figure out whether this option is one that could solve this scenario for us.

NicholasWMRitchie:
I did as you instructed.   It installs a 64-bit instance of Julia 1.7.3 and the instance works as expected.
I could not discover any way to install different Julia versions as the more complete Windows Store option does.

lungben:
Thanks, this is really helpful!

I suggest to add the link to the appinstaller to the Readme file of this repo so that it is easier to find for other users.

davidanthoff:
> I suggest to add the link to the appinstaller to the Readme file of this repo so that it is easier to find for other users.

For now it is an experiment, as soon as we decide that we want to have that option permantely I'll put a link in the README. Oh, and we need to change code signing from my personal cert to the JuliaComputing one as well before we really declare this _done_.

NicholasWMRitchie:
Is there a reason that the "juliaup self update" is not implemented on this version or is this just one of those things that you still need to complete?

I just reinstalled Julia.appinstaller to get the latest 1.8.0-rc4 version as it wasn't available on the initial test version.

davidanthoff:
> Is there a reason that the "juliaup self update" is not implemented on this version or is this just one of those things that you still need to complete?

Yep, that still needs to be implemented. I think in theory this appinstaller distributed version should auto-check for new versions, and then auto-update in the background, but there might be delays in that. Once I implement the `juliaup self update` thing that should force-trigger that.

I also plan to change how Juliaup in general learns about new Julia versions, at that point the delays should be minimal.

bilderbuchi:
>  I think in theory this appinstaller distributed version should auto-check for new versions, and then auto-update in the background, but there might be delays in that. 

[...]

PR #1104, opened 2024-11-10
docs: Include scoop method of installation in README

akielbowicz:
Hi, 
A few days ago in juliaup can be installed using [scoop](https://scoop.sh). It was enabled with https://github.com/ScoopInstaller/Main/pull/6210

Personally I been using scoop for a few years already on my personal and my work Windows machines and never had any issue. I prefer scoop over winget for cli tools as it eases the way to manage it and find issues.

Hopefully this helps with #343

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,13 @@ winget install julia -s msstore
 
 on a command line.
 
-If the Windows Store is blocked on a system, we have an alternative [MSIX App Installer](https://learn.microsoft.com/en-us/windows/msix/app-installer/app-installer-file-overview) based setup. Note that this is currently experimental, please report back successes and failures [here](https://github.com/JuliaLang/juliaup/issues/343). To use the App Installer version, download [this](https://install.julialang.org/Julia.appinstaller) file and open it by double clicking on it.
+Alternatively if the Windows Store is blocked on a system, or you don´t have Admin Rights on the system, you can install it through [scoop](https://scoop.sh) by execution in a PowerShell:
+
+```powershell
+scoop install main/juliaup
+```
+
+Other alternatives include installing directly with [MSIX App Installer](https://learn.microsoft.com/en-us/windows/msix/app-installer/app-installer-file-overview) based setup. Note that this is currently experimental, please report back successes and failures [here](https://github.com/JuliaLang/juliaup/issues/343). To use the App Installer version, download [this](https://install.julialang.org/Julia.appinstaller) file and open it by double clicking on it.
 
 If neither the Windows Store nor the App Installer version work on your Windows system, you can also use a MSI based installer. Note that this installation methods comes with serious limitations and is generally not recommended unless no other method works. For example, there is no automatic update mechanism for Juliaup with this installation method. The 64 bit version of the MSI installer can be downloaded from [here](https://install.julialang.org/Julia-x64.msi) and the 32 bit version from [here](https://install.julialang.org/Julia-x86.msi). By default the install will be a per-user install that does not require elevation. You can also do a system install by running the following command from a shell:
 


davidanthoff:
Couple of thoughts:
- Is there a plan to update the version that is available via scoop when we release new builds? For other platforms we automatically update things as part of the Juliaup release process (https://github.com/JuliaLang/juliaup/blob/main/.github/workflows/release.yml), we should probably do the same for scoop, right?
- If we actually have a package manager use the portable builds, we should probably upload them to S3 and then download from there, rather than from GitHub directly, but also see below about other options.
- Could this be listed as `julia` rather than `juliaup` in the scoop repo? So that the install command is `scoop install main/julia`. That is what we already have for `winget`, and given that Juliaup now is the official installer for Julia, I think that would make things much clearer.
- Does scoop support MSI or MSIX installers? If so, maybe we should use those instead? They handle more things properly, like modifying `PATH`, adding start menu items, potentially down the road package identity etc. And they exist and are available at the various official download pages.
- I'm happy to mention the scoop option here, but we should _not_ make it the next best option if the Windows Store doesn't work. At the end of the day, the "official" story should be: 1) Windows Store, 2) App installer, 3) MSI, 4) all the other third-party options (which include scoop).

akielbowicz:
Hi @davidanthoff, thanks for your thoughts

> Is there a plan to update the version that is available via scoop when we release new builds? For other platforms we automatically update things as part of the Juliaup release process (main/.github/workflows/release.yml), we should probably do the same for scoop, right?
If we actually have a package manager use the portable builds, we should probably upload them to S3 and then download from there, rather than from GitHub directly, but also see below about other options.

Scoop does update the packages versions manifest automatically through some scheduled process. The new version information can be obtained through different mechanisms, see [App Manifest Autoupdate](https://github.com/ScoopInstaller/Scoop/wiki/App-Manifest-Autoupdate). Currently for Juliaup is configured to get the information from the GitHub Releases, as you can see in the [manifest commit history](https://github.com/ScoopInstaller/Main/commits/aa5dad6e7b7520c5418cece4706ed9de4b8b739b/bucket/juliaup.json) that last week the manifest it was updated to the version **1.17.9**. 

Related to the autoupdate I decided to download from the GitHub Release directly because they were the only ones that had the version on the name, the links detailed on the README do not have the version information, (ie "https://install.julialang.org/Julia-x64.msi"). I suppose that this could be fixed by changing the [Rename Step](https://github.com/JuliaLang/juliaup/blob/b0cfb9e4fa3b20b23964dfea446cf4e58bcf791d/.github/workflows/release.yml#L1179) to copy the files instead of renaming it, in favor of not breaking anything that depends on the names as they are now.

> Could this be listed as julia rather than juliaup in the scoop repo? So that the install command is scoop install main/julia. That is what we already have for winget, and given that Juliaup now is the official installer for Julia, I think that would make things much clearer.

I will say that it's possible but it might bring some issues, there has been a [julia](https://github.com/ScoopInstaller/Main/commit/98fb72125ea5bff2e352c4087c5d50d21b5789ad) manifest since 2014 and I suspect that changing it will break other scoop users setups. 

> Does scoop support MSI or MSIX installers? If so, maybe we should use those instead? They handle more things properly, like modifying PATH, adding start menu items, potentially down the road package identity etc. And they exist and are available at the various official download pages.

Yes, the MSI installer seems to work without any issues. For example, the [rust](https://github.com/ScoopInstaller/Main/blob/aa5dad6e7b7520c5418cece4706ed9de4b8b739b/bucket/rust.json#L15) manifest uses it. As mentioned above, the MSI files must be published with the version in their name.

> I'm happy to mention the scoop option here, but we should not make it the next best option if the Windows Store doesn't work. At the end of the day, the "official" story should be: 1) Windows Store, 2) App installer, 3) MSI, 4) all the other third-party options (which include scoop).

Of course, I updated the PR to mention the Scoop installation under the *Software Repositories* section

PR #1108, opened 2024-11-14
Update .cshrc and .tcshrc if present

### LINKED ISSUE #1107: Add `tcsh` support
It would be great if besides `zsh` and `bash` (resp. `sh` compatible shells) also `tcsh` was supported. That would mean editing `~/.tcshrc` and `~/.cshrc` if they exist, and inserting suitable code there.

Asking because I just demoed `juliaup` to someone who has been a long year `tcsh` user ;-).

(`tcsh` used to be the default shell on Mac OS X for the first couple years).

fingolfin:
Resolves #1107

### DIFF
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -1050,6 +1050,8 @@ fn get_shell_script_juliaup_content(bin_path: &Path, path: &Path) -> Result<Vec<
     result.extend_from_slice(HEADER);
     if path.file_name().unwrap() == ".zshrc" {
         append_zsh_content(&mut result, bin_path_str);
+    } else if path.file_name().unwrap() == ".cshrc" || path.file_name().unwrap() == ".tcshrc" {
+        append_csh_content(&mut result, bin_path_str);
     } else {
         append_sh_content(&mut result, bin_path_str);
     }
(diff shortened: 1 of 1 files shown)


fingolfin:
@IanButterworth OK to merge?

src/operations.rs
```diff
@@ -1587,6 +1589,18 @@ fn append_zsh_content(buf: &mut Vec<u8>, path_str: &str) {
     buf.extend_from_slice(content.as_bytes());
 }
 
+fn append_csh_content(buf: &mut Vec<u8>, path_str: &str) {
+    // csh specific syntax for path extension
+    let content = formatdoc!(
+        "
+            set path = ($path {})
```
IanButterworth: Claude:
The juliaup bin directory should be prepended, not appended, so that juliaup's `julia` takes priority over any system-installed `julia`. This matches the behavior of `append_zsh_content` (`path=('{}' $path)`) and `append_sh_content` (`export PATH={}${PATH:+:${PATH}}`).
```suggestion
            set path = ({} $path)
```

fingolfin: yes my bad, thanks!

IanButterworth (approved):

PR #1138, opened 2025-01-12
feat: add fmt to status command

JosiahParry:
This PR adds an optional `fmt` command to `juliaup status` supporting `json`, `ndjson`, `csv`, and `tsv` formats. 

The intent is to allow other processes to consume the output of `juliaup`'s status command. 

In my particular use case the goal is to be able to run a julia script using the closest available version of julia possible. 

For example a `Manifest.toml` may have 

```toml
julia_version = "1.8.3"
```

And `juliaup status` may show 
```
$ juliaup st
 Default  Channel  Version                          Update
-----------------------------------------------------------
          1.8.2    1.8.2+0.aarch64.apple.darwin14
       *  1.11     1.11.2+0.aarch64.apple.darwin14
          1.11.1   1.11.1+0.aarch64.apple.darwin14
```
In this case I want to run version `1.8.2` because it should still be minor version compatible. However using juliaup's julia binary with `julia +1.8` we get an error: 

```
ERROR: Invalid Juliaup channel ``. Please run `juliaup list` to get a list of valid channels and versions.
```

With the `-f json` format I can parse the output using `serde_json` and manually match the versions using `semver` to order on precedence. 

```rs
use serde::{Deserialize, Serialize};
use std::{
    error::Error,
    process::{Command, Stdio},
};

#[derive(Serialize, Deserialize, Debug)]
struct JuliaChannel {
    name: String,
    version: String,
}

fn main() -> Result<(), Box<dyn Error>> {
    let output = Command::new("juliaup")
        .args(["status", "-f", "json"])
        .stdout(Stdio::piped())
        .spawn()?
        .wait_with_output()?;

    let vers = serde_json::from_slice::<Vec<JuliaChannel>>(&output.stdout)?;
    dbg!(&vers);
    Ok(())
}
```
```
[src/main.rs:21:5] &vers = [
    JuliaChannel {
        name: "1.8.2",
        version: "1.8.2+0.aarch64.apple.darwin14",
    },
    JuliaChannel {
        name: "1.11",
        version: "1.11.2+0.aarch64.apple.darwin14",
    },
    JuliaChannel {
        name: "1.11.1",
        version: "1.11.1+0.aarch64.apple.darwin14",
    },
]
```

### DIFF
--- a/src/bin/juliaup.rs
+++ b/src/bin/juliaup.rs
@@ -94,7 +94,7 @@ fn main() -> Result<()> {
         Juliaup::Default { channel } => run_command_default(&channel, &paths),
         Juliaup::Add { channel } => run_command_add(&channel, &paths),
         Juliaup::Remove { channel } => run_command_remove(&channel, &paths),
-        Juliaup::Status {} => run_command_status(&paths),
+        Juliaup::Status(args) => run_command_status(&paths, args.fmt),
         Juliaup::Update { channel } => run_command_update(channel, &paths),
         Juliaup::Gc { prune_linked } => run_command_gc(prune_linked, &paths),
         Juliaup::Link {
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,3 +1,4 @@
+use crate::command_status::StatusArgs;
 use clap::{Parser, ValueEnum};
 
 #[derive(Parser)]
@@ -31,7 +32,7 @@ pub enum Juliaup {
     Remove { channel: String },
     #[clap(alias = "st")]
     /// Show all installed Julia versions
-    Status {},
+    Status(StatusArgs),
     /// Garbage collect uninstalled Julia versions
     Gc {
         #[clap(long)]
--- a/src/command_status.rs
+++ b/src/command_status.rs
@@ -3,6 +3,7 @@ use crate::config_file::JuliaupConfigChannel;
 use crate::global_paths::GlobalPaths;
 use crate::versions_file::load_versions_db;
 use anyhow::{Context, Result};
+use clap::{Args, ValueEnum};
 use cli_table::format::HorizontalLine;
 use cli_table::format::Separator;
 use cli_table::ColorChoice;
@@ -12,8 +13,10 @@ use cli_table::{
 };
 use human_sort::compare;
 use itertools::Itertools;
+use serde::Deserialize;
+use serde::Serialize;
 
-#[derive(Table)]
+#[derive(Table, Serialize, Deserialize)]
 struct ChannelRow {
     #[table(title = "Default", justify = "Justify::Right")]
     default: &'static str,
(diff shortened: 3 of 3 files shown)


JosiahParry:
Please note that CI is failing due to a check in `julialauncher` which is not modified in this PR.

davidanthoff:
Can you try `juliaup api getconfig1`, would that provide what you need?

The general idea is that anything that is not hidden behind the `api` command is not promised to be stable across versions, so not ideal for tools to rely on. But the (hidden) commands behind the `api` command are meant for other tools to interact with, and the idea is that those are stable and will never change.

Currently the VS Code extension also uses that command to learn about existing Julia versions.

JosiahParry:
in the event anyone else needs to accomplish this using `juliaup api getconfig1`.

I can close this, but I do think having a format option similar to cargo would be quite nice :) 

```rs
/// Represent Julia Channel from `juliaup api getconfig1` output
/// see https://github.com/JuliaLang/juliaup/pull/1138
#[derive(Serialize, Deserialize, Debug)]
#[serde(rename_all = "PascalCase")]
pub struct JuliaChannelUnprocessed {
    pub name: String,
    pub version: String,
    #[serde(rename = "File")]
    path: String,
}

#[derive(Serialize, Deserialize, Debug)]
#[serde(rename_all = "PascalCase")]
pub struct JuliaConfigUnprocessed {
    pub default_channel: JuliaChannelUnprocessed,
    pub other_channels: Vec<JuliaChannelUnprocessed>,
}

impl JuliaConfigUnprocessed {
    pub fn as_channels(self) -> Result<JuliaChannels, semver::Error> {
        let mut res = Vec::with_capacity(1 + self.other_channels.len());
        res.push(self.default_channel.try_into()?);
        for channel in self.other_channels {
            res.push(channel.try_into()?);
        }
        Ok(JuliaChannels(res))
    }
}

/// Represents a julia channel based on the name and version
#[derive(Debug, Clone)]
pub struct JuliaChannel {
    pub name: String,
    pub version: Version,
    pub path: String,
}

impl TryFrom<JuliaChannelUnprocessed> for JuliaChannel {
    type Error = semver::Error;

    fn try_from(value: JuliaChannelUnprocessed) -> Result<Self, Self::Error> {
        let res = JuliaChannel {
            name: value.name,
            version: Version::parse(&value.version)?,
            path: value.path,
        };

        Ok(res)
    }
}

#[derive(Debug, Clone)]
pub struct JuliaChannels(pub Vec<JuliaChannel>);

pub fn discover_julia_channels() -> anyhow::Result<JuliaChannels> {
    let output = Command::new("juliaup")
        .args(["api", "getconfig1"])
        .stdout(Stdio::piped())
        .spawn()?
        .wait_with_output()?;
    let vers = serde_json::from_slice::<JuliaConfigUnprocessed>(&output.stdout)?;
    Ok(vers.as_channels()?)
}

ericphanson:
it seems weird that `api` is the stable command but it is hidden. To me hidden would imply undocumented / unstable / internal, not "the preferred api for tools"

PR #1145, opened 2025-01-17
support `x.y-nightly` channel to allow testing nightlies on release branches

KristofferC:
Think I finally managed to make a PR from my fork...

### DIFF
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -502,17 +502,33 @@ pub fn is_pr_channel(channel: &String) -> bool {
     return Regex::new(r"^(pr\d+)(~|$)").unwrap().is_match(channel);
 }
 
+fn parse_nightly_channel_or_id(channel: &str) -> Option<String> {
+    let nightly_re =
+        Regex::new(r"^((?:nightly|latest)|latest|(\d+\.\d+)-(?:nightly|latest))").unwrap();
+
+    let caps = nightly_re.captures(channel)?;
+    if let Some(xy_match) = caps.get(2) {
+        Some(xy_match.as_str().to_string())
+    } else {
+        Some("".to_string())
+    }
+}
+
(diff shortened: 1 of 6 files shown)


src/operations.rs
```diff
@@ -621,41 +637,86 @@ pub fn install_non_db_version(
...
-    let arch = parts.next().expect("Failed to parse channel name.");
-    let download_url_path = if id == "latest" {
+
+    let mut id = parts
+        .next()
+        .expect("Failed to parse channel name.")
+        .to_string();
+    let mut arch = parts.next().expect("Failed to parse channel name.");
+
+    // Check for the case where name is given as "x.y-latest-...", in which case
```
KristofferC, line 649: I'm not sure this is the most elegant way to do this.

LilithHafner:
I agree that this is a good feature to have.

If we want to support the `master` branch and the `release-x.y` branch, perhaps we should support all branches? The channel `nightly[-{BRANCH}]` could refer to whatever is the most recent build on `{BRANCH}` (with the default being the default branch). This would make the branch name here `nightly-release-x.y` instead of `x.y-nightly`.

davidanthoff:
@LilithHafner I like the idea! @KristofferC maybe just change the PR for now to use `nightly-release-x.y` but not implement the full general thing for now, i.e. that could come in a future PR? But really up to you.

Otherwise it looks good to me! That is the nice thing about Rust: if it compiles, it often is done ;)

KristofferC:
I prefer having it like this to have it consistent with e.g. the syntax in Github Actions. I don't think we make nightlies for other than the release branches.

davidanthoff (approved):
Ok, I have to admit, I don't really care :) From my end this can merge either way. I also don't consider these kind of things of any "public api" of Juliaup, so we can also change down the road.

Maybe we wait for a few more days if anyone else feels strongly about this, and then merge?

Issue #1129, opened 2024-12-15
gc doesn't work for pr-{number} channels


LilithHafner:
Somehow I ended up with a lot of julia-pr{number} folders taking up space in my ~/.julia/juliaup folder. I'm assuming juliaup created them and failed to gc them:
```
x@x:~$ juliaup status
 Default  Channel  Version                              Update           
-------------------------------------------------------------------------
          1        1.11.2+0.aarch64.linux.gnu                            
          1.0      1.0.5+0.aarch64.linux.gnu                             
          1.3      1.3.1+0.aarch64.linux.gnu                             
          1.6      1.6.7+0.aarch64.linux.gnu                             
          1.9      1.9.4+0.aarch64.linux.gnu                             
          1.10     1.10.7+0.aarch64.linux.gnu                            
          1.11     1.11.2+0.aarch64.linux.gnu                            
          lts      1.10.7+0.aarch64.linux.gnu                            
          nightly  Development version 1.12.0-DEV.1717  Update available 
       *  release  1.11.2+0.aarch64.linux.gnu                            
x@x:~$ juliaup gc
x@x:~$ du -hd1 .julia/juliaup/
373M	.julia/juliaup/julia-1.6.7+0.aarch64.linux.gnu
257M	.julia/juliaup/julia-1.0.5+0.aarch64.linux.gnu
480M	.julia/juliaup/julia-1.9.4+0.aarch64.linux.gnu
446M	.julia/juliaup/julia-temp-KCfKZ7
432M	.julia/juliaup/julia-1.8.5+0.aarch64.linux.gnu
373M	.julia/juliaup/julia-1.6.0+0.aarch64.linux.gnu
229M	.julia/juliaup/julia-temp-cxrr5B
403M	.julia/juliaup/julia-1.7.3+0.aarch64.linux.gnu
1.1G	.julia/juliaup/julia-pr55599
1018M	.julia/juliaup/julia-pr55754
993M	.julia/juliaup/julia-pr55988
992M	.julia/juliaup/julia-pr55970
1.1G	.julia/juliaup/julia-pr55869
1.1G	.julia/juliaup/julia-pr55165
610M	.julia/juliaup/julia-1.10.5+0.aarch64.linux.gnu
995M	.julia/juliaup/julia-pr56113
285M	.julia/juliaup/julia-1.3.1+0.aarch64.linux.gnu
262M	.julia/juliaup/julia-temp-3JIZev
[...]

PR #1146, opened 2025-01-17
Prune versions not referenced in `juliaup.json` with `--prune-orphans`

christiangnrd:
I took a look at #1129 and it seems like if a direct download channel was deleted and the .juliaup folder was not, it would forever be ignored by the `gc` command. I attempt to address it by adding a `--prune-orphans` flag similar to the `--prune-linked` flag that will remove any folder that starts with "julia-" that is not referenced by the "juliaup.json" config file.

I also add some clarifying comments to the SystemChannel code to make it easier to understand.

### DIFF
--- a/src/bin/juliaup.rs
+++ b/src/bin/juliaup.rs
@@ -96,7 +96,10 @@ fn main() -> Result<()> {
         Juliaup::Remove { channel } => run_command_remove(&channel, &paths),
         Juliaup::Status {} => run_command_status(&paths),
         Juliaup::Update { channel } => run_command_update(channel, &paths),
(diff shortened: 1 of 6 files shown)


ghyatzo:
#1153 could be related to this.
There seems to be some path normalization shenanigans going on. Have you ever encountered this?

christiangnrd:
Now that you mention it, I do remember noticing that while writing this PR, but with how busy the maintainers are, I try (emphasis on try) to make my PRs as easy to merge and with the least back and forth as possible

christiangnrd:
Closing because https://github.com/JuliaLang/juliaup/pull/1205 was merged!

PR #1189, opened 2025-03-28
Only check for direct download versions when necessary

### LINKED ISSUE #1170: Slow, unnecessary version check updates on unrelated command
When installing channel 1.8, I don't want to check for updates to other channels. Especially not if that slows down the instillation of the channel I'm asking for.

In this case, the "Checking for new version" steps took longer than the "Installing" step.
```
$ julia +1.8
ERROR: `1.8` is not installed. Please run `juliaup add 1.8` to install channel or version.
$ juliaup add 1.8
Checking for new version on channel 'pr55152' is taking a while... This can be slow due to server caching
Checking for new version on channel 'nightly' is taking a while... This can be slow due to server caching
Checking for new version on channel 'pr56882' is taking a while... This can be slow due to server caching
Checking for new version on channel 'pr56707' is taking a while... This can be slow due to server caching
Installing Julia 1.8.5+0.aarch64.linux.gnu
```

christiangnrd:
Closes #1170

Ironically, when adding a direct download version, `update_version_db` is not run, which means that having a nightly or pr channel only slowed down adding stable channels, but not other direct download channels


christiangnrd:
FreeBSD failure unrelated?

davidanthoff (approved):
Awesome, thanks! I'm merging this as is, and then the other option you mentioned is probably not needed anymore?

christiangnrd:
It would avoid querying the internet to update the version db when trying to add an already-installed channel, but I guess people can just check beforehand if they have an internet connection where this would significantly slow things down.

Probably not worth the trouble.

PR #1205, opened 2025-05-10
Don't orphan old julia versions when failing to delete them.

### LINKED ISSUE #1153: Julia versions assumed removed even if failing to remove them, leads to orphaned versions.
Hello,

I've been getting these errors: `WARNING: Failed to delete C:\Users\User\.julia\juliaup\.\julia-1.11.2+0.x64.w64.mingw32. You can try to delete at a later point by running 'juliaup gc'.` after installing new versions

interestingly there seems to be an extra `.` that is not normalized (after `juliaup`).

also, running `juliaup gc` just returns, without deleting or printing anything.
It seems like there is a problem with path normalization.

ghyatzo:
Alternative to #1146 fixes #1153 

only remove a version from the installed version database if we successfully deleted them.
When failing to remove a version, at least on windows, it is most likely that that version is still running somewhere.
Added a bit of text to hint at that possibility.

With this change `juliaup gc` works as intended after that message.

regarding #1146: it might still be useful to have orphan pruning capabilities as a fallback, but i've opted to keep the two PRs separate for easier review and merging

### DIFF
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -770,10 +770,15 @@ pub fn garbage_collect_versions(
             let path_to_delete = paths.juliauphome.join(&detail.path);
             let display = path_to_delete.display();
 
-            if std::fs::remove_dir_all(&path_to_delete).is_err() {
-                eprintln!("WARNING: Failed to delete {}. You can try to delete at a later point by running `juliaup gc`.", display)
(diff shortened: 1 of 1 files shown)


davidanthoff:
Awesome, thanks!

PR #1218, opened 2025-07-14
Auto install versions via channel selection

IanButterworth:
Makes it such that if you specify a channel which you don't have installed but is valid, it installs and launches it, with a message.
```
% julia +nightly
Info: Installing Julia nightly as it was requested but not installed.
Installing Julia latest-macos-aarch64
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.13.0-DEV.860 (2025-07-12)
 _/ |\__'_|_|_|\__'_|  |  Commit 6ddb3d64106 (1 day old master)
|__/                   |

julia>
```

Works for all valid channel descriptors (including PRs).

```
% julia +badchannel
ERROR: Invalid Juliaup channel `badchannel`. Please run `juliaup list` to get a list of valid channels and versions.
```

Developed with Claude.

----

Note unrelated bugfixes that are included:
- [`7f7e956` (#1218)](https://github.com/JuliaLang/juliaup/pull/1218/commits/7f7e95663edbe35462dfc563facfbda0477a8732) - fix: add missing win32 handling in install_non_db_version
- [`c6eafe3` (#1218)](https://github.com/JuliaLang/juliaup/pull/1218/commits/c6eafe3d7fb76c2e1dfb2134fe094d7d5beab90e) - fix win32 sorting panic

If this PR isn't a simple merge it might be worth separating those out as a bugfix.

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -2,8 +2,7 @@ use anyhow::{anyhow, Context, Result};
 use console::{style, Term};
 use is_terminal::IsTerminal;
 use itertools::Itertools;
-use juliaup::command_add::run_command_add;
-use juliaup::config_file::{load_config_db, JuliaupConfig, JuliaupConfigChannel};
+use juliaup::config_file::{load_config_db, load_mut_config_db, save_config_db, JuliaupConfig, JuliaupConfigChannel};
 use juliaup::global_paths::get_paths;
 use juliaup::jsonstructs_versionsdb::JuliaupVersionDB;
 use juliaup::operations::{is_pr_channel, is_valid_channel};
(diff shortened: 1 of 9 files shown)


IanButterworth:
@davidanthoff I would prefer to be able to force push tidying these commits up into meaningful commits, but force push remains disallowed here. I think it should only be disabled for `main` and release branches.

IanButterworth:
@StefanKarpinski FYI. I think you mentioned this idea somewhere.

davidanthoff:
Cool, I really like this feature!

A few comments:
- I think we should have a little more opt-in for this, i.e. I don't think it is ideal that launching Julia might start a pretty large download. I think ideally what we want is this: If Julia is started in such a way that the REPL will eventually appear, then there should be an interactive prompt "Do you want to install channel X (y/n)", if Julia is started in non-interactive mode we should have a flag like `--force-juliaup`or something like that. I think we could also add a config setting where people can switch to the kind of fully automatic mode that is the default in this PR right now.
- So far the `julialauncher` binary does not contain any of the more complicated code that adds stuff etc. Instead it largely spawns `juliaup`with secret command line flags that do things. The reasoning was twofold: 1) I wanted to keep the binary size of `julialauncher`down (not sure how relevant that actually is), and 2) I have some ideas that would require `juliaup`to adopt `tokio`, and then we definitely don't want that in `julialauncher`, and I think mixing things would make that more difficult. So maybe there is a way that have the actual install step spawn `juliaup`and then wait until that finishes?

Implementation wise, it might be the best strategy to separate the interactive mode detection code from https://github.com/JuliaLang/juliaup/pull/1095 as a separate PR, merge that, and then my first point above could probably be implemented fairly quickly?

I hear you on the push restriction, but without it things happened in the past where I ended up literally spending hours trying to fix push/merge conflicts and I really don't want to deal with that :) The history here is a total mess in any case already, so there is no point in putting time into a nice history for a PR, it would be a diamond in a very rough sea ;)

IanButterworth:
UI updated under my mouse.. accidental close.

---

@davidanthoff sounds good. I've made some changes along those lines

When a channel isn't installed, users now get prompted:
```
Question: The Julia channel 'nightly' is not installed. Would you like to install it?
  [y]es (default), [Y]es and remember, [n]o
Choice: 
```
In non-interactive mode (no TTY), it defaults to "no" and shows the standard error message.

The config is 3-state:
- `juliaup config autoinstallchannels true` - Always auto-install (no prompts)
- `juliaup config autoinstallchannels false` - Never auto-install (no prompts) 
- `juliaup config autoinstallchannels default` - Reset to prompt behavior
- Unset (default) - Prompt in interactive mode, fail in non-interactive mode

And the launcher now spawns `juliaup add <channel>` as a subprocess rather than calling installation functions directly.

src/bin/julialauncher.rs
```diff
@@ -132,6 +134,119 @@ fn run_selfupdate(_config_file: &juliaup::config_file::JuliaupReadonlyConfigFile
     Ok(())
 }
 
+fn is_interactive() -> bool {
+    std::io::stdin().is_terminal() && std::io::stderr().is_terminal()
```
davidanthoff: I think this is a necessary condition, but I think in addition we need to parse the command line arguments and figure out whether the Julia process will jump into the REPL or execute either a script/code snippet. I think generally that if someone runs `julia +FOO main.jl`we should not show anything interactive, regardless of whether things are running in a terminal or not.

I think the PR I linked to above should have code that does that.

src/bin/julialauncher.rs
```diff
@@ -132,6 +134,119 @@ fn run_selfupdate(_config_file: &juliaup::config_file::JuliaupReadonlyConfigFile
...
+    // Prompt the user for auto-installation
+    eprint!(
+        "{} The Julia channel '{}' is not installed. Would you like to install it?\n  [y]es (default), [Y]es and remember my choice, [n]o\nChoice: ",
+        style("Question:").yellow().bold(),
+        channel
+    );
+    std::io::stderr().flush()?;
+
+    let mut input = String::new();
+    std::io::stdin().read_line(&mut input)?;
```
davidanthoff: We already use https://github.com/console-rs/dialoguer in the installer, we could use it here as well and have a consistent UI? But I'm not sure whether it actually has the kind of one-key-press option that I think would be ideal here.

I'm happy to merge as is if you want to keep it that way, just wanted to point to it :)

davidanthoff (changes requested):
Cool, that was fast :)

The only thing that I think we need to address is the interactive detection, otherwise looks great.

davidanthoff:
Oh, and I should say: Once we have that, we should do a similar thing for the message that a channel is out-of-date, it also should just show a prompt offering to update right away if we are in interactive mode!

src/bin/julialauncher.rs
IanButterworth: Oh that's nicer. I wasn't aware. Updated

src/bin/julialauncher.rs
IanButterworth: I couldn't find that check in the linked PR, but added this check based on julia args.

christiangnrd:
Can we fix the Win32 errors with #1222 instead of regressing to 1.10 being listed below 1.1 instead of 1.9?

The commit to revert is https://github.com/JuliaLang/juliaup/pull/1218/commits/1e5527bad816c8d9c53a49c6bdbb6dc3f481f1d0.

https://github.com/JuliaLang/juliaup/pull/1218/commits/c6eafe3d7fb76c2e1dfb2134fe094d7d5beab90e should also probably be reverted since does not fix the crashes. I don't know if what claude states about iterator chain sort stability is true, but there should probably be some discussion before merging this since personally I view having the direct-download channels all dumped at the bottom as a feature since it makes it easier to know which channels will probably take a while to download. Maybe those channels should be more explicitly separated (or have some sort of indication that they're direct-download).

IanButterworth:
Bump @davidanthoff 

Also, 

> Once we have that, we should do a similar thing for the message that a channel is out-of-date, it also should just show a prompt offering to update right away if we are in interactive mode!

I've been thinking about this and I don't think we should. I very often don't want to update because I just want to test what I have (and not hit more precompilation). I think adding a prompt in there might be more annoying than helpful.

KristofferC:
Yes, please no prompts.

IanButterworth:
@davidanthoff bump.

To try and help this along I'll see what copilot makes of it.

src/bin/juliaup.rs
```diff
@@ -3,6 +3,8 @@ use clap::Parser;
 use juliaup::cli::{ConfigSubCmd, Juliaup, OverrideSubCmd, SelfSubCmd};
 use juliaup::command_api::run_command_api;
 use juliaup::command_completions::generate_completion_for_command;
+use juliaup::command_completions::run_command_completions;
```
IanButterworth:
```suggestion
```

Keno (approved):
Seems reasonable to me. (Reviewed for functionality, not implementation, I'm not a rust programmer)

Issue #1204, opened 2025-05-03
Julia fails to launch without network connectivity


tecosaur:
I'm looking at running Julia in a lightly sandboxed environment, and it seems like Juliaup's propensity to check for new versions is making that a bit difficult.

```
~$ sudo unshare --net --fork --mount-proc $(which julia) -- --project=@ -E 'println(22)'
Checking for new Julia versions
Error: Failed to run `run_command_add` from the `run_command_initial_setup_from_launcher` command.

Caused by:
    0: Failed to update versions db.
    1: Failed to download current version db version.
    2: Failed to download from url `https://julialang-s3.julialang.org/juliaup/RELEASEPREVIEWCHANNELDBVERSION`.
    3: error sending request for url (https://julialang-s3.julialang.org/juliaup/RELEASEPREVIEWCHANNELDBVERSION)
    4: client error (Connect)
    5: dns error: failed to lookup address information: Try again
    6: failed to lookup address information: Try again
Error: The Julia launcher failed to figure out which juliaup channel to use.
```

*Linux, Juliaup 1.17.19*

tecosaur:
Ah, this looks like another version of #1184.

tecosaur:
I'm not sure whether these two things are worth keeping separate:
1. A way to skip the startup check
2. Juliaup failing to update shouldn't prevent Julia from running

davidanthoff:
I'm closing this as a duplicate of https://github.com/JuliaLang/juliaup/issues/1184.

Just to note, the error messages here are a bit misleading, I have an explanation of what is causing this in the linked issue.

tecosaur:
I've run into this again, and so had another look at this report as well as #1184. I've come to the conclusion that this is distinct from #1184 as the problem there is `/root/.julia` being checked when running with `sudo`, while the issue I've raised is juliaup failing to run in network-resticted environments.

nhz2:
I've made https://github.com/nhz2/install-julia-linux as an alternative installer that avoids this issue by installing symlinks to the julia binary instead of using a launcher.

IanButterworth:
Try the version on `releasepreview`.

nhz2:
v1.20.7 is still broken:

With install-julia-linux:

```sh
nathan@fedora:~$ sudo unshare --net --fork --mount-proc $(which julia) -e 'println(22)'
22
nathan@fedora:~$ install-julia.sh switch ~/packages/juliaup/julia
[...]

Issue #1184, opened 2025-03-18
Find a solution for a sudo based workflow


cstjean:
I'm working in a secured environment. `juliaup` worked fine, so does `julia`, but `sudo julia` fails with:

```
Checking for new Julia versions
Error: Failed to run `run_command_add` from the `run_command_initial_setup_from_launcher` command.

Caused by:
    0: Failed to update versions db.
    1: Failed to download current version db version.
    2: Failed to download from url `https://julialang-s3.julialang.org/juliaup/RELEASECHANNELDBVERSION`.
    3: error sending request for url (https://julialang-s3.julialang.org/juliaup/RELEASECHANNELDBVERSION)
    4: operation timed out
Error: The Julia launcher failed to figure out which juliaup channel to use.
```

Obviously, I can spend a week negociating with BigCo's IT department to fix the underlying issue, but I'd appreciate a shorter work-around. Is there any way to skip that check? Or to run the non-juliaup'ed `julia` binary directly?

Related: https://github.com/JuliaLang/juliaup/issues/705.

pddshk:
It also looks like `julia --version` also performs startup check which increases startup time. (Things like starship prompt timeout-fails when trying to determine julia version)

davidanthoff:
I think what is happening here is that when you run `sudo julia`, the launcher is run as the `root` user. It will then try to find the Juliaup config file in a folder in the profile of the `root` user instead of your user profile. Of course, nothing Juliaup specific is installed in the `root` profile, and in that case Juliaup tries to go through its initial setup procedure to install the `release` channel, create the Juliaup config file etc.

I am not entirely sure what the right course of action is here...

One option would be to check whether we are run via `sudo` and check something like `SUDO_USER`, and then try to locate the Juliaup config stuff in the profile for that user.

Another thinking might be that when one runs `sudo juliaup` or `sudo julia` one actually _wants_ to access a Juliaup depot that is in the root profile...

I am not familiar enough with Linux to say which of these is the right approach :)

Keno:
> I am not familiar enough with Linux to say which of these is the right approach :)

Should behave as root.

davidanthoff:
Ok, so (I think) that is what it is doing right now.

So essentially the story here is that if you install Juliaup as non root, and then run `sudo julia` you don't have a Juliaup install for root, and you need network access to get an initial install for root going.

cstjean:
[...]

PR #1229, opened 2025-08-21
Make `julia` more robust when `juliaup` has errors

IanButterworth:
Fixes https://github.com/JuliaLang/juliaup/issues/1204

Maybe there's more we could do?

Needs tests.

fyi @Keno 

Developed with Claude.


davidanthoff:
I've consolidated the problem description for this in https://github.com/JuliaLang/juliaup/issues/1184. The root cause is actually unrelated to version checks at startup, I describe the problem over in that issue. I'm closing this here, as I don't think it is the right fix for the problem.

PR #1234, opened 2025-08-26
Colored + wrapped help menu

MilesCranmer:
This adds standard colors to the help menus. It also enabled "wrap_text" on the binary so it can stay nicely formatted on narrow terminals.

Note that this default clap coloring automatically detects terminal compatibility before adding colors to the menu.

Before:

image

After:

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -252,6 +252,7 @@ dependencies = [
  "anstyle",
  "clap_lex",
  "strsim",
+ "terminal_size",
 ]
 
 [[package]]
@@ -2127,6 +2128,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "06794f8f6c5c898b3275aebefa6b8a1cb24cd2c6c79397ab15774837a0bc5755"
 dependencies = [
  "winapi-util",
+]
+
+[[package]]
+name = "terminal_size"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "60b8cb979cb11c32ce1603f8137b22262a9d131aaa5c37b5678025f22b8becd0"
+dependencies = [
+ "rustix",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ codegen-units = 1
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-clap = { version = "4.5", features = ["derive"] }
+clap = { version = "4.5", features = ["derive", "color", "wrap_help"] }
 clap_complete = "4.5"
 clap_complete_nushell = "4.5"
 dirs = "6.0.0"
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,10 +1,11 @@
+use crate::cli_styles;
 use clap::{Parser, ValueEnum};
 
 /// Shell options for completions
 #[derive(Clone, ValueEnum)]
 pub enum CompletionShell {
     Bash,
-    Elvish, 
+    Elvish,
     Fish,
     Nushell,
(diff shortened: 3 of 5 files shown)


MilesCranmer:
Might be nice to have this on Julia itself too...

StefanKarpinski:
> Might be nice to have this on Julia itself too...

That seems like it would require either duplicating the clap.rs logic in Julia or something kind of complex and clunky. Maybe Julia could have a `--help=json` option or something like that which dumps the help data as JSON, which a `juliaup` wrapper could then parse and reformat using clap.rs. When the `juliaup` launcher detects that help will be printed, it could skip launching the actual Julia process and just dump the help data. Another possibility since the Julia help data is relatively static, would be that we just ship it with all the other Julia version metadata.

MilesCranmer:
> That seems like it would require either duplicating the clap.rs logic in Julia or something kind of complex and clunky. Maybe Julia could have a `--help=json` option or something like that which dumps the help data as JSON, which a `juliaup` wrapper could then parse and reformat using clap.rs.

I think this could work. However, adding `--help=json` as a permanent option might not be worth it if we move to a better strategy later. For the moment I think it's easier to manually sync them (maybe a CI job could compare them) as this wouldn't expand the API surface.

Another alternative could be a julia-MMtk type approach (https://github.com/JuliaLang/julia/pull/56989) and make a new BinaryBuilder artifact for the help menu. The downside of this is that it would mean the help menu would live outside the source tree.

Probably the cleanest longterm option is clap.rs integration directly in the Julia source tree. I just re-read your answer [here](https://discourse.julialang.org/t/introducing-rust-alongside-c-in-julias-source-tree/124559/17?u=milescranmer) -

> Then there’s the basic OS runtime stuff. This is written in C and could more plausibly be implemented in Rust. However, a very large amount of this would have to be unsafe. As I said, it’s a very unusual program: it inherently needs to do a lot of unsafe (in the Rust sense) low level memory manipulation and does very little dynamic memory allocation that isn’t subsequently managed by Julia’s own GC. There’s a very small amount of concurrent data structure work, but it’s pretty minimal and unlikely to grow or change too much. Rust could maybe help there, but it seems better to keep the runtime as simple and lowest common denominator as possible, which favors C.

If the rust code were confined to the entrypoint – just clap.rs providing the help menu, shell completions, option parsing, and then passing the validated arguments into Julia – I think this might be a reasonable direction to explore?

MilesCranmer:
On second thought I don't think `--help=json` is possible, because clap.rs CLIs are compiled

StefanKarpinski:
Just replacing the CLI entry-point with Rust would be reasonable, but I think adding Rust to the Julia compilation toolchain may not be a popular move among core devs.

StefanKarpinski:
But, could be worth a discussion. If the Rust entry-point has enough to recommend it, then could be worthwhile.

MilesCranmer:
I can definitely vouch for clap.rs being worth it. It’s very well maintained and optimized, with 22,000 direct dependent packages in rust alone. In my view, its existence is a major reason behind the proliferation of rust-based shell tools.

Some key selling points:

- efficient validation of flags and arguments, with automatic fixes suggested (e.g., `--startupfile` generates a helpful error pointing the user to `--startup-file`)
- generates detailed shell completions for bash/zsh/fish/pwsh/elvish/nushell (this is at the level of being able to interactively tab through `-O 0`, `-O 1`, `-O 2`, `-O 3` in the julia command)
- generates roff man pages
- still gives tiny binaries despite lots of features  
- configurable and polished styling that automatically adapts to a wide variety of terminal settings

In general, clap.rs is also super portable and constantly stress-tested across OSes and terminals. 

Maybe we trial it in juliaup, and if folks like it (and core devs are open), there could be a discussion about making it the standard entry point?

MilesCranmer:
@IanButterworth do you think this PR could be merged? This one doesn't add anything new, it's just to enable the text wrapping and color in the existing clap.rs menu for juliaup itself

PR #1236, opened 2025-08-26
clap.rs colored help menus for Julia launcher

MilesCranmer:
Currently the Julia help menu within juliaup is missing some of the juliaup-specific stuff, such as the `[+channel]` specifier. It also (1) lacks shell completions, (2) lacks colored output (makes it hard to read, especially with how long it has gotten), and (3) it cannot automatically wrap to terminal width.

This PR adds a full clap.rs help menu for the julia launcher. This gives us:

1. Shell completion generation for Julia across multiple shells (`--generate-completions <shell>`)
2. Colored hierarchy
3. Automatic text wrapping

This also adds three flags:

1. `--help-raw` to access the raw Julia help menu (such as for core developers).
2. `--help-hidden-raw` equivalent but for `--help-hidden`.
3. `--generate-completions` to generate completions for Julia using clap.rs.

(This is forked from #1234; that one needs to be merged first in any case).

Here is the current julia launcher help menu on 1.12:

image

And here is it on a narrow terminal:

image

With this PR, it looks like this:

image

and in a narrow terminal:

image

### DIFF
--- a/src/julia_completions.rs
+++ b/src/julia_completions.rs
@@ -339,7 +339,7 @@ fn julia_cli_impl(only_hidden: bool) -> Command {
             .value_name("tracefile.info")
             .help("Append coverage information to the LCOV tracefile (filename supports format tokens)")
             .action(ArgAction::Set)
-            .conflicts_with_all(&["code-coverage", "code-coverage-path"]),
+            .conflicts_with_all(["code-coverage", "code-coverage-path"]),
         // Track allocation with path
         Arg::new("track-allocation-path")
             .long("track-allocation")
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -325,7 +325,7 @@ fn generate_julia_completions(shell: CompletionShell) -> Result<()> {
 }
 
 fn handle_help_and_completion_args(
-    args: &mut Vec<String>,
+    args: &mut [String],
     arg_index: usize,
 ) -> Result<Option<i32>> {
     // Check for help request - show our colored help


src/julia_completions.rs
```diff
@@ -0,0 +1,487 @@
...
+pub fn julia_cli() -> Command {
+    julia_cli_impl(false)
+}
+
+pub fn julia_cli_with_hidden() -> Command {
+    // Show ONLY the hidden options - filter to just hidden args
+    julia_cli_impl(true)
+}
+
+// NOTE: This was last generated against the Julia 1.12.0 help menu
```
fingolfin, line 13: What does that mean? Who or what "generated" it? And does this mean the list is specific to 1.12.0? So if 1.13 adds a new option, or removes one, it is not reflected?

fingolfin:
This looks pretty nice but has merge conflicts by now :-(

src/julia_completions.rs
IanButterworth: Can we just make it pass-through? Otherwise using `--help` via `+nightly` etc will be confusing, and i'm sure we'll forget to update this.

fingolfin: Yeah I am worried we'd be showing stale `--help` data.

One way to deal with that would be to add a new hidden argument to Julia, something like `--dump-commands` which dumps JSON (?) describing all options, and whether they should be shown on `--help` or `--help-hidden`, etc. Then we could use that here. (And for the finite number of older versions, either this feature is disabled, or else we just hard code the JSON needed for them in `juliaup`

fingolfin:
On concern with this PR is that it means that `julia --help` would look different depending on how Julia is installed. Now I am not saying this is a definite blocker, but it *is* a tradeoff to consider, and at least explicitly address and state "yes we know but we want it anyway" (count me in the "want it anyway" camp ;-) )

PR #1237, opened 2025-08-27
Expand `link` to handle aliases

IanButterworth:
Alternative to #1115 

See discussion around https://github.com/JuliaLang/juliaup/pull/1115#issuecomment-3228138215

Closes https://github.com/JuliaLang/juliaup/pull/838
Closes https://github.com/JuliaLang/juliaup/issues/830

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -21,7 +21,7 @@ fn main() -> Result<()> {
 
     let db_path = Path::new(&env::var("CARGO_MANIFEST_DIR").unwrap())
         .join("versiondb")
-        .join(format!("versiondb-{}.json", target_platform));
+        .join(format!("versiondb-{target_platform}.json"));
 
     let version_db_path = out_path.join("versionsdb.json");
     std::fs::copy(&db_path, &version_db_path).unwrap();
@@ -35,8 +35,7 @@ fn main() -> Result<()> {
     std::fs::write(
         &bundled_version_path,
         format!(
-            "pub const BUNDLED_JULIA_VERSION: &str = {}; pub const BUNDLED_DB_VERSION: &str = {};",
-            bundled_version_as_string, bundled_dbversion_as_string
+            "pub const BUNDLED_JULIA_VERSION: &str = {bundled_version_as_string}; pub const BUNDLED_DB_VERSION: &str = {bundled_dbversion_as_string};"
         ),
     )
     .unwrap();
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -1,4 +1,4 @@
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use console::{style, Term};
 use is_terminal::IsTerminal;
 use itertools::Itertools;
@@ -164,6 +164,23 @@ enum JuliaupChannelSource {
     Default,
 }
 
+fn resolve_channel_alias(
+    config_data: &JuliaupConfig,
+    channel: &str,
+    max_depth: usize,
+) -> Result<String> {
+    if max_depth == 0 {
+        bail!("Channel alias chain too deep for `{}`. There might be a circular reference.", channel);
+    }
+
+    match config_data.installed_channels.get(channel) {
+        Some(JuliaupConfigChannel::AliasChannel { target }) => {
+            resolve_channel_alias(config_data, target, max_depth - 1)
+        }
+        _ => Ok(channel.to_string()),
+    }
+}
+
 fn get_julia_path_from_channel(
     versions_db: &JuliaupVersionDB,
     config_data: &JuliaupConfig,
@@ -171,42 +188,49 @@ fn get_julia_path_from_channel(
     juliaupconfig_path: &Path,
     juliaup_channel_source: JuliaupChannelSource,
 ) -> Result<(PathBuf, Vec<String>)> {
-    let channel_valid = is_valid_channel(versions_db, &channel.to_string())?;
+    // First resolve any aliases
+    let resolved_channel = resolve_channel_alias(config_data, channel, 10)?;
+
+    let channel_valid = is_valid_channel(versions_db, &resolved_channel)?;
     let channel_info = config_data
             .installed_channels
-            .get(channel)
+            .get(&resolved_channel)
             .ok_or_else(|| match juliaup_channel_source {
                 JuliaupChannelSource::CmdLine => {
                     if channel_valid {
-                        UserError { msg: format!("`{}` is not installed. Please run `juliaup add {}` to install channel or version.", channel, channel) }
-                    } else if is_pr_channel(channel) {
-                        UserError { msg: format!("`{}` is not installed. Please run `juliaup add {}` to install pull request channel if available.", channel, channel) }
+                        UserError { msg: format!("`{}` is not installed. Please run `juliaup add {}` to install channel or version.", resolved_channel, resolved_channel) }
+                    } else if is_pr_channel(&resolved_channel) {
+                        UserError { msg: format!("`{}` is not installed. Please run `juliaup add {}` to install pull request channel if available.", resolved_channel, resolved_channel) }
                     } else {
-                        UserError { msg: format!("Invalid Juliaup channel `{}`. Please run `juliaup list` to get a list of valid channels and versions.",  channel) }
+                        UserError { msg: format!("Invalid Juliaup channel `{}`. Please run `juliaup list` to get a list of valid channels and versions.",  resolved_channel) }
                     }
                 },
                 JuliaupChannelSource::EnvVar=> {
                     if channel_valid {
-                        UserError { msg: format!("`{}` from environment variable JULIAUP_CHANNEL is not installed. Please run `juliaup add {}` to install channel or version.", channel, channel) }
-                    } else if is_pr_channel(channel) {
-                        UserError { msg: format!("`{}` from environment variable JULIAUP_CHANNEL is not installed. Please run `juliaup add {}` to install pull request channel if available.", channel, channel) }
+                        UserError { msg: format!("`{}` from environment variable JULIAUP_CHANNEL is not installed. Please run `juliaup add {}` to install channel or version.", resolved_channel, resolved_channel) }
+                    } else if is_pr_channel(&resolved_channel) {
+                        UserError { msg: format!("`{}` from environment variable JULIAUP_CHANNEL is not installed. Please run `juliaup add {}` to install pull request channel if available.", resolved_channel, resolved_channel) }
                     } else {
-                        UserError { msg: format!("Invalid Juliaup channel `{}` from environment variable JULIAUP_CHANNEL. Please run `juliaup list` to get a list of valid channels and versions.",  channel) }
+                        UserError { msg: format!("Invalid Juliaup channel `{}` from environment variable JULIAUP_CHANNEL. Please run `juliaup list` to get a list of valid channels and versions.",  resolved_channel) }
                     }
                 },
                 JuliaupChannelSource::Override=> {
                     if channel_valid {
-                        UserError { msg: format!("`{}` from directory override is not installed. Please run `juliaup add {}` to install channel or version.", channel, channel) }
-                    } else if is_pr_channel(channel) {
-                        UserError { msg: format!("`{}` from directory override is not installed. Please run `juliaup add {}` to install pull request channel if available.", channel, channel) }
+                        UserError { msg: format!("`{}` from directory override is not installed. Please run `juliaup add {}` to install channel or version.", resolved_channel, resolved_channel) }
+                    } else if is_pr_channel(&resolved_channel) {
+                        UserError { msg: format!("`{}` from directory override is not installed. Please run `juliaup add {}` to install pull request channel if available.", resolved_channel, resolved_channel) }
                     } else {
-                        UserError { msg: format!("Invalid Juliaup channel `{}` from directory override. Please run `juliaup list` to get a list of valid channels and versions.",  channel) }
+                        UserError { msg: format!("Invalid Juliaup channel `{}` from directory override. Please run `juliaup list` to get a list of valid channels and versions.",  resolved_channel) }
                     }
                 },
-                JuliaupChannelSource::Default => UserError {msg: format!("The Juliaup configuration is in an inconsistent state, the currently configured default channel `{}` is not installed.", channel) }
+                JuliaupChannelSource::Default => UserError {msg: format!("The Juliaup configuration is in an inconsistent state, the currently configured default channel `{}` is not installed.", resolved_channel) }
             })?;
 
     match channel_info {
+        JuliaupConfigChannel::AliasChannel { target: _ } => {
+            // This should not happen after alias resolution, but just in case
+            bail!("Unexpected alias channel after resolution: {}", resolved_channel);
+        }
         JuliaupConfigChannel::LinkedChannel { command, args } => {
             Ok((
                 PathBuf::from(command),
@@ -216,12 +240,12 @@ fn get_julia_path_from_channel(
         JuliaupConfigChannel::SystemChannel { version } => {
             let path = &config_data
                 .installed_versions.get(version)
-                .ok_or_else(|| anyhow!("The juliaup configuration is in an inconsistent state, the channel {} is pointing to Julia version {}, which is not installed.", channel, version))?.path;
+                .ok_or_else(|| anyhow!("The juliaup configuration is in an inconsistent state, the channel {} is pointing to Julia version {}, which is not installed.", resolved_channel, version))?.path;
 
-            check_channel_uptodate(channel, version, versions_db).with_context(|| {
+            check_channel_uptodate(&resolved_channel, version, versions_db).with_context(|| {
                 format!(
                     "The Julia launcher failed while checking whether the channel {} is up-to-date.",
-                    channel
+                    resolved_channel
                 )
             })?;
             let absolute_path = juliaupconfig_path
@@ -247,7 +271,7 @@ fn get_julia_path_from_channel(
             version: _,
         } => {
             if local_etag != server_etag {
-                if channel.starts_with("nightly") {
+                if resolved_channel.starts_with("nightly") {
                     // Nightly is updateable several times per day so this message will show
                     // more often than not unless folks update a couple of times a day.
                     // Also, folks using nightly are typically more experienced and need
@@ -258,12 +282,12 @@ fn get_julia_path_from_channel(
                 } else {
                     eprintln!(
                         "A new version of Julia for the `{}` channel is available. Run:",
-                        channel
+                        resolved_channel
                     );
                     eprintln!();
                     eprintln!("  juliaup update");
                     eprintln!();
-                    eprintln!("to install the latest Julia for the `{}` channel.", channel);
+                    eprintln!("to install the latest Julia for the `{}` channel.", resolved_channel);
                 }
             }
 
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -4,7 +4,7 @@ use clap::{Parser, ValueEnum};
 #[derive(Clone, ValueEnum)]
 pub enum CompletionShell {
     Bash,
-    Elvish, 
+    Elvish,
     Fish,
     Nushell,
     PowerShell,
@@ -23,10 +23,14 @@ pub enum Juliaup {
     Default { channel: String },
     /// Add a specific Julia version or channel to your system. Access via `julia +{channel}` e.g. `julia +1.6`
     Add { channel: String },
-    /// Link an existing Julia binary to a custom channel name
+    /// Link an existing Julia binary to a custom channel name, or create a channel alias
     Link {
+        /// Name of the new channel to create
         channel: String,
+        /// Path to Julia binary, or +CHANNEL to create an alias
+        #[clap(help = "Path to Julia binary, or +CHANNEL to create an alias (e.g. +release)", value_name = "TARGET")]
         file: String,
+        /// Additional arguments for the Julia binary (not used for aliases)
         args: Vec<String>,
     },
     /// List all available channels
@@ -62,9 +66,9 @@ pub enum Juliaup {
     #[clap(subcommand, name = "self")]
     SelfSubCmd(SelfSubCmd),
     /// Generate tab-completion scripts for your shell
-    Completions { 
+    Completions {
         #[arg(value_enum, value_name = "SHELL")]
-        shell: CompletionShell 
+        shell: CompletionShell
     },
     // This is used for the cron jobs that we create. By using this UUID for the command
     // We can identify the cron jobs that were created by juliaup for uninstall purposes
--- a/src/command_api.rs
+++ b/src/command_api.rs
@@ -43,17 +43,41 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
         "Failed to load configuration file while running the getconfig1 API command."
     })?;
 
-    for (key, value) in config_file.data.installed_channels {
-        let curr = match value {
-            JuliaupConfigChannel::SystemChannel {
-                version: fullversion,
-            } => {
-                let (platform, mut version) = parse_versionstring(&fullversion)
+    for (key, value) in &config_file.data.installed_channels {
+        let curr = match &value {
+            JuliaupConfigChannel::AliasChannel { target } => {
+                // For aliases, we need to resolve to the target and get its info
+                // Skip if the target doesn't exist to avoid infinite recursion
+                if let Some(target_channel) = config_file.data.installed_channels.get(target) {
+                    match target_channel {
+                        JuliaupConfigChannel::AliasChannel { .. } => {
+                            // Avoid infinite recursion for alias-to-alias
+                            continue;
+                        }
+                        _ => {
+                            // Recursively get info for the target
+                            // For now, just indicate it's an alias
+                            JuliaupChannelInfo {
+                                name: key.clone(),
+                                file: format!("alias-to-{}", target),
+                                args: Vec::new(),
+                                version: format!("alias to {}", target),
+                                arch: "".to_string(),
+                            }
+                        }
+                    }
+                } else {
+                    // Target doesn't exist, skip this alias
+                    continue;
+                }
+            }
+            JuliaupConfigChannel::SystemChannel { version: fullversion } => {
+                let (platform, mut version) = parse_versionstring(fullversion)
                     .with_context(|| "Encountered invalid version string in the configuration file while running the getconfig1 API command.")?;
 
                 version.build = semver::BuildMetadata::EMPTY;
 
-                match config_file.data.installed_versions.get(&fullversion) {
+                match config_file.data.installed_versions.get(fullversion) {
                     Some(channel) => JuliaupChannelInfo {
                         name: key.clone(),
                         file: paths.juliauphome
@@ -75,13 +99,13 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
             JuliaupConfigChannel::LinkedChannel { command, args } => {
                 let mut new_args: Vec<String> = Vec::new();
 
-                for i in args.as_ref().unwrap() {
+                for i in args.as_ref().unwrap_or(&Vec::new()) {
                     new_args.push(i.to_string());
                 }
 
                 new_args.push("--version".to_string());
 
-                let res = std::process::Command::new(&command)
+                let res = std::process::Command::new(command)
                     .args(&new_args)
                     .output();
 
@@ -101,7 +125,7 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
                         JuliaupChannelInfo {
                             name: key.clone(),
                             file: command.clone(),
-                            args: args.unwrap_or_default(),
+                            args: args.as_ref().unwrap_or(&Vec::new()).clone(),
                             version: version.to_string(),
                             arch: "".to_string(),
                         }
@@ -130,7 +154,7 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
 
         match config_file.data.default {
             Some(ref default_value) => {
-                if &key == default_value {
+                if key == default_value {
                     ret_value.default = Some(curr.clone());
                 } else {
                     ret_value.other_versions.push(curr);
@@ -146,7 +170,7 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
     let j = serde_json::to_string(&ret_value)?;
 
     // Print, write to a file, or send to an HTTP server.
-    println!("{}", j);
+    println!("{j}");
 
     Ok(())
 }
--- a/src/command_link.rs
+++ b/src/command_link.rs
@@ -27,24 +27,51 @@ pub fn run_command_link(
     }
 
     if is_valid_channel(&versiondb_data, &channel.to_string())? {
-        eprintln!("WARNING: The channel name `{}` is also a system channel. By linking your custom binary to this channel you are hiding this system channel.", channel);
+        eprintln!("WARNING: The channel name `{channel}` is also a system channel. By linking your custom binary to this channel you are hiding this system channel.");
     }
 
-    let absolute_file_path = Path::new(file)
-        .absolutize()
-        .with_context(|| format!("Failed to convert path `{}` to absolute path.", file))?;
+    // Check if this is a channel alias (starts with +)
+    if let Some(target_channel) = file.strip_prefix('+') {
+        // Remove the + prefix
 
-    if !is_valid_julia_path(&absolute_file_path.to_path_buf()) {
-        eprintln!("WARNING: There is no julia binary at {}. If this was a mistake, run `juliaup remove {}` and try again.", absolute_file_path.to_string_lossy(), channel);
+        // Validate that the target channel exists or is valid
+        if !config_file.data.installed_channels.contains_key(target_channel)
+            && !is_valid_channel(&versiondb_data, &target_channel.to_string())? {
+            bail!("Target channel `{}` is not installed and is not a valid system channel. Please run `juliaup add {}` first or check `juliaup list` for available channels.", target_channel, target_channel);
+        }
+
+        if !args.is_empty() {
+            bail!("Arguments are not supported when creating channel aliases. Remove the extra arguments: {:?}", args);
+        }
+
+        config_file.data.installed_channels.insert(
+            channel.to_string(),
+            JuliaupConfigChannel::AliasChannel {
+                target: target_channel.to_string(),
+            },
+        );
+
+        eprintln!("Channel alias `{channel}` created, pointing to `{target_channel}`.");
+    } else {
+        // Original behavior for linking to binary files
+        let absolute_file_path = Path::new(file)
+            .absolutize()
+            .with_context(|| format!("Failed to convert path `{file}` to absolute path."))?;
+
+        if !is_valid_julia_path(&absolute_file_path.to_path_buf()) {
+            eprintln!("WARNING: There is no julia binary at {}. If this was a mistake, run `juliaup remove {}` and try again.", absolute_file_path.to_string_lossy(), channel);
+        }
+
+        config_file.data.installed_channels.insert(
+            channel.to_string(),
+            JuliaupConfigChannel::LinkedChannel {
+                command: absolute_file_path.to_string_lossy().to_string(),
+                args: Some(args.to_vec()),
+            },
+        );
+
+        eprintln!("Channel `{}` linked to `{}`.", channel, absolute_file_path.to_string_lossy());
     }
-
-    config_file.data.installed_channels.insert(
-        channel.to_string(),
-        JuliaupConfigChannel::LinkedChannel {
-            command: absolute_file_path.to_string_lossy().to_string(),
-            args: Some(args.to_vec()),
-        },
-    );
 
     #[cfg(not(windows))]
     let create_symlinks = config_file.data.settings.create_channel_symlinks;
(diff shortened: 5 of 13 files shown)


IanButterworth:
@MilesCranmer would you mind reviewing this. Thanks

build.rs
```diff
@@ -21,7 +21,7 @@ fn main() -> Result<()> {
 
     let db_path = Path::new(&env::var("CARGO_MANIFEST_DIR").unwrap())
         .join("versiondb")
-        .join(format!("versiondb-{}.json", target_platform));
+        .join(format!("versiondb-{target_platform}.json"));
```
MilesCranmer: nit: would prefer these stylistic changes in a different PR

src/operations.rs
```diff
@@ -786,7 +787,10 @@ pub fn garbage_collect_versions(
     }
 
     if versions_to_uninstall.is_empty() {
-        eprintln!("Nothing to remove.");
+        eprintln!(
+            "{} No unused Julia installations to clean up.",
+            style("GC").cyan().bold()
+        );
```
MilesCranmer:
```suggestion
        eprintln!("{style("GC").cyan().bold()}: No unused Julia installations to clean up.");
```

src/operations.rs
```diff
@@ -858,6 +862,10 @@ pub fn create_symlink(
     let updating = _remove_symlink(&symlink_path)?;
 
     match channel {
```
MilesCranmer, line 984: This match is getting a bit unwieldy. Perhaps each match could be routed to a dedicated function, rather than having the code inlined

src/command_update.rs
```diff
@@ -75,6 +99,13 @@ fn update_channel(
                 );
             }
         }
+        JuliaupConfigChannel::AliasChannel { target: _ } => {
+            // This should never happen since we resolve aliases before calling this function
+            bail!(
+                "Internal error: Tried to update an alias channel '{}' directly.",
```
MilesCranmer: The style should be the same, if we are switching to the `{channel}` syntax?

src/command_update.rs
```diff
@@ -11,6 +11,30 @@ use anyhow::{anyhow, bail, Context, Result};
...
+    let mut visited = std::collections::HashSet::new();
+
+    loop {
+        if visited.contains(current_channel) {
+            bail!("Channel alias chain contains a cycle: {}", channel_name);
+        }
+        visited.insert(current_channel.to_string());
+
+        if visited.len() > 10 {
+            bail!("Channel alias chain too deep: {}", channel_name);
```
MilesCranmer: Same comment about format style

src/command_update.rs
```diff
@@ -11,6 +11,30 @@ use anyhow::{anyhow, bail, Context, Result};
 use console::style;
 use std::path::PathBuf;
 
+fn resolve_channel_alias(config_db: &JuliaupConfig, channel_name: &str) -> Result<String> {
+    let mut current_channel = channel_name;
+    let mut visited = std::collections::HashSet::new();
+
+    loop {
+        if visited.contains(current_channel) {
+            bail!("Channel alias chain contains a cycle: {}", channel_name);
```
MilesCranmer: Does the existing link code already handle this? I don't think detecting cycles is specific to aliases, because I could create a similar scenario with regular `link` on binaries (because each `link` creates a symlink file in the juliaup folder)

src/command_status.rs
```diff
@@ -104,6 +107,7 @@ pub fn run_command_status(paths: &GlobalPaths) -> Result<()> {
                         command: _,
                         args: _,
                     } => "".to_string(),
+                    JuliaupConfigChannel::AliasChannel { target: _ } => "".to_string(),
```
MilesCranmer: Probably more idiomatic to return `None` here. Then the `format!("Update to {} available", channel.version)` would need to get wrapped in an `Option`.

(It's similar to `Union{Nothing,T}`; in Rust it is `Option<T>(something)`)

tests/command_link.rs
MilesCranmer, line 1: Lots of repeat code in here. Probably the

```rust
    Command::cargo_bin("juliaup")
        .unwrap()
        .env("JULIA_DEPOT_PATH", depot_dir.path())
        .env("JULIAUP_DEPOT_PATH", depot_dir.path())
```

could be put into a function or some sort of interface

src/cli.rs
```diff
@@ -23,10 +23,17 @@ pub enum Juliaup {
...
     /// Add a specific Julia version or channel to your system. Access via `julia +{channel}` e.g. `julia +1.6`
     Add { channel: String },
-    /// Link an existing Julia binary to a custom channel name
+    /// Link an existing Julia binary to a custom channel name, or create a channel alias
     Link {
+        /// Name of the new channel to create
         channel: String,
+        /// Path to Julia binary, or +CHANNEL to create an alias
+        #[clap(
+            help = "Path to Julia binary, or +CHANNEL to create an alias (e.g. +release)",
```
MilesCranmer: Redundant with the `///` (clap reads from the docstring)

src/cli.rs
```diff
@@ -23,10 +23,17 @@ pub enum Juliaup {
...
+            value_name = "TARGET"
+        )]
         file: String,
```
MilesCranmer: Maybe call this "target" so it could stand for either a file or a channel

src/cli.rs
```diff
@@ -23,10 +23,17 @@ pub enum Juliaup {
     Default { channel: String },
     /// Add a specific Julia version or channel to your system. Access via `julia +{channel}` e.g. `julia +1.6`
     Add { channel: String },
-    /// Link an existing Julia binary to a custom channel name
+    /// Link an existing Julia binary to a custom channel name, or create a channel alias
```
MilesCranmer:
```suggestion
    /// Link an existing Julia binary or channel to a custom channel name
```

src/bin/julialauncher.rs
```diff
@@ -434,6 +460,10 @@ fn get_julia_path_from_installed_channel(
     channel_info: &JuliaupConfigChannel,
 ) -> Result<(PathBuf, Vec<String>)> {
     match channel_info {
+        JuliaupConfigChannel::AliasChannel { target: _ } => {
+            // This should not happen after alias resolution, but just in case
```
MilesCranmer:
```suggestion
```

(Typical Claude redundant comments :grin:)

src/bin/julialauncher.rs
```diff
@@ -325,6 +325,26 @@ enum JuliaupChannelSource {
     Default,
 }
 
+fn resolve_channel_alias(
```
MilesCranmer: I think aliases to aliases is overkill; better to force the user to refer to a concrete channel name.

src/command_link.rs
```diff
@@ -27,24 +27,59 @@ pub fn run_command_link(
...
-        },
-    );
+        if !is_valid_julia_path(&absolute_file_path.to_path_buf()) {
+            eprintln!("WARNING: There is no julia binary at {}. If this was a mistake, run `juliaup remove {}` and try again.", absolute_file_path.to_string_lossy(), channel);
+        }
+
+        config_file.data.installed_channels.insert(
+            channel.to_string(),
+            JuliaupConfigChannel::LinkedChannel {
+                command: absolute_file_path.to_string_lossy().to_string(),
```
MilesCranmer, line 74: Looks a bit sketchy, not sure what's going on here but seems to indicate a misuse somewhere

src/command_api.rs
```diff
@@ -75,13 +99,13 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
             JuliaupConfigChannel::LinkedChannel { command, args } => {
                 let mut new_args: Vec<String> = Vec::new();
 
-                for i in args.as_ref().unwrap() {
+                for i in args.as_ref().unwrap_or(&Vec::new()) {
```
MilesCranmer: Not sure what this is for

src/bin/julialauncher.rs
```diff
@@ -365,25 +388,28 @@ fn get_julia_path_from_channel(
...
 
                 // Reload the config to get the newly installed channel
                 let updated_config_file = load_config_db(paths, None)
                     .with_context(|| "Failed to reload configuration after installing channel.")?;
 
-                if let Some(channel_info) = updated_config_file.data.installed_channels.get(channel)
+                if let Some(channel_info) = updated_config_file
+                    .data
+                    .installed_channels
+                    .get(&resolved_channel)
```
MilesCranmer: A bit hard to read like this, I would say it deserves its own variable

src/command_api.rs
```diff
@@ -43,17 +43,41 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
...
-                let (platform, mut version) = parse_versionstring(&fullversion)
+    for (key, value) in &config_file.data.installed_channels {
+        let curr = match &value {
+            JuliaupConfigChannel::AliasChannel { target } => {
+                // For aliases, we need to resolve to the target and get its info
+                // Skip if the target doesn't exist to avoid infinite recursion
+                if let Some(target_channel) = config_file.data.installed_channels.get(target) {
+                    match target_channel {
+                        JuliaupConfigChannel::AliasChannel { .. } => {
+                            // Avoid infinite recursion for alias-to-alias
```
MilesCranmer: I vote we just ban aliases-of-aliases. Makes the code harder to maintain and less robust, and it saves a lazy user a few keystrokes at best (and that's just a one-time saving). Not to mention an alias to an alias is error prone in case the user forgets what it actually refers to.

src/command_api.rs
```diff
@@ -130,7 +154,7 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
 
         match config_file.data.default {
             Some(ref default_value) => {
-                if &key == default_value {
+                if key == default_value {
```
MilesCranmer, line 135: Seems unrelated to the PR

src/command_api.rs
```diff
@@ -101,7 +125,7 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
                         JuliaupChannelInfo {
                             name: key.clone(),
                             file: command.clone(),
-                            args: args.unwrap_or_default(),
+                            args: args.as_ref().unwrap_or(&Vec::new()).clone(),
```
MilesCranmer: Looks overcomplicated/sketchy

src/command_update.rs
```diff
@@ -143,7 +174,17 @@ pub fn run_command_update(channel: &Option<String>, paths: &GlobalPaths) -> Resu
                 );
             }
 
-            update_channel(&mut config_file.data, channel, &version_db, false, paths)?;
+            // Resolve any aliases to get the actual target channel
+            let resolved_channel = resolve_channel_alias(&config_file.data, channel)
+                .with_context(|| format!("Failed to resolve channel '{}'", channel))?;
```
MilesCranmer: I think `resolve_channel_alias` should be adding this error context, rather than `run_command_update`

src/command_api.rs
IanButterworth: Not sure why yet, but this change was needed
```
error[E0277]: can't compare `&std::string::String` with `std::string::String`
   --> src/command_api.rs:136:25
    |
136 |                 if &key == default_value {
    |                         ^^ no implementation for `&std::string::String == std::string::String`
    |
    = help: the trait `PartialEq<std::string::String>` is not implemented for `&std::string::String`
    = note: required for `&&std::string::String` to implement `PartialEq<&std::string::String>`
help: consider removing the borrow
    |
136 -                 if &key == default_value {
136 +                 if key == default_value {
    |

For more information about this error, try `rustc --explain E0277`.
error: could not compile `juliaup` (lib) due to 1 previous error
```

MilesCranmer: Weird. Ok fine by me

src/command_link.rs
IanButterworth: I didn't really understand this but got Claude to guess what you meant. See commit [`4790091` (#1237)](https://github.com/JuliaLang/juliaup/pull/1237/commits/47900912c2b1724fc0643e1e73e39774b126a710)

MilesCranmer: Thx. Is it ok if I push directly to your branch? I'm not sure I have access rights though.

IanButterworth: Open a PR against my branch?

MilesCranmer: ok

MilesCranmer: Actually my changes are pretty minor, ill just suggest them here

src/cli.rs
```diff
@@ -25,10 +25,14 @@ pub enum Juliaup {
...
     /// Add a specific Julia version or channel to your system. Access via `julia +{channel}` e.g. `julia +1.6`
     Add { channel: String },
-    /// Link an existing Julia binary to a custom channel name
+    /// Link an existing Julia binary or channel to a custom channel name
     Link {
+        /// Name of the new channel to create
         channel: String,
-        file: String,
+        /// Path to Julia binary, or +CHANNEL to create an alias (e.g. +release)
+        #[clap(value_name = "TARGET")]
```
MilesCranmer:
```suggestion
```

(Redundant)

src/command_api.rs
```diff
@@ -43,17 +43,25 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
...
-    for (key, value) in config_file.data.installed_channels {
-        let curr = match value {
-            JuliaupConfigChannel::SystemChannel {
-                version: fullversion,
-            } => {
-                let (platform, mut version) = parse_versionstring(&fullversion)
+    for (key, value) in &config_file.data.installed_channels {
+        let curr = match &value {
+            JuliaupConfigChannel::AliasChannel { target } => {
+                // Since we no longer support alias-to-alias chains, this is simpler
```
MilesCranmer:
```suggestion
```

(Seems like claude meta commentary?)

src/command_api.rs
```diff
@@ -73,15 +81,10 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
                 }
             }
             JuliaupConfigChannel::LinkedChannel { command, args } => {
-                let mut new_args: Vec<String> = Vec::new();
-
-                for i in args.as_ref().unwrap() {
-                    new_args.push(i.to_string());
-                }
-
+                let mut new_args = args.as_deref().unwrap_or_default().to_vec();
```
MilesCranmer:
```suggestion
                let mut new_args = args.clone().unwrap_or_default();
```

src/command_api.rs
```diff
@@ -101,9 +104,9 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
                         JuliaupChannelInfo {
                             name: key.clone(),
                             file: command.clone(),
-                            args: args.unwrap_or_default(),
+                            args: args.as_deref().unwrap_or_default().to_vec(),
```
MilesCranmer:
```suggestion
                            args: args.clone().unwrap_or_default(),
```

src/command_link.rs
```diff
@@ -53,13 +86,19 @@ pub fn run_command_link(
         .with_context(|| "`link` command failed to save configuration db.")?;
 
     #[cfg(not(windows))]
-    if create_symlinks {
+    if create_symlinks && !target.starts_with('+') {
+        // Only create symlinks for binary links, not channel aliases
+        // We need to recreate the absolute path since it goes out of scope
+        let absolute_file_path = Path::new(target)
+            .absolutize()
+            .with_context(|| format!("Failed to convert path `{target}` to absolute path."))?;
```
MilesCranmer:
```suggestion
        let absolute_file_path = std::fs::canonicalize(target)
            .with_context(|| format!("Failed to convert path `{target}` to absolute path."))?;
```

src/command_link.rs
```diff
@@ -27,24 +27,57 @@ pub fn run_command_link(
...
+                target: target_channel.to_string(),
+            },
+        );
+
+        eprintln!("Channel alias `{channel}` created, pointing to `{target_channel}`.");
+    } else {
+        // Original behavior for linking to binary files
+        let absolute_file_path = Path::new(target)
+            .absolutize()
+            .with_context(|| format!("Failed to convert path `{target}` to absolute path."))?;
```
MilesCranmer:
```suggestion
        let absolute_file_path = std::fs::canonicalize(target)
            .with_context(|| format!("Failed to convert path `{target}` to absolute path."))?;
```

src/command_link.rs
```diff
@@ -27,24 +27,57 @@ pub fn run_command_link(
...
-        channel.to_string(),
-        JuliaupConfigChannel::LinkedChannel {
-            command: absolute_file_path.to_string_lossy().to_string(),
-            args: Some(args.to_vec()),
-        },
-    );
+        config_file.data.installed_channels.insert(
+            channel.to_string(),
+            JuliaupConfigChannel::LinkedChannel {
+                command: absolute_file_path.to_string_lossy().into_owned(),
```
MilesCranmer:
```suggestion
                command: absolute_file_path.to_string_lossy().to_string(),
```

src/command_status.rs
```diff
@@ -84,39 +84,83 @@ pub fn run_command_status(paths: &GlobalPaths) -> Result<()> {
...
-                    } => {
-                        if local_etag != server_etag {
-                            "Update available".to_string()
-                        } else {
-                            "".to_string()
+                        JuliaupConfigChannel::LinkedChannel {
+                            command: _,
+                            args: _,
+                        } => None,
+                        JuliaupConfigChannel::AliasChannel { target } => {
```
MilesCranmer: Getting a bit complex; could this be moved into a function?

tests/command_link.rs
```diff
@@ -0,0 +1,356 @@
...
+
+    // Test that creating an alias with extra arguments fails (the argument parser should reject this)
+    Command::cargo_bin("juliaup")
+        .unwrap()
+        .arg("link")
+        .arg("alias_with_args")
+        .arg("+release")
+        .arg("--some-arg")
+        .env("JULIA_DEPOT_PATH", depot_dir.path())
+        .env("JULIAUP_DEPOT_PATH", depot_dir.path())
```
MilesCranmer: These (and the ones below) can also be refactored so its not writing out the env vars every time

tests/command_link.rs
```diff
@@ -0,0 +1,356 @@
+use assert_cmd::Command;
+use predicates::prelude::*;
```
MilesCranmer:
```suggestion
use assert_cmd::Command;
use assert_fs::TempDir;
use predicates::prelude::*;
```

MilesCranmer:
I think you could clean up the tests a lot with something like this:

```rust
struct TestEnv {
    depot_dir: assert_fs::TempDir,
}

impl TestEnv {
    fn new() -> Self {
        Self {
            depot_dir: assert_fs::TempDir::new().unwrap(),
        }
    }

    fn juliaup(&self) -> Command {
        self.command("juliaup")
    }

    fn julia(&self) -> Command {
        self.command("julia")
    }

    fn command(&self, bin: &str) -> Command {
        let mut cmd = Command::cargo_bin(bin).unwrap();
        cmd.env("JULIA_DEPOT_PATH", self.depot_dir.path());
        cmd.env("JULIAUP_DEPOT_PATH", self.depot_dir.path());
        cmd
    }
}
```

Then each test would look like:

```rust
  #[test]
  fn command_link_alias() {
      let env = TestEnv::new();

      env.juliaup()
          .arg("add")
          .arg("1.10.10")
          .assert()
          .success();

      env.juliaup()
          .arg("link")
          .arg("stable")
          .arg("+1.10.10")
          .assert()
          .success()
          .stderr(predicate::str::contains(
              "Channel alias `stable` created, pointing to `1.10.10`.",
          ));

      env.juliaup()
          .arg("status")
          .assert()
          .success()
          .stdout(
              predicate::str::contains("stable").and(predicate::str::contains("Alias to `1.10.10`")),
          );
  }
```

MilesCranmer:
Actually sorry @IanButterworth it looks like there is a difference between the stdlib `canonicalize` and `path_absolutize` which was in use before: https://docs.rs/path-absolutize/latest/path_absolutize/

> The difference between absolutize and canonicalize methods is that absolutize does not care about whether the file exists and what the file really is.


So perhaps it should be kept just in case it causes behaviour differences?

src/command_status.rs
```diff
@@ -13,6 +13,46 @@ use cli_table::{
 use itertools::Itertools;
 use numeric_sort::cmp;
 
+fn get_alias_update_info(
+    target: &str,
+    config_file: &crate::config_file::JuliaupReadonlyConfigFile,
```
MilesCranmer: Can put this at the top:

```rust
use crate::config_file::{JuliaupConfigChannel, JuliaupReadonlyConfigFile};
```

src/command_status.rs
```diff
@@ -13,6 +13,46 @@ use cli_table::{
...
+            JuliaupConfigChannel::SystemChannel { version } => {
+                match versiondb_data.available_channels.get(target) {
+                    Some(channel) => {
+                        if channel.version != *version {
+                            Some(format!("Update to {} available", channel.version))
+                        } else {
+                            None
+                        }
+                    }
+                    None => None,
```
MilesCranmer:
```suggestion
                    Some(channel) if channel.version != *version => {
                        Some(format!("Update to {} available", channel.version))
                    }
                    _ => None,
```

src/command_status.rs
```diff
@@ -13,6 +13,46 @@ use cli_table::{
...
+                url: _,
+                local_etag,
+                server_etag,
+                version: _,
+            } => {
+                if local_etag != server_etag {
+                    Some("Update available".to_string())
+                } else {
+                    None
+                }
```
MilesCranmer:
```suggestion
                (local_etag != server_etag).then(|| "Update available".to_string())
```

tests/command_update.rs
```diff
@@ -1,27 +1,58 @@
 use assert_cmd::Command;
 
+fn juliaup_command(depot_dir: &tempfile::TempDir) -> Command {
+    let mut cmd = Command::cargo_bin("juliaup").unwrap();
+    cmd.env("JULIA_DEPOT_PATH", depot_dir.path())
+        .env("JULIAUP_DEPOT_PATH", depot_dir.path());
+    cmd
+}
```
MilesCranmer: Can share the utilities from `command_link.rs` maybe. Is there some tests/utils.rs?

src/command_status.rs
```diff
@@ -84,39 +114,48 @@ pub fn run_command_status(paths: &GlobalPaths) -> Result<()> {
...
+                            url: _,
+                            local_etag,
+                            server_etag,
+                            version: _,
+                        } => {
+                            if local_etag != server_etag {
+                                Some("Update available".to_string())
+                            } else {
+                                None
+                            }
```
MilesCranmer:
```suggestion
                            (local_etag != server_etag).then(|| "Update available".to_string())
```

A matter of taste but slightly more idiomatic

IanButterworth: Can you check slack, there's some messages there :)

src/command_status.rs
```diff
@@ -84,39 +114,48 @@ pub fn run_command_status(paths: &GlobalPaths) -> Result<()> {
...
+                            match versiondb_data.available_channels.get(i.0) {
+                                Some(channel) => {
+                                    if &channel.version != version {
+                                        Some(format!("Update to {} available", channel.version))
+                                    } else {
+                                        None
+                                    }
                                 }
+                                None => None,
                             }
```
MilesCranmer:
```suggestion
                            match versiondb_data.available_channels.get(i.0) {
                                Some(channel) if &channel.version != version =>
                                    Some(format!("Update to {} available", channel.version)),
                                _ => None,
                            }
```

src/operations.rs
```diff
@@ -844,6 +848,144 @@ pub fn remove_symlink(symlink_name: &String) -> Result<()> {
     Ok(())
 }
 
+#[cfg(not(windows))]
+fn create_alias_symlink() -> Result<()> {
```
MilesCranmer: needed?

src/command_status.rs
```diff
@@ -13,6 +13,36 @@ use cli_table::{
 use itertools::Itertools;
 use numeric_sort::cmp;
 
+fn get_alias_update_info(
+    target: &str,
+    config_file: &JuliaupReadonlyConfigFile,
+    versiondb_data: &crate::jsonstructs_versionsdb::JuliaupVersionDB,
```
MilesCranmer: Similar comment here, can move the crate import to the top

src/command_status.rs
```diff
@@ -13,6 +13,36 @@ use cli_table::{
...
+                    _ => None,
+                }
+            }
+            JuliaupConfigChannel::DirectDownloadChannel {
+                path: _,
+                url: _,
+                local_etag,
+                server_etag,
+                version: _,
+            } => (local_etag != server_etag).then(|| "Update available".to_string()),
```
MilesCranmer:
```suggestion
            JuliaupConfigChannel::DirectDownloadChannel { local_etag, server_etag, .. } =>
                (local_etag != server_etag).then(|| "Update available".to_string()),
```

MilesCranmer: actually one sec, we could re-do the whole match into one...

src/command_status.rs
```diff
@@ -13,6 +13,36 @@ use cli_table::{
...
+            // LinkedChannels and nested aliases don't have updates
+            _ => None,
+        },
+        None => None, // Target channel doesn't exist
+    }
```
MilesCranmer: You could collapse this... But feel free to not, it's pretty stylistic change at that point. But the `DirectDownloadChannel` doesn't need all the parameters spelled out, note.

```suggestion
    match config_file.data.installed_channels.get(target) {
        Some(JuliaupConfigChannel::SystemChannel { version }) =>
            match versiondb_data.available_channels.get(target) {
                Some(channel) if channel.version != *version =>
                    Some(format!("Update to {} available", channel.version)),
                _ => None,
            },
        Some(JuliaupConfigChannel::DirectDownloadChannel { local_etag, server_etag, .. }) =>
            (local_etag != server_etag).then(|| "Update available".to_string()),
        _ => None,  // Target channel doesn't exist or not updatable
    }
```

src/command_status.rs
```diff
@@ -84,39 +114,38 @@ pub fn run_command_status(paths: &GlobalPaths) -> Result<()> {
...
+                            get_alias_update_info(target, &config_file, &versiondb_data)
                         }
-                    }
+                        JuliaupConfigChannel::DirectDownloadChannel {
+                            path: _,
+                            url: _,
+                            local_etag,
+                            server_etag,
+                            version: _,
+                        } => (local_etag != server_etag).then(|| "Update available".to_string()),
```
MilesCranmer:
```suggestion
                        JuliaupConfigChannel::DirectDownloadChannel { local_etag, server_etag, .. } =>
                            (local_etag != server_etag).then(|| "Update available".to_string()),
```

src/command_status.rs
```diff
@@ -84,39 +114,38 @@ pub fn run_command_status(paths: &GlobalPaths) -> Result<()> {
...
-                        version: _,
-                    } => {
-                        if local_etag != server_etag {
-                            "Update available".to_string()
-                        } else {
-                            "".to_string()
+                        JuliaupConfigChannel::LinkedChannel {
+                            command: _,
+                            args: _,
+                        } => None,
```
MilesCranmer:
```suggestion
                        JuliaupConfigChannel::LinkedChannel { .. } => None,
```

IanButterworth:
@MilesCranmer Please let me know when you've concluded your review. Acting on these comments individually is suboptimal. Also I've sent you some slack messages, if you can check there.

MilesCranmer:
I think you can "Add suggestion to batch"? 

And yeah sorry I should really be sending them all at once.

Done now.

src/cli.rs
```diff
@@ -25,10 +25,13 @@ pub enum Juliaup {
     Default { channel: String },
     /// Add a specific Julia version or channel to your system. Access via `julia +{channel}` e.g. `julia +1.6`
     Add { channel: String },
-    /// Link an existing Julia binary to a custom channel name
+    /// Link an existing Julia binary or channel to a custom channel name
     Link {
+        /// Name of the new channel to create
         channel: String,
-        file: String,
+        /// Path to Julia binary, or +CHANNEL to create an alias (e.g. +release)
```
MilesCranmer:
```suggestion
        /// Path to Julia binary, or +{channel} to create an alias
```

src/cli.rs
```diff
@@ -25,10 +25,13 @@ pub enum Juliaup {
...
+        target: String,
+        /// Additional arguments for the Julia binary (not used for aliases)
```
MilesCranmer:
```suggestion
        /// Additional arguments for the Julia binary
```
I think a helpful error messages is better than noting this in the docstring, wdyt?

MilesCranmer: Also. Why can't we pass additional arguments for aliases? Is this just not yet implemented? Couldn't we just do `args: Some(args.to_vec())`? I don't think there's anything that wouldn't be compatible, right?

src/command_link.rs
```diff
@@ -27,24 +27,56 @@ pub fn run_command_link(
...
+        config_file.data.installed_channels.insert(
+            channel.to_string(),
+            JuliaupConfigChannel::LinkedChannel {
+                command: absolute_file_path.to_string_lossy().to_string(),
+                args: Some(args.to_vec()),
+            },
+        );
+
+        eprintln!(
+            "Channel `{}` linked to `{}`.",
```
MilesCranmer, line 80: These keep switching back and forth in convention. Maybe we just do a standardized convention later?

src/command_update.rs
```diff
@@ -11,6 +11,14 @@ use anyhow::{anyhow, bail, Context, Result};
 use console::style;
 use std::path::PathBuf;
 
+fn resolve_channel_alias(config_db: &JuliaupConfig, channel_name: &str) -> Result<String> {
+    match config_db.installed_channels.get(channel_name) {
+        Some(JuliaupConfigChannel::AliasChannel { target }) => Ok(target.to_string()),
+        Some(_) => Ok(channel_name.to_string()),
+        None => bail!("Channel '{}' not found", channel_name),
```
MilesCranmer, line 18: (Convention - which one?)

tests/command_link.rs
```diff
@@ -0,0 +1,272 @@
...
+    // Create an alias to the installed version
+    env.juliaup()
+        .arg("link")
+        .arg("r")
+        .arg("+1.10.10")
+        .assert()
+        .success();
+
+    // Update through the alias - should work and update the target
+    env.juliaup().arg("update").arg("r").assert().success();
```
MilesCranmer: Anything to check in the stdout here? Just to be a bit stricter

src/command_update.rs
```diff
@@ -75,6 +83,9 @@ fn update_channel(
                 );
             }
         }
+        JuliaupConfigChannel::AliasChannel { .. } => {
+            bail!("Internal error: Tried to update an alias channel '{channel}' directly.");
+        }
```
MilesCranmer: Not sure I understand this error. Why can't we update through the alias? Maybe I don't follow but I thought there was even a test for this

MilesCranmer (commented):
Looking good - here are some nearly-final comments

src/command_link.rs
IanButterworth: Yeah. I noticed my local clippy which seems stricter fixes all these. I suggest we apply my local clippy in a PR following this and figure out why the CI one is more relaxed.

src/command_update.rs
IanButterworth: Above

src/cli.rs
IanButterworth: Good point. Added support for it

src/command_link.rs
IanButterworth: Also I think we should add an AGENTS.md to at least inform what tests, formatting and linting should be run before commits.

MilesCranmer: Maybe a .pre-commit-config.yaml file? I don't know AGENTS.md

src/command_api.rs
```diff
@@ -43,17 +43,24 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
...
 
-    for (key, value) in config_file.data.installed_channels {
-        let curr = match value {
-            JuliaupConfigChannel::SystemChannel {
-                version: fullversion,
-            } => {
-                let (platform, mut version) = parse_versionstring(&fullversion)
+    for (key, value) in &config_file.data.installed_channels {
+        let curr = match &value {
+            JuliaupConfigChannel::AliasChannel { target, args: _ } => {
```
MilesCranmer: Shouldn't these args be propagated forward?

I guess we need a test for this if this wasn't caught by CI

src/command_update.rs
```diff
@@ -75,6 +83,9 @@ fn update_channel(
                 );
             }
         }
+        JuliaupConfigChannel::AliasChannel { .. } => {
+            unreachable!("Alias channels should be resolved before calling update_channel. This is a programming error.");
```
MilesCranmer:
```suggestion
            unreachable!("Alias channels should be resolved before calling update_channel. Please submit a bug report.");
```

src/command_status.rs
```diff
@@ -84,39 +109,36 @@ pub fn run_command_status(paths: &GlobalPaths) -> Result<()> {
...
-                        local_etag,
-                        server_etag,
-                        version: _,
-                    } => {
-                        if local_etag != server_etag {
-                            "Update available".to_string()
-                        } else {
-                            "".to_string()
+                        JuliaupConfigChannel::LinkedChannel { .. } => None,
+                        JuliaupConfigChannel::AliasChannel { target, args: _ } => {
```
MilesCranmer:
```suggestion
                        JuliaupConfigChannel::AliasChannel { target, .. } => {
```

tests/command_link.rs
MilesCranmer, line 1: Need test of alias with args

src/operations.rs
```diff
@@ -761,6 +761,7 @@ pub fn garbage_collect_versions(
                 command: _,
                 args: _,
             } => true,
+            JuliaupConfigChannel::AliasChannel { target: _, args: _ } => true,
```
MilesCranmer:
```suggestion
            JuliaupConfigChannel::AliasChannel { .. } => true,
```

src/operations.rs
```diff
@@ -854,122 +990,34 @@ pub fn create_symlink(
         .with_context(|| "Failed to retrieve binary directory while trying to create a symlink.")?;
 
     let symlink_path = symlink_folder.join(symlink_name);
-
     let updating = _remove_symlink(&symlink_path)?;
 
     match channel {
+        JuliaupConfigChannel::AliasChannel { target: _, args: _ } => {
```
MilesCranmer:
```suggestion
        JuliaupConfigChannel::AliasChannel { .. } => {
```

src/command_link.rs
```diff
@@ -27,24 +27,61 @@ pub fn run_command_link(
...
+        }
 
-    if !is_valid_julia_path(&absolute_file_path.to_path_buf()) {
-        eprintln!("WARNING: There is no julia binary at {}. If this was a mistake, run `juliaup remove {}` and try again.", absolute_file_path.to_string_lossy(), channel);
-    }
+        config_file.data.installed_channels.insert(
+            channel.to_string(),
+            JuliaupConfigChannel::AliasChannel {
+                target: target_channel.to_string(),
+                args: if args.is_empty() {
```
MilesCranmer, line 49: This logic is different than LinkedChannel: Some(args.to_vec()) is used regardless of being empty

MilesCranmer: Basically I think both are fine, but I would just pick one convention

src/config_file.rs
```diff
@@ -51,6 +51,12 @@ pub enum JuliaupConfigChannel {
         #[serde(rename = "Args")]
         args: Option<Vec<String>>,
     },
+    AliasChannel {
+        #[serde(rename = "Target")]
+        target: String,
+        #[serde(rename = "Args", skip_serializing_if = "Option::is_none")]
```
MilesCranmer: I don't know what this skip_serializing_if does. If the LLM wrote it, maybe double check it makes sense? And test it. Note that LinkedChannel does not use this despite having the same Args so I'm not sure it's needed (see other comment about different logic between the two. I think they should have identical logic though).

src/command_remove.rs
```diff
@@ -39,40 +39,49 @@ pub fn run_command_remove(channel: &str, paths: &GlobalPaths) -> Result<()> {
         );
     }
 
-    let x = config_file.data.installed_channels.get(channel).unwrap();
+    let channel_info = config_file.data.installed_channels.get(channel).unwrap();
+
+    // Determine what type of channel is being removed for better messaging
+    let channel_type = match channel_info {
+        JuliaupConfigChannel::AliasChannel { target, args: _ } => {
```
MilesCranmer:
```suggestion
        JuliaupConfigChannel::AliasChannel { target, .. } => {
```
Missed this one. But same comment whenever this appears basically. We don't need to write out the arg if it's not used

MilesCranmer:
Resolved my comments as I put them into #1242.

By the way, will this mess up any of the existing serialized files at all? I'm assuming no but I really don't have a clear enough understanding for the internals to say.

MilesCranmer:
From a design perspective I'm a little worried by how LinkedChannel and AliasChannel seem to be so different. e.g., LinkedChannel looks to be creating shell scripts, whereas AliasChannel is using runtime resolution. I don't see a strong reason for this split. I think they should use a similar mechanism.

MilesCranmer:
To be honest I think the alias solution is better than the existing LinkedChannel solution. The LinkedChannel solution seems to rely on generating shell scripts (?) which seems kinda fragile.

IanButterworth:
@MilesCranmer 

> will this mess up any of the existing serialized files at all?

Maybe if people downgrade after using this new feature? I don't think we need to worry about that, we can try it out in the prerelease.

I think your other comment could be a follow on consideration?

Are you happy to approve this so we can move ahead?

MilesCranmer:
I think it's probably good to go. I haven't messed around with it locally though which is something I'd like to try. Have you given it some testing in your terminal at all? I think we should try to break things a little

IanButterworth:
We had a bug and a test gap for a general `juliaup update` when an alias is present. Fixed. I think we're good to go.

MilesCranmer:
@IanButterworth FYI I think both of my PRs #1241 and #1242 got squashed into a single commit without me included as coauthor

IanButterworth:
Odd. Squash merges usually retain authorship.

MilesCranmer:
I think if you do a `<ctrl-A> + <del>` inside the extended commit message it might have deleted the `Co-authored-by:` trailer

IanButterworth:
I didn't edit it. Maybe the GitHub default changed it.

MilesCranmer:
Did you use some CLI to merge it maybe? The website UI still has the same defaults as before:
image

IanButterworth:
Nope the website. I have an enhanced GitHub chrome extension, it could be that. I'll look when at my computer.

PR #1243, opened 2025-09-08
fix: incorrect data for symlink

MilesCranmer:
While messing around in #1242 I noticed this. I'm pretty sure `create_symlink` should be getting the updated `channel_data`, rather than simply propagating the previous info. I think this means that a symlink would have its version info forever frozen.

### DIFF
--- a/src/command_update.rs
+++ b/src/command_update.rs
@@ -97,21 +97,11 @@ fn update_channel(
 
                 config_db
                     .installed_channels
-                    .insert(channel.clone(), channel_data);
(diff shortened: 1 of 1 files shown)


IanButterworth (commented):
If this is wrong it would be good to have a test that shows it was, and is now fixed.

MilesCranmer:
Sadly won't have time to figure out testing for this myself (seems really tricky to test this), but I'll leave this open in case someone else wants to try in a fork of the branch

PR #1247, opened 2025-09-11
README: Give explicit example of the `--path` argument for installing Juliaup into a custom location

DilumAluthge:
IMO, the README currently doesn't make it super clear how to accomplish this task (installing Juliaup into a custom location).

People ask for this specific task every now and then, so I think that it's worth having a clear example.


DilumAluthge:
Just as one example, we had a user ask for this recently in Slack: https://julialang.slack.com/archives/C046NU37GSX/p1753277312566059

There have been other examples, but I think they've been lost to posterity.

Issue #1256, opened 2025-09-27
Suppressing the Juliaup outdated Julia version message


asinghvi17:
When running Julia in applications that look for stderr output as indications of errors, this Juliaup output can become highly annoying.

https://github.com/JuliaLang/juliaup/blob/5cc3c1be5fc9d09f06a38dac0f4e610903f21408/src/bin/julialauncher.rs#L482-L508

Would it be possible to control that via ENV variable?

IanButterworth:
Maybe it shouldn't be shown at all when Julia isn't being run interactively (we have a function for determining that now)

KristofferC:
That would be great.

PR #1258, opened 2025-09-29
Don't show `julia` update messages when not interactive

IanButterworth:
Fixes https://github.com/JuliaLang/juliaup/issues/1256

@KristofferC


IanButterworth:
Clippy failures addressed in #1259

IanButterworth:
Tests need work. It seems we don't actually test that upgrading works?

IanButterworth:
I think the src changes are simple enough here. Review still welcome but I'm going to proceed.

PR #1259, opened 2025-09-29
Pin clippy and rustfmt to toolchain 1.89.0

IanButterworth:
Just pin clippy and rustfmt to toolchain 1.89.0 so it doesn't change requirements from under our feet.

~~The simplest fix is to add `#[allow(dead_code)]` to these structs since they are constructed by serde during deserialization, not directly in the code.~~

### DIFF
--- a/src/jsonstructs_versionsdb.rs
+++ b/src/jsonstructs_versionsdb.rs
@@ -2,21 +2,18 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
 #[derive(Serialize, Deserialize)]
-#[allow(dead_code)]
 pub struct JuliaupVersionDBVersion {
     #[serde(rename = "UrlPath")]
     pub url_path: String,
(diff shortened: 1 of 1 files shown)


MilesCranmer (commented):
> since they are constructed by serde during deserialization, not directly in the code.

I'm not sure the relevance of this. If serde constructs these, then do they get used somewhere afterwards? If they aren't used anywhere then it's a valid warning, right?

IanButterworth:
Maybe it's a clippy setup issue. Can you investigate? 

Also, I don't think clippy should be allowed to update itself and start to fail like this. We should clamp it to a version IMO.

MilesCranmer:
I can't do much this week, sorry

IanButterworth:
It seems to be additional checks that were added in toolchain 1.90.0 so I pinned to 1.89.0 which previously passed, and reverted the prior change.

PR #1263, opened 2025-10-07
Move check for `is_interactive` earlier

IanButterworth:
To avoid a little more work if it's not going to be used.


IanButterworth:
Minor change (and reviews are hard to come by around here) so merging.

PR #1276, opened 2025-10-17
[WIP] Enhance release.yml for aarch64 Windows targets. DO NOT MERGE.

ViralBShah:
Added support for aarch64 Windows targets in the release workflow.


davidanthoff:
Just as a FYI, this would require a lot more work, as we need to integrate all of this into the MSIX installer story.

PR #1282, opened 2025-10-20
Update windows progress immediately

### LINKED ISSUE #1281: On windows the progress bar is delayed?
Image

See the lower left corner traffic monitor. The progress bar seems displayed after the download is finished. On linux it seems fine.

IanButterworth:
Fixes #1281 

Using Claude:

On Windows, the download process uses the Windows Runtime DataReader API through a custom `DataReaderWrap` struct. The issue was:

1. `LoadAsync()` downloads data from the network into a Windows buffer
2. `ReadBytes()` copies data from the Windows buffer to the Rust buffer
3. The progress bar only updated after the `read(`) method returned

This meant the actual network download (step 1) completed before the progress bar updated (after step 3), causing the visible lag.

### DIFF
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -134,19 +134,26 @@ pub fn download_extract_sans_parent(
 }
 
 #[cfg(windows)]
-struct DataReaderWrap(windows::Storage::Streams::DataReader);
+struct DataReaderWrap {
+    reader: windows::Storage::Streams::DataReader,
+    pb: ProgressBar,
+}
 
 #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


IanButterworth:
@xgdgsc as I'm not on windows, can you try building this and see if it fixes the issue? https://github.com/JuliaLang/juliaup/wiki/Developing-Juliaup#to-build-juliaup-locally

xgdgsc:
Built. Same behavior.

IanButterworth:
Actual fix https://github.com/JuliaLang/juliaup/pull/1592

Issue #1076, opened 2024-10-17
"Checking standard library notarization" should be a spinner


adienes:
this is very minor, and maybe not a universally shared opinion, but rather than the
```Checking standard library notarization............................................................done.```

where it appears the number of `.....` is unbounded with time, I'd prefer that just be a `%` spinner or something

KristofferC:
I think every dot represent one stdlib that is getting "checked".

Jollywatt:
Just out of curiosity, what does "Checking standard library notarization" mean?:)

IanButterworth:
Because juliaup serves julia via tar.gz rather than .dmg, it doesn't have the macOS gatekeeper certificates "stapled" to it, so macOS has to ask the internet for them the first time it loads every .dylib (all the stdlibs).

Options are either
- juliaup serves .dmgs on macOS #1010 
- we load every stdlib during install to get those checks done during install, so first proper julia usage isn't slow. i.e. very slow Pkg repl mode load


Also, each `.` is a stdlib, it's not unbounded.
It could be replaced with a `n/N complete` or a progress bar.. this just avoided overprinting.

PR #1285, opened 2025-10-20
show an increasing count for notarizing stdlibs instead of just dots

KristofferC:
fixes https://github.com/JuliaLang/juliaup/issues/1076

https://github.com/user-attachments/assets/bfcbac0d-0eac-4322-a8e6-a4e0015c0fc9

developed with Claude 🤖

### DIFF
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -410,23 +410,32 @@ pub fn install_version(
                 )
             })?;
 
-        eprint!("Checking standard library notarization");
-        let _ = std::io::stdout().flush();
-
         let exit_status = std::process::Command::new(julia_path)
             .env("JULIA_LOAD_PATH", "@stdlib")
             .arg("--startup-file=no")
             .arg("-e")
-            .arg("foreach(p -> begin print(stderr, '.'); @eval(import $(Symbol(p))) end, filter!(x -> isfile(joinpath(Sys.STDLIB, x, \"src\", \"$(x).jl\")), readdir(Sys.STDLIB)))")
+            .arg(
+                r#"
+                let
+                    pkgs = filter!(x -> isfile(joinpath(Sys.STDLIB, x, "src", "$(x).jl")), readdir(Sys.STDLIB))
+                    total = length(pkgs)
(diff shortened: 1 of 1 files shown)


KristofferC:
I'm wondering if `print(stderr, "\r\x1b[KChecking standard library notarization [$i/$total]")` would become spammy if it is run during CI or soemthign

IanButterworth:
Yeah. Also I can't figure out why tests pass, because they check for `.and(predicate::str::ends_with("done.\n"))`

KristofferC:
They don't pass (on mac at least) :P 

```
Unexpected stderr, failed (var.starts_with("Checking for new Julia versions\nInstalling Julia 1.") && ((! var.contains(apple.darwin14)) || (var.contains(Checking standard library notarization) && var.ends_with("done.\n"))))
```

The log also looks quite spammy. Seems to be some julia precompilation interleaved with the output?

IanButterworth:
Oh right.. pre-coffee. I forgot this is mac-only on CI, and macOS sometimes fails for other reasons.

IanButterworth:
Seems we don't run CI tests on aarch64 macOS.. just `check`. That isn't good.

IanButterworth:
> Seems we don't run CI tests on aarch64 macOS

Now we do. Needs an update from main

KristofferC:
Can't rebase because

> Repository rule violations found Cannot force-push to this branch

-.-

KristofferC:
I removed that rule.

IanButterworth:
We can side step this if we can get this in https://github.com/JuliaLang/juliaup/pull/1320

KristofferC:
That seems preferable indeed

davidanthoff:
> I'm wondering if `print(stderr, "\r\x1b[KChecking standard library notarization [$i/$total]")` would become spammy if it is run during CI or soemthign

We already use the `is_terminal` crate in other parts of the code to detect whether things are running in a terminal or not, so that could be used here to switch between different output styles.

KristofferC:
Let's just use dmgs :)

PR #1295, opened 2025-10-27
Make `save_config_db` atomic

IanButterworth:
Fixes https://github.com/JuliaLang/juliaup/issues/1221

I believe this is part of https://github.com/JuliaLang/juliaup/pull/1238 which has stalled, so this just tackles not corrupting the config file.

### DIFF
--- a/src/config_file.rs
+++ b/src/config_file.rs
@@ -5,6 +5,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::fs::{File, OpenOptions};
 use std::io::{BufReader, ErrorKind, Seek, SeekFrom, Write};
+#[cfg(target_os = "windows")]
 use std::mem;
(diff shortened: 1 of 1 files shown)


IanButterworth:
Getting reviews here is pretty hard, but I think this is important to do, so I'm going to go ahead. Reviewers, please post merge review.

davidanthoff:
This is great to see!

I think this all looks good, I'm just wondering whether we actually need to keep the config files open at all, given that we have the separate lock file? I think we might be able to move to a world where on read of the config file we just 1) open file, 2) read data, 3) close file handle right away. That would make the whole write story easier on Windows, right?

I _think_ (but don't have a good memory of it) that I kept the files open originally before I added the lock file story... Maybe... Probably worth thinking through a bit more, but it could potentially simplify things, right?

PR #1313, opened 2025-10-30
rustfmt CI fixes

IanButterworth:
@tecosaur 

The other issue is https://github.com/JuliaLang/juliaup/actions/runs/17244028982/job/48928853337?pr=1232#step:6:263

```
reviewdog: This GitHub Token doesn't have write permission of Review API [1],
so reviewdog will report results via logging command [2] and create annotations similar to
github-pr-check reporter as a fallback.
```

So it was incorrectly green and reviewdog comments didn't get posted.

@DilumAluthge can you help resolve the token issue (I don't have settings access)

### DIFF
--- a/.github/workflows/rustfmt.yml
+++ b/.github/workflows/rustfmt.yml
@@ -36,4 +36,4 @@ jobs:
         uses: reviewdog/action-suggester@v1
         with:
           tool_name: rustfmt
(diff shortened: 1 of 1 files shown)


IanButterworth:
Seems reviewdog can't post comments on PRs from forks, so this PR is the most we can do.

Issue #1111, opened 2024-11-17
Default project


davidanthoff:
Here is an idea that takes the ideas in https://github.com/JuliaLang/juliaup/pull/1095 to the next step.

We could have a default project at `.julia/environments/default` and `julialauncher` would start the actual `julia` process with that project active if the user didn't provide an explicit project via the command line or an environment variable.

This would have a couple of interesting implications:
- The "default" channel setting of Juliaup would be more or less unnecessary. The "default" Julia version on a user system would just be the one recorded in the "default" project manifest.
- It would be very rare to run Julia without an active project, it would essentially be either the default project or a user specified project is active. I think this would have major benefit in that users couldn't so easily add packages to the vX.Y projects that are always loaded because of stacked environments (essentially my point https://discourse.julialang.org/t/speculations-about-the-default-environment-or-a-new-draft-environment/78967/28?u=davidanthoff).
- The whole logic on how `julialauncher` informs users about new Julia versions would presumably be a bit different. I think it would be 1) check the `Project.toml` for a `compat` entry for `julia`, 2) look at the `Manifest.toml` and the database of existing Julia versions to decide whether there is a newer version that the user should be alerted to. This would of course also be useful just for #1095.
- For this to happen, we would need _much_ better tools to update manifests across Julia builds.

PR #1315, opened 2025-10-30
match against environment manifest version before using the default channel

KristofferC:
alternative to https://github.com/JuliaLang/juliaup/pull/1095

I am not much of a rust programmer so I kind of wrote pseudo code and had Claude Code translate it for me. Should still do some manual cleanups.

In use:

```
~/JuliaPkgs/TimerOutputs.jl
❯ julia --project -q
Info: Installing Julia 1.10.10 automatically per juliaup settings...
Info: Successfully installed Julia 1.10.10.
julia> VERSION
v"1.10.10"

# Change the manifest file

~/JuliaPkgs/TimerOutputs.jl
❯ julia --project -q
Info: Installing Julia 1.10.8 automatically per juliaup settings...
Installing Julia 1.10.8+0.aarch64.apple.darwin14
Info: Successfully installed Julia 1.10.8.
julia> VERSION
v"1.10.8"
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -219,16 +219,16 @@ jobs:
         CARGO_TARGET_aarch64-unknown-linux-musl: ${{matrix.rustflags}}
     - name: FreeBSD release build
       if: ${{ contains(matrix.target, 'freebsd') }}
-      uses: vmactions/freebsd-vm@v1
+      uses: cross-platform-actions/action@v0.30.0
       with:
-        release: "13.4"
-        usesh: true
-        mem: 8192
-        copyback: true
-        prepare: |
-          pkg install -y curl
-          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain ${{matrix.toolchain}}
+        operating_system: freebsd
+        version: '13.4'
+        architecture: x86-64
+        cpu_count: 4
+        memory: '12G'
+        shell: sh
         run: |
+          fetch https://sh.rustup.rs -o - | sh -s -- -y --default-toolchain ${{matrix.toolchain}}
           . "${HOME}/.cargo/env"
           cargo build --release ${{matrix.bins}} --target ${{matrix.target}} --features ${{matrix.features}}
     - uses: actions/upload-artifact@v5
@@ -329,16 +329,16 @@ jobs:
         CARGO_TARGET_aarch64-unknown-linux-musl: ${{matrix.rustflags}}
     - name: Test FreeBSD
       if: ${{ contains(matrix.target, 'freebsd') }}
-      uses: vmactions/freebsd-vm@v1
+      uses: cross-platform-actions/action@v0.30.0
       with:
-        release: "13.4"
-        usesh: true
-        mem: 8192
-        copyback: false
-        prepare: |
-          pkg install -y curl
-          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain ${{matrix.toolchain}}
+        operating_system: freebsd
+        version: '13.4'
+        architecture: x86-64
+        cpu_count: 4
+        memory: '12G'
+        shell: sh
         run: |
+          fetch https://sh.rustup.rs -o - | sh -s -- -y --default-toolchain ${{matrix.toolchain}}
           . "${HOME}/.cargo/env"
           cargo test --target ${{matrix.target}} --features ${{matrix.features}}
   package-unix:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -214,16 +214,16 @@ jobs:
         CARGO_TARGET_aarch64-unknown-linux-musl: ${{matrix.rustflags}}
     - name: Test FreeBSD
       if: ${{ contains(matrix.target, 'freebsd') }}
-      uses: vmactions/freebsd-vm@v1
+      uses: cross-platform-actions/action@v0.30.0
       with:
-        release: "13.4"
-        usesh: true
-        mem: 8192
-        copyback: false
-        prepare: |
-          pkg install -y curl
-          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain ${{matrix.toolchain}}
+        operating_system: freebsd
+        version: '13.4'
+        architecture: x86-64
+        cpu_count: 4
+        memory: '12G'
+        shell: sh
         run: |
+          fetch https://sh.rustup.rs -o - | sh -s -- -y --default-toolchain ${{matrix.toolchain}}
           . "${HOME}/.cargo/env"
           export RUST_BACKTRACE=full
           cargo test --target ${{matrix.target}} --features ${{matrix.features}}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,27 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+  - repo: https://github.com/doublify/pre-commit-rust
+    rev: v1.0
+    hooks:
+      - id: fmt
+        name: cargo fmt
+        description: Format Rust code with rustfmt
+        entry: cargo fmt --all --
+        language: system
+        types: [rust]
+        pass_filenames: false
+      - id: cargo-check
+        name: cargo check
+        description: Check Rust code with cargo check
+        entry: cargo check
+        language: system
+        types: [rust]
+        pass_filenames: false
+      - id: clippy
+        name: cargo clippy
+        description: Lint Rust code with clippy
+        entry: bash -c 'cargo clippy --workspace -- -D warnings'
+        language: system
+        types: [rust]
+        pass_filenames: false
--- a/versiondb/versiondb-aarch64-apple-darwin.json
+++ b/versiondb/versiondb-aarch64-apple-darwin.json
@@ -501,6 +501,12 @@
         "1.11.7+0.x64.apple.darwin14": {
             "UrlPath": "bin/mac/x64/1.11/julia-1.11.7-mac64.tar.gz"
         },
+        "1.11.8+0.aarch64.apple.darwin14": {
+            "UrlPath": "bin/mac/aarch64/1.11/julia-1.11.8-macaarch64.tar.gz"
+        },
+        "1.11.8+0.x64.apple.darwin14": {
+            "UrlPath": "bin/mac/x64/1.11/julia-1.11.8-mac64.tar.gz"
+        },
         "1.12.0-beta1+0.aarch64.apple.darwin14": {
             "UrlPath": "bin/mac/aarch64/1.12/julia-1.12.0-beta1-macaarch64.tar.gz"
         },
@@ -554,6 +560,24 @@
         },
         "1.12.1+0.x64.apple.darwin14": {
             "UrlPath": "bin/mac/x64/1.12/julia-1.12.1-mac64.tar.gz"
+        },
+        "1.12.2+0.aarch64.apple.darwin14": {
+            "UrlPath": "bin/mac/aarch64/1.12/julia-1.12.2-macaarch64.tar.gz"
+        },
+        "1.12.2+0.x64.apple.darwin14": {
+            "UrlPath": "bin/mac/x64/1.12/julia-1.12.2-mac64.tar.gz"
+        },
+        "1.12.3+0.aarch64.apple.darwin14": {
+            "UrlPath": "bin/mac/aarch64/1.12/julia-1.12.3-macaarch64.tar.gz"
+        },
+        "1.12.3+0.x64.apple.darwin14": {
+            "UrlPath": "bin/mac/x64/1.12/julia-1.12.3-mac64.tar.gz"
+        },
+        "1.13.0-alpha2+0.aarch64.apple.darwin14": {
+            "UrlPath": "bin/mac/aarch64/1.13/julia-1.13.0-alpha2-macaarch64.tar.gz"
+        },
+        "1.13.0-alpha2+0.x64.apple.darwin14": {
+            "UrlPath": "bin/mac/x64/1.13/julia-1.13.0-alpha2-mac64.tar.gz"
         }
     },
     "AvailableChannels": {
@@ -576,7 +600,7 @@
             "Version": "0.7.0+0.x64.apple.darwin14"
         },
         "1": {
-            "Version": "1.12.1+0.aarch64.apple.darwin14"
+            "Version": "1.12.3+0.aarch64.apple.darwin14"
         },
         "1.0": {
             "Version": "1.0.5+0.x64.apple.darwin14"
@@ -807,7 +831,7 @@
             "Version": "1.10.10+0.x64.apple.darwin14"
         },
         "1.11": {
-            "Version": "1.11.7+0.aarch64.apple.darwin14"
+            "Version": "1.11.8+0.aarch64.apple.darwin14"
         },
         "1.11.0": {
             "Version": "1.11.0+0.aarch64.apple.darwin14"
@@ -953,14 +977,23 @@
         "1.11.7~x64": {
             "Version": "1.11.7+0.x64.apple.darwin14"
         },
+        "1.11.8": {
+            "Version": "1.11.8+0.aarch64.apple.darwin14"
+        },
+        "1.11.8~aarch64": {
+            "Version": "1.11.8+0.aarch64.apple.darwin14"
+        },
+        "1.11.8~x64": {
+            "Version": "1.11.8+0.x64.apple.darwin14"
+        },
         "1.11~aarch64": {
-            "Version": "1.11.7+0.aarch64.apple.darwin14"
+            "Version": "1.11.8+0.aarch64.apple.darwin14"
         },
         "1.11~x64": {
-            "Version": "1.11.7+0.x64.apple.darwin14"
+            "Version": "1.11.8+0.x64.apple.darwin14"
         },
         "1.12": {
-            "Version": "1.12.1+0.aarch64.apple.darwin14"
+            "Version": "1.12.3+0.aarch64.apple.darwin14"
         },
         "1.12.0": {
             "Version": "1.12.0+0.aarch64.apple.darwin14"
@@ -1043,11 +1076,47 @@
         "1.12.1~x64": {
             "Version": "1.12.1+0.x64.apple.darwin14"
         },
+        "1.12.2": {
+            "Version": "1.12.2+0.aarch64.apple.darwin14"
+        },
+        "1.12.2~aarch64": {
+            "Version": "1.12.2+0.aarch64.apple.darwin14"
+        },
+        "1.12.2~x64": {
+            "Version": "1.12.2+0.x64.apple.darwin14"
+        },
+        "1.12.3": {
+            "Version": "1.12.3+0.aarch64.apple.darwin14"
+        },
+        "1.12.3~aarch64": {
+            "Version": "1.12.3+0.aarch64.apple.darwin14"
+        },
+        "1.12.3~x64": {
+            "Version": "1.12.3+0.x64.apple.darwin14"
+        },
         "1.12~aarch64": {
-            "Version": "1.12.1+0.aarch64.apple.darwin14"
+            "Version": "1.12.3+0.aarch64.apple.darwin14"
         },
         "1.12~x64": {
-            "Version": "1.12.1+0.x64.apple.darwin14"
+            "Version": "1.12.3+0.x64.apple.darwin14"
+        },
+        "1.13": {
+            "Version": "1.13.0-alpha2+0.aarch64.apple.darwin14"
+        },
+        "1.13.0-alpha2": {
+            "Version": "1.13.0-alpha2+0.aarch64.apple.darwin14"
+        },
+        "1.13.0-alpha2~aarch64": {
+            "Version": "1.13.0-alpha2+0.aarch64.apple.darwin14"
+        },
+        "1.13.0-alpha2~x64": {
+            "Version": "1.13.0-alpha2+0.x64.apple.darwin14"
+        },
+        "1.13~aarch64": {
+            "Version": "1.13.0-alpha2+0.aarch64.apple.darwin14"
+        },
+        "1.13~x64": {
+            "Version": "1.13.0-alpha2+0.x64.apple.darwin14"
         },
         "1.1~x64": {
             "Version": "1.1.1+0.x64.apple.darwin14"
@@ -1554,28 +1623,28 @@
             "Version": "1.9.4+0.x64.apple.darwin14"
         },
         "1~aarch64": {
-            "Version": "1.12.1+0.aarch64.apple.darwin14"
+            "Version": "1.12.3+0.aarch64.apple.darwin14"
         },
         "1~x64": {
-            "Version": "1.12.1+0.x64.apple.darwin14"
+            "Version": "1.12.3+0.x64.apple.darwin14"
         },
         "alpha": {
-            "Version": "1.12.1+0.aarch64.apple.darwin14"
+            "Version": "1.13.0-alpha2+0.aarch64.apple.darwin14"
         },
         "alpha~aarch64": {
-            "Version": "1.12.1+0.aarch64.apple.darwin14"
+            "Version": "1.13.0-alpha2+0.aarch64.apple.darwin14"
         },
         "alpha~x64": {
-            "Version": "1.12.1+0.x64.apple.darwin14"
+            "Version": "1.13.0-alpha2+0.x64.apple.darwin14"
         },
         "beta": {
-            "Version": "1.12.1+0.aarch64.apple.darwin14"
+            "Version": "1.12.3+0.aarch64.apple.darwin14"
         },
         "beta~aarch64": {
-            "Version": "1.12.1+0.aarch64.apple.darwin14"
+            "Version": "1.12.3+0.aarch64.apple.darwin14"
         },
         "beta~x64": {
-            "Version": "1.12.1+0.x64.apple.darwin14"
+            "Version": "1.12.3+0.x64.apple.darwin14"
         },
         "lts": {
             "Version": "1.10.10+0.aarch64.apple.darwin14"
@@ -1587,23 +1656,23 @@
             "Version": "1.10.10+0.x64.apple.darwin14"
         },
         "rc": {
-            "Version": "1.12.1+0.aarch64.apple.darwin14"
+            "Version": "1.12.3+0.aarch64.apple.darwin14"
         },
         "rc~aarch64": {
-            "Version": "1.12.1+0.aarch64.apple.darwin14"
+            "Version": "1.12.3+0.aarch64.apple.darwin14"
         },
         "rc~x64": {
-            "Version": "1.12.1+0.x64.apple.darwin14"
+            "Version": "1.12.3+0.x64.apple.darwin14"
         },
         "release": {
-            "Version": "1.12.1+0.aarch64.apple.darwin14"
+            "Version": "1.12.3+0.aarch64.apple.darwin14"
         },
         "release~aarch64": {
-            "Version": "1.12.1+0.aarch64.apple.darwin14"
+            "Version": "1.12.3+0.aarch64.apple.darwin14"
         },
         "release~x64": {
-            "Version": "1.12.1+0.x64.apple.darwin14"
+            "Version": "1.12.3+0.x64.apple.darwin14"
         }
     },
-    "Version": "1.0.75"
+    "Version": "1.0.78"
 }
--- a/versiondb/versiondb-aarch64-unknown-linux-gnu.json
+++ b/versiondb/versiondb-aarch64-unknown-linux-gnu.json
@@ -348,6 +348,9 @@
         "1.11.7+0.aarch64.linux.gnu": {
             "UrlPath": "bin/linux/aarch64/1.11/julia-1.11.7-linux-aarch64.tar.gz"
         },
+        "1.11.8+0.aarch64.linux.gnu": {
+            "UrlPath": "bin/linux/aarch64/1.11/julia-1.11.8-linux-aarch64.tar.gz"
+        },
         "1.12.0-beta1+0.aarch64.linux.gnu": {
             "UrlPath": "bin/linux/aarch64/1.12/julia-1.12.0-beta1-linux-aarch64.tar.gz"
         },
@@ -374,6 +377,15 @@
         },
         "1.12.1+0.aarch64.linux.gnu": {
             "UrlPath": "bin/linux/aarch64/1.12/julia-1.12.1-linux-aarch64.tar.gz"
+        },
+        "1.12.2+0.aarch64.linux.gnu": {
+            "UrlPath": "bin/linux/aarch64/1.12/julia-1.12.2-linux-aarch64.tar.gz"
+        },
+        "1.12.3+0.aarch64.linux.gnu": {
+            "UrlPath": "bin/linux/aarch64/1.12/julia-1.12.3-linux-aarch64.tar.gz"
+        },
+        "1.13.0-alpha2+0.aarch64.linux.gnu": {
+            "UrlPath": "bin/linux/aarch64/1.13/julia-1.13.0-alpha2-linux-aarch64.tar.gz"
         }
     },
     "AvailableChannels": {
@@ -396,7 +408,7 @@
             "Version": "0.6.3+0.aarch64.linux.gnu"
         },
         "1": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.12.3+0.aarch64.linux.gnu"
         },
         "1.0": {
             "Version": "1.0.5+0.aarch64.linux.gnu"
@@ -552,7 +564,7 @@
             "Version": "1.10.10+0.aarch64.linux.gnu"
         },
         "1.11": {
-            "Version": "1.11.7+0.aarch64.linux.gnu"
+            "Version": "1.11.8+0.aarch64.linux.gnu"
         },
         "1.11.0": {
             "Version": "1.11.0+0.aarch64.linux.gnu"
@@ -650,11 +662,17 @@
         "1.11.7~aarch64": {
             "Version": "1.11.7+0.aarch64.linux.gnu"
         },
+        "1.11.8": {
+            "Version": "1.11.8+0.aarch64.linux.gnu"
+        },
+        "1.11.8~aarch64": {
+            "Version": "1.11.8+0.aarch64.linux.gnu"
+        },
         "1.11~aarch64": {
-            "Version": "1.11.7+0.aarch64.linux.gnu"
+            "Version": "1.11.8+0.aarch64.linux.gnu"
         },
         "1.12": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.12.3+0.aarch64.linux.gnu"
         },
         "1.12.0": {
             "Version": "1.12.0+0.aarch64.linux.gnu"
@@ -710,8 +728,32 @@
         "1.12.1~aarch64": {
             "Version": "1.12.1+0.aarch64.linux.gnu"
         },
+        "1.12.2": {
+            "Version": "1.12.2+0.aarch64.linux.gnu"
+        },
+        "1.12.2~aarch64": {
+            "Version": "1.12.2+0.aarch64.linux.gnu"
+        },
+        "1.12.3": {
+            "Version": "1.12.3+0.aarch64.linux.gnu"
+        },
+        "1.12.3~aarch64": {
+            "Version": "1.12.3+0.aarch64.linux.gnu"
+        },
         "1.12~aarch64": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.12.3+0.aarch64.linux.gnu"
+        },
+        "1.13": {
+            "Version": "1.13.0-alpha2+0.aarch64.linux.gnu"
+        },
+        "1.13.0-alpha2": {
+            "Version": "1.13.0-alpha2+0.aarch64.linux.gnu"
+        },
+        "1.13.0-alpha2~aarch64": {
+            "Version": "1.13.0-alpha2+0.aarch64.linux.gnu"
+        },
+        "1.13~aarch64": {
+            "Version": "1.13.0-alpha2+0.aarch64.linux.gnu"
         },
         "1.1~aarch64": {
             "Version": "1.1.1+0.aarch64.linux.gnu"
@@ -1209,19 +1251,19 @@
             "Version": "1.9.4+0.aarch64.linux.gnu"
         },
         "1~aarch64": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.12.3+0.aarch64.linux.gnu"
         },
         "alpha": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.13.0-alpha2+0.aarch64.linux.gnu"
         },
         "alpha~aarch64": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.13.0-alpha2+0.aarch64.linux.gnu"
         },
         "beta": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.12.3+0.aarch64.linux.gnu"
         },
         "beta~aarch64": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.12.3+0.aarch64.linux.gnu"
         },
         "lts": {
             "Version": "1.10.10+0.aarch64.linux.gnu"
@@ -1230,17 +1272,17 @@
             "Version": "1.10.10+0.aarch64.linux.gnu"
         },
         "rc": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.12.3+0.aarch64.linux.gnu"
         },
         "rc~aarch64": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.12.3+0.aarch64.linux.gnu"
         },
         "release": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.12.3+0.aarch64.linux.gnu"
         },
         "release~aarch64": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.12.3+0.aarch64.linux.gnu"
         }
     },
-    "Version": "1.0.75"
+    "Version": "1.0.78"
 }
--- a/versiondb/versiondb-aarch64-unknown-linux-musl.json
+++ b/versiondb/versiondb-aarch64-unknown-linux-musl.json
@@ -348,6 +348,9 @@
         "1.11.7+0.aarch64.linux.gnu": {
             "UrlPath": "bin/linux/aarch64/1.11/julia-1.11.7-linux-aarch64.tar.gz"
         },
+        "1.11.8+0.aarch64.linux.gnu": {
+            "UrlPath": "bin/linux/aarch64/1.11/julia-1.11.8-linux-aarch64.tar.gz"
+        },
         "1.12.0-beta1+0.aarch64.linux.gnu": {
             "UrlPath": "bin/linux/aarch64/1.12/julia-1.12.0-beta1-linux-aarch64.tar.gz"
         },
@@ -374,6 +377,15 @@
         },
         "1.12.1+0.aarch64.linux.gnu": {
             "UrlPath": "bin/linux/aarch64/1.12/julia-1.12.1-linux-aarch64.tar.gz"
+        },
+        "1.12.2+0.aarch64.linux.gnu": {
+            "UrlPath": "bin/linux/aarch64/1.12/julia-1.12.2-linux-aarch64.tar.gz"
+        },
+        "1.12.3+0.aarch64.linux.gnu": {
+            "UrlPath": "bin/linux/aarch64/1.12/julia-1.12.3-linux-aarch64.tar.gz"
+        },
+        "1.13.0-alpha2+0.aarch64.linux.gnu": {
+            "UrlPath": "bin/linux/aarch64/1.13/julia-1.13.0-alpha2-linux-aarch64.tar.gz"
         }
     },
     "AvailableChannels": {
@@ -396,7 +408,7 @@
             "Version": "0.6.3+0.aarch64.linux.gnu"
         },
         "1": {
-            "Version": "1.12.1+0.aarch64.linux.gnu"
+            "Version": "1.12.3+0.aarch64.linux.gnu"
         },
         "1.0": {
             "Version": "1.0.5+0.aarch64.linux.gnu"
@@ -552,7 +564,7 @@
             "Version": "1.10.10+0.aarch64.linux.gnu"
         },
         "1.11": {
-            "Version": "1.11.7+0.aarch64.linux.gnu"
+            "Version": "1.11.8+0.aarch64.linux.gnu"
         },
         "1.11.0": {
             "Version": "1.11.0+0.aarch64.linux.gnu"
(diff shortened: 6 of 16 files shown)


src/bin/julialauncher.rs
```diff
@@ -179,6 +182,310 @@ fn is_interactive() -> bool {
...
+        }
+        Some(_) => {
+            // Invalid manifest field type, fall through to default search
+        }
+        None => {}
+    }
+
+    // Search for manifest files in priority order
+    // JuliaManifest.toml takes precedence over Manifest.toml
+    for manifest_name in ["JuliaManifest.toml", "Manifest.toml"] {
```
DilumAluthge: Needs support for `{Julia,}Manifest-vX.Y.toml`

DilumAluthge:
How should this work if the project directory has multiple versioned manifests, e.g. `Manifest-1.10.toml` and `Manifest-1.12.toml`?

KristofferC:
Probably the highest version?

DilumAluthge:
> Probably the highest version?

That seems okay.

If the user explicitly provides the desired version (e.g. `julia +1.10 --project` or `julia +1.10.3 --project` or `julia +lts --project`), we'd respect the Julia version specified by `julia +ver`, right?

src/bin/julialauncher.rs
DilumAluthge: One idea:
1. Loop over all files in the directory. Ignore folders.
2. Collect only the filenames that match the allowed patterns.

KristofferC:
> If the user explicitly provides the desired version (e.g. julia +1.10 --project or julia +1.10.3 --project or julia +lts --project), we'd respect the Julia version specified by julia +ver, right?

Yes, if you look at the priority in the README (which I updated here) it is:

```
1. A command line Julia version specifier, such as `julia +release`.
2. The `JULIAUP_CHANNEL` environment variable.
3. A directory override, set with the `juliaup override set` command.
3. The default Juliaup channel.
4. Automatic version selection based on the active project.
5. The default Juliaup channel.
```

So an explicitly provided version has top-priority and will always be used.

DilumAluthge:
~~BTW I think item 3 should be removed from that list? Redundant with item 5, and IIUC you prefer item 4 over item 5?~~

EDIT: This was just a confusion over a typo in a GitHub comment. The actual code in the PR was correct.

src/bin/julialauncher.rs
davidanthoff: I think it would also be reasonable (and in my mind a bit less magical) to say that version selection via the manifest _only_ happens from a `JuliaManifest.toml` or `Manifest.toml` file.

The `{Julia,}Manifest-vX.Y.toml` manifests would then only come into play if one used a higher precedent order override.

KristofferC:
I don't fully understand why it is redundant or why this PR makes it more redundant.

davidanthoff:
I believe @DilumAluthge was just looking at the list as pasted in the comment here, where there are two 3) items. It is all correct in the PR that modifies the README.

KristofferC:
Oops, I must have copy pasted from the diff.

KristofferC:
I realize a better way of testing this is (of course) to call julia itself with the same project flag and env variable and get the truth from that.

IanButterworth:
Just had a thought.. how does this handle manifests from julia master, i.e. 1.14, given that there's no "1.14" channel? I guess there isn't a "1.13" either.

KristofferC:
It uses nightly channel if the version is higher than any known version. But maybe it should emit some warning when that happens...

tecosaur:
I may be missing something, but why is this ~3x the LOC of #1095? Is this PR larger in scope?

I do notice that #1095 was written before versioned manifests became a thing, but that seems like just a handful of extra lines.

KristofferC:
I can remove tests if you want to... But as I have said, you are free to reopen your PR and we can try them on a bunch of cases. But based on

> I don't think it's realistic that I'll get around to writing the missing tests, let alone resolving the merge conflicts any time soon.

I thought you didn't plan on getting back to that "any time soon" so I started this. I don't know if four days counts as "any time soon" or if just the fact that someone else started working on it made your desire come back.

However, some things that this does that might be different from that branch:

- Launch `+nightly` if the version is > than anyone version we know
- Launch `1.x+nightly` if x is > than any version we now for 1.x
- Expand `~` to homedir
- `LOAD_PATH`
- Logging
- Versioned manifests
- Maybe more, didn't look more.

tecosaur:
> I can remove tests if you want to... But as I have said, you are free to reopen your PR and we can try them on a bunch of cases. But based on

Oh, I'm talking about the non-testing code. Taking a slightly closer look, it seems like ~700 LOC compared to ~400 in the original PR, which doesn't seem too bad given the extra functionality you mentioned.

> I thought you didn't plan on getting back to that "any time soon" so I started this. I don't know if four days counts as "any time soon" or if just the fact that someone else started working on it made your desire come back.

Yea, I wasn't thinking of getting back to this any time soon when I said that. I also have a track record of getting around to things later than I expected, which I try to account for.

After making that comment, and given you said that functionality seems needed, I then thought I should just "take a peek" at it that evening to see how bad the merge conflicts really are, and then we talked about how the functionality should work over Slack, and all that moved the work from back-of-mind to front-of-mind. I tend to work on what I'm thinking about.

Net result: I've ended up poking at this months earlier than I thought I would. If there is value in me spinning up that branch again, it's much more viable than it was four days ago. However, if Sonnet 4.5 can't take care of the test writing, or for some other reason the work ends up in limbo again, I don't currently have the motivation in this area to keep working on this for a prolonged period.

Looking at how hard adding project/nightly/versioned manifest support is, and doing it if it's easy, I am up for (but not much more).

KristofferC:
I rewrote a lot of the implementation here because I realized (while being on the treadmill :running_man:) that it made a lot of sense to just directly "transpile" the julia code itself that does the project and manifest file resolving. That should make it easier to review, easier to update and more likely to be correct. 

As a test of one of the functionality I wanted to use this for I did:

```
(@v1.12) pkg> app add https://github.com/KristofferC/Rot13.jl
...
[ Info: For package: Rot13 installed apps rot13
```

which installs the `rot13` Julia app. I then set  the julia command to run apps with to the `juliaup` `julia` and we can see how it detects the julia version in the manifest for the app (which requires `JULIA_LOAD_PATH` handling support) and uses the correct version:

```
❯ JULIAUP_LOG=Debug JULIA_APPS_JULIA_CMD=~/juliaup/target/debug/julia rot13 foobar
[2025-11-03T13:54:58Z DEBUG julia] VersionDetect::Found valid project in JULIA_LOAD_PATH entry: /home/kc/.julia/environments/apps/Rot13
[2025-11-03T13:54:58Z DEBUG julia] VersionDetect::Using project file: /home/kc/.julia/environments/apps/Rot13/Project.toml
[2025-11-03T13:54:58Z DEBUG julia] VersionDetect::Detected manifest file: /home/kc/.julia/environments/apps/Rot13/Manifest.toml
[2025-11-03T13:54:58Z DEBUG julia] VersionDetect::Read Julia version from manifest: 1.12.1
sbbone
```

KristofferC:
I think I am fairly happy with this now. Maybe the number of tests is a bit excessive; I basically added a test as soon as I found something missing or buggy.

DilumAluthge:
Personally I'm always a fan of having lots of tests.

KristofferC:
Bump, I've run this locally for a while and it works very nicely for me at least.

DilumAluthge:
The idea sounds good to me. It'd be nice to have someone review the Rust code itself. I'm not familiar with Rust. @MilesCranmer would you be able to review the code?

README.md
```diff
@@ -156,10 +156,25 @@ The Julia launcher `julia` automatically determines which specific version of Ju
 1. A command line Julia version specifier, such as `julia +release`.
 2. The `JULIAUP_CHANNEL` environment variable.
 3. A directory override, set with the `juliaup override set` command.
-3. The default Juliaup channel.
+4. Automatic version selection based on the active project.
```
DilumAluthge: Can you make this a clickable link to the "Project-based Version Selection" section below?

DilumAluthge:
Also needs a rebase because of conflicts in the Cargo lockfile.

MilesCranmer:
Can this be broken up into smaller PRs? 1843 lines in the main source file is a bit hard to review

MilesCranmer:
Oh sorry I didn't realise the tests were embedded into `src/bin/julialauncher.rs` itself. Those should be put in `tests` folder instead. It does technically work, but all the other tests live in `tests/` so it will simplify things to be consistent

Cargo.toml
```diff
@@ -56,11 +56,13 @@ dialoguer = "0.11.0"
...
 env_proxy = "0.4.1"
 bstr = "1.10"
-indoc = "2.0.7"
+indoc = "2.0.3"
 is-terminal = "0.4"
 path-absolutize = "3.1.0"
 numeric-sort = "0.1.5"
 regex = "1.10"
+toml = "0.8"
+expand-tilde = "0.6.1"
```
MilesCranmer: You should use `shellexpand` since that is already used elsewhere in juliaup

DilumAluthge:
Thanks so much @MilesCranmer!

src/bin/julialauncher.rs
```diff
@@ -180,6 +183,680 @@ fn is_interactive() -> bool {
...
+                        log::debug!("VersionDetect::No project specification found");
+                        return Ok(None);
+                    }
+                }
+            } else {
+                log::debug!("VersionDetect::No project specification found");
+                return Ok(None);
+            }
+        }
+    };
```
MilesCranmer: Feels a little overcomplicated/hard to read. Also it seems this returns from the function but also assigns to a variable?

src/bin/julialauncher.rs
```diff
@@ -791,3 +1512,1051 @@ fn main() -> Result<std::process::ExitCode> {
     // TODO https://github.com/rust-lang/rust/issues/111688 is finalized, we should use that instead of calling exit
     std::process::exit(client_status?);
 }
+
+#[cfg(test)]
+mod tests {
```
MilesCranmer: I think these tests should be refactored, there is a lot of code that gets repeated

src/bin/julialauncher.rs
```diff
@@ -553,6 +1247,48 @@ fn get_override_channel(
...
+    } else if let Some(channel) = override_channel {
+        Ok((channel, JuliaupChannelSource::Override))
+    } else if let Some(channel) = get_auto_channel(args, versions_db) {
+        Ok((channel, JuliaupChannelSource::Auto))
+    } else if let Some(channel) = default_channel {
+        Ok((channel, JuliaupChannelSource::Default))
+    } else {
+        Err(anyhow!("Failed to determine juliaup channel"))
+    }
+}
```
MilesCranmer: This seems like an unrelated refactoring?

KristofferC:
> Oh sorry I didn't realise the tests were embedded into src/bin/julialauncher.rs itself. Those should be put in tests folder instead. It does technically work, but all the other tests live in tests/ so it will simplify things to be consistent

This is not true? Unit tests seem to go in-file while "integration tests" seem to be in `test`.

src/bin/julialauncher.rs
```diff
@@ -180,6 +183,680 @@ fn is_interactive() -> bool {
...
+    for key in versions_db.available_channels.keys() {
+        if let Ok(version) = parse_db_version(key) {
+            if version.major == major && version.minor == minor {
+                max_version = match &max_version {
+                    Some(current) if current >= &version => Some(current.clone()),
+                    _ => Some(version),
+                };
+            }
+        }
+    }
```
MilesCranmer: Feels overly nested

src/bin/julialauncher.rs
```diff
@@ -180,6 +183,680 @@ fn is_interactive() -> bool {
...
+        // Skip empty entries and special entries that don't represent projects
+        if entry.is_empty() || entry == "@" || entry.starts_with("@v") || entry == "@stdlib" {
+            continue;
+        }
+
+        // Try to expand this load path entry as a project
+        if let Ok(Some(project_file)) = load_path_expand_impl(entry, current_dir, depot_path) {
+            // Check if this project has a manifest
+            if project_file_manifest_path(&project_file).is_some() {
+                log::debug!(
```
MilesCranmer: Not sure what the convention is for Juliaup. But note that `log::debug!` is only used once in the codebase; this PR adds 10 more uses.

src/bin/julialauncher.rs
```diff
@@ -180,6 +183,680 @@ fn is_interactive() -> bool {
...
+        opt,
+        // no_argument options
+        "--version" | "--help" | "--help-hidden" | "--interactive" | "--quiet"
+            | "--experimental" | "--lisp" | "--image-codegen" | "--rr-detach"
+            | "--strip-metadata" | "--strip-ir" | "--gc-sweep-always-full"
+            | "--trace-compile-timing"
+            // optional_argument options
+            | "--project" | "--code-coverage" | "--track-allocation" | "--optimize"
+            | "--min-optlevel" | "--debug-info" | "--worker" | "--trim" | "--trace-eval"
+    )
```
MilesCranmer: Should really be using clap.rs for this. I mean the best solution is something like #1236 imo. But for now maybe this will do. It's definitely not idiomatic rust tho.

src/bin/julialauncher.rs
```diff
@@ -180,6 +183,680 @@ fn is_interactive() -> bool {
...
+        ));
+    }
+
+    // Otherwise, it's a path
+    let path = expand_tilde(env)?.into_owned();
+    let path = if path.is_absolute() {
+        path
+    } else {
+        current_dir.join(path)
+    };
```
MilesCranmer: Seems like this is trying to mutate an immutable; maybe just write it like

```rust
let mut path = expand_tilde(env)?.into_owned();
if path.is_relative() {
    path = current_dir.join(path);
}
```

But as mentioned elsewhere, you should avoid `expand_tilde` as there's already another package loaded for that

src/bin/julialauncher.rs
```diff
@@ -180,6 +183,680 @@ fn is_interactive() -> bool {
...
+
+    // Handle short options: from shortopts = "+vhqH:e:E:L:J:C:it:p:O:g:m:"
+    // Options WITHOUT ':' are no_argument, 'O' and 'g' are optional_argument
+    if let Some(short) = opt.strip_prefix('-').and_then(|s| {
+        if !s.starts_with('-') && s.len() == 1 {
+            s.chars().next()
+        } else {
+            None
+        }
+    }) {
```
MilesCranmer: A bit unreadable; I would make a variable for this instead

MilesCranmer:
These are integration tests though

src/bin/julialauncher.rs
KristofferC: I don't get it. How is clap.rs relevant for this? That crate is for making a CLI no?

src/bin/julialauncher.rs
```diff
@@ -180,6 +183,680 @@ fn is_interactive() -> bool {
...
+    for entry in entries.flatten() {
+        let path = entry.path();
+        if let Some(filename) = path.file_name().and_then(|n| n.to_str()) {
+            // Check for versioned manifests: JuliaManifest-v*.toml or Manifest-v*.toml
+            let (prefix, target) = if filename.starts_with("JuliaManifest-v") {
+                ("JuliaManifest-v", &mut highest_julia)
+            } else if filename.starts_with("Manifest-v") {
+                ("Manifest-v", &mut highest_manifest)
+            } else {
+                continue;
```
MilesCranmer: Just from a readability perspective I'd rather avoid having something that reads like `(prefix, target) = continue` (even though that's not what is actually executed, of course)

src/bin/julialauncher.rs
KristofferC: Yeah, I find them useful... Here is an example:

```
❯ JULIAUP_LOG=debug  ~/juliaup2/target/debug/julia --project -q
[2025-11-13T20:22:13Z DEBUG julia] AutoVersionDetect::Using project specification: @.
[2025-11-13T20:22:13Z DEBUG julia] AutoVersionDetect::Found Project.toml at: /Users/kc/JuliaPkgs/Pkg.jl/Project.toml
[2025-11-13T20:22:13Z DEBUG julia] AutoVersionDetect::Using Manifest.toml
[2025-11-13T20:22:13Z DEBUG julia] AutoVersionDetect::Detected manifest file: /Users/kc/JuliaPkgs/Pkg.jl/Manifest.toml
[2025-11-13T20:22:13Z DEBUG julia] AutoVersionDetect::Read Julia version from manifest: 1.14.0-DEV
```

I can remove some if you like? How many should there be?

src/bin/julialauncher.rs
KristofferC: No, because it makes this part of the code unit testable and we want to test that we get the correct channel in many different cases with the version detection.

KristofferC:
> These are integration tests though

The integration test was put in `tests/channel_selection.rs`

src/bin/julialauncher.rs
```diff
@@ -180,6 +183,680 @@ fn is_interactive() -> bool {
...
+            };
+
+            if let Some(stripped) = filename.strip_prefix(prefix) {
+                if let Some(version_str) = stripped.strip_suffix(".toml") {
+                    if let Some(version) = parse_version_lenient(version_str) {
+                        update_highest(target, version, path);
+                    }
+                }
+            }
```
MilesCranmer:
```suggestion
            if let Some(version) = filename
                .strip_prefix(prefix)
                .and_then(|s| s.strip_suffix(".toml"))
                .and_then(parse_version_lenient)
            {
                update_highest(target, version, path);
            }
            
```

MilesCranmer: Though actually it might be nice to just have a dedicated function for parsing version numbers from .toml files?

MilesCranmer:
> The integration test was put in `tests/channel_selection.rs`

Oh yeah ok. Idk. It's just a lot of tests! I still kinda feel like these guys should sit in the tests folder. If it was a few property tests that someone could read to help understand the code, ok, but generating temporary folders and mocking up files feels like it should sit in the normal tests folder?

src/bin/julialauncher.rs
KristofferC: Yes, but note that this is for basically parsing `x.y` from the `Manifest-x.y.toml` file name.

src/bin/julialauncher.rs
MilesCranmer: Ok. Could it be put in a separate PR though?

src/bin/julialauncher.rs
MilesCranmer: Yeah. It also validates options. The main reason is to avoid maintaining multiple sources of truth for all the julia options, which is what this does. i.e., whenever there's a new julia option, we would have to add it here as well

src/bin/julialauncher.rs
```diff
@@ -179,6 +183,680 @@ fn is_interactive() -> bool {
...
+
+    log::debug!("VersionDetect::Manifest file exists but does not contain julia_version field");
+    Ok(None)
+}
+
+/// Find highest versioned manifest, preferring JuliaManifest over Manifest for same version
+fn find_highest_versioned_manifest(project_root: &Path) -> Option<PathBuf> {
+    let Ok(entries) = fs::read_dir(project_root) else {
+        return None;
+    };
```
MilesCranmer: Does this read all files in the directory? Maybe we could just filter matching files, and then do a single reduction operation on top of that? It might be easier to read too

KristofferC:
>  I still kinda feel like these guys should sit in the tests folder. If it was a few property tests that someone could read to help understand the code, ok, but generating temporary folders and mocking up files feels like it should sit in the normal tests folder?

I am not very experienced with rust, but AFAIU the `test` tests are for public functions (integration tests) while you put tests for private functions into the file itself like is done here. I am not sure how to change things to run these tests from a separate folder, making everything public doesn't seem nice?

src/bin/julialauncher.rs
```diff
@@ -179,6 +183,680 @@ fn is_interactive() -> bool {
...
+        })
+    };
+
+    let project = match project {
+        None => return Ok(None),
+        Some(p) => p,
+    };
+
+    // Expand the project spec using load_path_expand
+    load_path_expand_impl(&project, current_dir, depot_path)
```
MilesCranmer:
```suggestion
    match project {
        None => Ok(None),
        Some(p) => load_path_expand_impl(&p, current_dir, depot_path)
    }
```

src/bin/julialauncher.rs
MilesCranmer: Yea it seems useful. Maybe `julia -v` could trigger it?

src/bin/julialauncher.rs
```diff
@@ -179,6 +183,680 @@ fn is_interactive() -> bool {
...
+    } else {
+        // Check JULIA_PROJECT env
+        julia_project.map(|v| {
+            if v.trim().is_empty() {
+                "@.".to_string() // Empty JULIA_PROJECT means "@."
+            } else {
+                v.to_string()
+            }
+        })
+    };
```
MilesCranmer: A lot of this stuff I think we should just handle with clap.rs (e.g., via #1236), then we don't have to write a custom parser for command line options

KristofferC: These need to exactly mimic julia's style of parsing. (Like the weird `--proj` stuff). But if clap can be made to do that after your PR we can switch to it.

src/bin/julialauncher.rs
KristofferC: Not really, it is not needed in a separate PR?

src/bin/julialauncher.rs
KristofferC: It is giving debug messages about the running juliaup code. `julia -v` prints the julia version?

MilesCranmer:
> I am not very experienced with rust, but AFAIU the `test` tests are for public functions (integration tests) while you put tests for private functions into the file itself like is done here. I am not sure how to change things to run these tests from a separate folder, making everything public doesn't seem nice?

juliaup is a binary crate, not a library, so there isn't any external API surface to protect, and nothing changes in the binary itself when you declare something public/private. From the perspective of readability I think it's just easier to maintain (in my view) to keep the tests and src separate, especially if they are substantial like these ones.

Note that you can also declare public test-only entry points with a `cfg(test)` directive.

KristofferC:
Ok, I'll make them pub and move them.

KristofferC:
> Ok, I'll make them pub and move them.

I didn't get this to work, the only way I could split out the tests file was to cut up `julialauncher.rs` into `julialauncher/main.rs` and `julialauncher/tests.rs` and do like a 

```
#[cfg(test)]
mod tests;
```

inside `main.rs`. Would that suffice?

MilesCranmer:
Actually, it might be better to move the new utility functions to a file `src/version_selection.rs`, and then both the `julialauncher.rs` file and a tests file `tests/version_selection.rs` can refer to those functions?

src/bin/julialauncher.rs
MilesCranmer: Oh yeah, oops! Then, I don't know. Is there any equivalent design to compare to in normal Julia, like a way to get more verbose information at launch, about, e.g., what Project/Manifest is being used, what depot is activated, etc? Otherwise I just have no frame of reference other than the current juliaup codebase not ever doing stuff like this - so it feels a bit inconsistent.

DilumAluthge: For normal Julia, you can get some debug logs on code loading by setting e.g. `JULIA_DEBUG=loading` (or `JULIA_DEBUG=all`).

src/bin/julialauncher.rs
MilesCranmer: Ok. I have no idea about this part though, so maybe someone else can review it. It's also probably worth dogfooding to try to break it.

It's hard for me to think of failure cases by looking at it, since I have very little familiarity with the way the Julia CLI parsing works. Since this would immediately impact many Julia users it feels very important to get right. So if this was generated by claude, I am a bit worried to give approval on this part without a better understanding of what is supposed to happen

MilesCranmer: (Of course I understand my approval isn't necessary; just saying I wouldn't be able to say "Yes, this part is valid" on this chunk of the code)

src/bin/julialauncher.rs
MilesCranmer: Ok not a big issue from me if it doesn't feel right to split it

src/bin/julialauncher.rs
```diff
@@ -179,6 +182,684 @@ fn is_interactive() -> bool {
...
+                break;
+            }
+        }
+
+        // Move to parent
+        match current.parent() {
+            Some(parent) if parent != current => current = parent,
+            _ => break,
+        }
+    }
```
MilesCranmer: What happens for symbolic links? Is it possible to hit a cycle => infinite loop? As a quick fix maybe just set some large maximum search depth

MilesCranmer: Actually this isn't an issue, nevermind!

src/bin/julialauncher.rs
MilesCranmer: Cool. What is the format of these? Can we make these consistent to the julia logs?

davidanthoff:
I started to look through this. Not finished, but I thought I'd better leave some feedback early.

Lots of tests are obviously amazing, all in favor.

Agree with ideally moving things into more files, i.e. tests into their own files, and maybe also things like command line parsing etc.

I think I'm most stuck right now on the question whether we should mimic this whole `LOAD_PATH` story from base, especially given that we would have a slightly modified version no matter what (I think?). I think what I find most confusing about the base story is that it seems one can both reference the active project as an entry in `LOAD_PATH` via a `@` entry, but then it also seems that if no active project is explicitly set (via `JULIA_PROJECT` or `--project`), then Julia seems to use some heuristic to set an active project based on the content of `LOAD_PATH`. I find that latter step really confusing and somewhat circular, and I'm wondering whether we could just skip that for this PR. The most common case of "active project is set by looking through the `LOAD_PATH`" is actually one that we could _not_ replicate in the `julialauncher`, which to me is even more of an argument to just not mimic the `LOAD_PATH` behavior in the `julialauncher` (where the most common case is that no active project was set via env var or command line, and then the second entry `@v#.#` becomes the active project, which of course we can't do in the launcher). So right now I think my preference would be that we just completely ignore `JULIA_LOAD_PATH` for determining the active project and _only_ rely on the content of `JULIA_PROJECT` and the command line flag. That would simplify things a ton, and I don't really see a big downside. And then the story would just be that `LOAD_PATH` only controls package loading.

I also still think that https://github.com/JuliaLang/juliaup/issues/1111 would be a good idea, and if we were to do that, then this whole `LOAD_PATH` complexity would really not be needed.

The previous PR https://github.com/JuliaLang/juliaup/pull/1095 put all of this new stuff behind a feature configuration option, and I'm wondering whether we should do the same here. The thinking back then was that some of this design work might need iteration, and now with this `LOAD_PATH` story I think that is even more true. Replicating that feature flag story here would allow us to ship a rough version quicker, iterate on it based on opt-in user feedback and then at some point roll it out fully to everyone.

I'll continue to look through the PR :) But also just thank you for taking over, great to see progress here.

KristofferC:
> I think I'm most stuck right now on the question whether we should mimic this whole LOAD_PATH story from base, especially given that we would have a slightly modified version no matter what (I think?). ...

I don't really understand the argument here. Julia itself is the "oracle", it will resolve to whatever project it does. If we do not manage to resolve to the same project we will not launch the corresponding manifest julia version. The code right now is made to be as similar to what Julia itself does. You say that we should change this so we do not resolve the same project as julia and launch the wrong julia version? Why? Because the julia project resolution process is messy? It is not juliaups job to decide on how project resolution happens, changes to that can be made in julia and once they are simplified there, the corresponding simplification can be done here.

> I also still think that https://github.com/JuliaLang/juliaup/issues/1111 would be a good idea, and if we were to do that, then this whole LOAD_PATH complexity would really not be needed.

Again, I am confused. That seems like something that should be discussed in the julia repo, not here? This is just a launcher.

> The previous PR https://github.com/JuliaLang/juliaup/pull/1095 put all of this new stuff behind a feature configuration option, and I'm wondering whether we should do the same here.

I really think not. If you are worried it is buggy it could either be launched on the pre-release channel and have people test it out or it can be put behind a (run time) opt in. But having it as basically a compile time options seems weird.

MilesCranmer:
I think a feature configuration option would also be nice tbh. Something like:

```
cargo install --git https://github.com/JuliaLang/juliaup --features auto-channel
```

These sorts of things are pretty common in rust b/c it's fairly easy to enable/disable stuff behind compilation options.

Auto-selecting the channel would be a behavioural change so I don't think there's much harm in announcing it as opt-in and having a dogfood period. Or even keeping it opt-in for a while maybe. At the moment I do have the intuition that a `julia` command will always select the default channel, so this would fundamentally require a change in thinking, even though it's obviously a positive change

src/bin/julialauncher.rs
KristofferC: Again, this is to give debug messages about the juliaup code, I am confused why we are talking about `JULIA_DEBUG=loading` etc.

KristofferC:
> These sorts of things are pretty common in rust b/c it's fairly easy to enable/disable stuff behind compilation options.

What's the advantage? To avoid the toml dependency? Makes it much more complicated for people to try out if it is opt-in. Just because something can be done (and even easily) doesn't mean it should be done.

MilesCranmer:
I think runtime opt-in would be good too (since it is a behavioural change), it just means there's a bit more complexity compared to a compilation flag. It is subjective though. Also I don't feel strongly about this so I'll leave it up to you

KristofferC:
> Actually, it might be better to move the new utility functions to a file src/version_selection.rs, and then both the julialauncher.rs file and a tests file tests/version_selection.rs can refer to those functions?

The `determine_channel` function would have to go there too, then since we use that for the unit tests (and you cannot do the integration tests in `test` of functions in the binary).

But I mean, If I read the rust manual https://doc.rust-lang.org/book/ch11-03-test-organization.html it says:

> The purpose of unit tests is to test each unit of code in isolation from the rest of the code to quickly pinpoint where code is and isn’t working as expected. You’ll put unit tests in the src directory in each file with the code that they’re testing. The convention is to create a module named tests in each file to contain the test functions and to annotate the module with cfg(test).

which is what is done here. Why are we going against what the manual suggests in terms of code organization? If you want, I can just delete some tests.

KristofferC:
Anyway, I tried some kind of reorganization in the last commit.

MilesCranmer:
I'm not sure, I would guess they just mean tiny pure unittests of specific properties of helper functions? The tests here feel heavier. Of course this is all subjective. Thanks for moving it though.

For me a useful point of reference was to read through other crates like https://github.com/serde-rs/json, https://github.com/clap-rs/clap/tree/master/clap_complete, https://github.com/dtolnay/anyhow, which tend to avoid the inline tests (for whatever reason)

src/version_selection.rs
```diff
@@ -0,0 +1,664 @@
...
+        if versions_db
+            .available_channels
+            .contains_key(&versioned_nightly)
+        {
+            eprintln!(
+                "{} Manifest specifies prerelease Julia {}. Using {} channel.",
+                console::style("Info:").cyan().bold(),
+                required,
+                versioned_nightly
+            );
```
MilesCranmer: Maybe use this all the time, or `log::debug!` all the time? Or is there some difference between these and how they are used?

Also all of these `eprintln!` use the same console style so should be put into a separate function (a macro in this case)

davidanthoff:
I'll respond to the other points when I have more time, but just on this one:

> I really think not. If you are worried it is buggy it could either be launched on the pre-release channel and have people test it out or it can be put behind a (run time) opt in. But having it as basically a compile time options seems weird.

It is a runtime opt-in. You can probably cherry-pick https://github.com/JuliaLang/juliaup/commit/3a2c51ac2f3b36fa524a6f4a69a6c069e863cdb7 and then https://github.com/JuliaLang/juliaup/commit/343a4243e6458a1afaad221af6fe58cc655e581f was the actual runtime `if` part of the feature flag.

src/version_selection.rs
```diff
@@ -0,0 +1,664 @@
...
+    let update_highest =
+        |highest: &mut Option<(Version, PathBuf)>, version: Version, path: PathBuf| match highest {
+            Some((current_version, _)) if version > *current_version => {
+                *highest = Some((version, path));
+            }
+            None => {
+                *highest = Some((version, path));
+            }
+            _ => {}
+        };
```
MilesCranmer:
```suggestion
```

src/version_selection.rs
```diff
@@ -0,0 +1,664 @@
...
+            } else {
+                ("Manifest-v", &mut highest_manifest)
+            };
+
+            if let Some(version) = filename
+                .strip_prefix(prefix)
+                .and_then(|s| s.strip_suffix(".toml"))
+                .and_then(parse_version_lenient)
+            {
+                update_highest(target, version, path);
```
MilesCranmer:
```suggestion
                // Update highest if this version is greater or if none exists yet
                let should_update = match target {
                    Some((current_version, _)) => version > *current_version,
                    None => true,
                };
                if should_update {
                    *target = Some((version, path));
                }
```

src/version_selection.rs
```diff
@@ -0,0 +1,664 @@
...
+
+    log::debug!("VersionDetect::Manifest file exists but does not contain julia_version field");
+    Ok(None)
+}
+
+/// Find highest versioned manifest, preferring JuliaManifest over Manifest for same version
+pub fn find_highest_versioned_manifest(project_root: &Path) -> Option<PathBuf> {
+    let Ok(entries) = fs::read_dir(project_root) else {
+        return None;
+    };
```
MilesCranmer:
```suggestion
    let entries = fs::read_dir(project_root).ok()?;
```

src/version_selection.rs
```diff
@@ -0,0 +1,664 @@
...
+pub fn get_auto_channel(args: &[String], versions_db: &JuliaupVersionDB) -> Option<String> {
+    determine_project_version_spec(args)
+        .and_then(|opt_version| {
+            opt_version
+                .map(|version| resolve_auto_channel(version, versions_db))
+                .transpose()
+        })
+        .ok()
+        .flatten()
+}
```
MilesCranmer:
```suggestion
pub fn get_auto_channel(args: &[String], versions_db: &JuliaupVersionDB) -> Result<Option<String>>
{
    if let Some(required_version) = determine_project_version_spec(args)? {
        resolve_auto_channel(required_version, versions_db).map(Some)
    } else {
        Ok(None)
    }
}
```

I _think_ this should be equivalent

src/bin/julialauncher.rs
```diff
@@ -553,6 +580,48 @@ fn get_override_channel(
...
+
+    // Priority order
+    if let Some(channel) = channel_from_cmd_line {
+        Ok((channel, JuliaupChannelSource::CmdLine))
+    } else if let Some(channel) = env_channel {
+        Ok((channel, JuliaupChannelSource::EnvVar))
+    } else if let Some(channel) = override_channel {
+        Ok((channel, JuliaupChannelSource::Override))
+    } else if let Some(channel) = get_auto_channel(args, versions_db) {
```
MilesCranmer:
```suggestion
    } else if let Some(channel) = get_auto_channel(args, versions_db)? {
```

Assuming you take the other change

src/version_selection.rs
```diff
@@ -0,0 +1,664 @@
...
+
+    // Return highest version, preferring JuliaManifest for same version
+    match (highest_julia, highest_manifest) {
+        (Some((jv, jpath)), Some((mv, mpath))) => {
+            if jv >= mv {
+                Some(jpath)
+            } else {
+                Some(mpath)
+            }
+        }
```
MilesCranmer:
```suggestion
        (Some((jv, jpath)), Some((mv, mpath))) => Some(if jv >= mv { jpath } else { mpath }),
```

src/version_selection.rs
```diff
@@ -0,0 +1,664 @@
...
+    })?;
+
+    // Handle prerelease versions (e.g., 1.12.1-DEV, 1.13.0-rc1)
+    // Prereleases should use nightly channels because they represent development/testing versions
+    if !required_version.pre.is_empty() {
+        // Check if a version-specific nightly channel exists (e.g., 1.12-nightly)
+        let versioned_nightly = format!(
+            "{}.{}-nightly",
+            required_version.major, required_version.minor
+        );
```
MilesCranmer: Used a few times, I think it would be nice to have a 

```rust
fn versioned_nightly_channel(major: u64, minor: u64) -> String {
    format!("{}.{}-nightly", major, minor)
}
```

KristofferC:
> It is a runtime opt-in. 

Oh sorry, I thought we were talking about these type of features https://doc.rust-lang.org/cargo/reference/features.html

Edit: Added it

MilesCranmer:
(I did too, oops!)

src/version_selection.rs
```diff
@@ -0,0 +1,664 @@
...
+    Version::parse(base).with_context(|| format!("Failed to parse version `{}`.", base))
+}
+
+pub fn max_available_version(versions_db: &JuliaupVersionDB) -> Result<Option<Version>> {
+    Ok(versions_db
+        .available_versions
+        .keys()
+        .filter_map(|key| parse_db_version(key).ok())
+        .max())
+}
```
MilesCranmer: A bit too functional for rust imo, would be more idiomatic to write smth like

```suggestion
impl JuliaVersionDB {
    pub fn max_available_version(versions_db: &self) -> Result<Option<Version>> {
        Ok(versions_db
            .available_versions
            .keys()
            .filter_map(|key| parse_db_version(key).ok())
            .max())
    }
}
```

Then the call would be `versions_db.max_available_version()`

src/version_selection.rs
```diff
@@ -0,0 +1,664 @@
...
+    major: u64,
+    minor: u64,
+) -> Result<Option<Version>> {
+    Ok(versions_db
+        .available_channels
+        .keys()
+        .filter_map(|key| parse_db_version(key).ok())
+        .filter(|version| version.major == major && version.minor == minor)
+        .max())
+}
```
MilesCranmer: I think t his could also live in an `impl` block for JuliaupVersionDB, then you can do `versions_db.max_version_for_minor(major, minor)`. Again, doesn't really impact anything, it's just more rusty

src/version_selection.rs
MilesCranmer: Btw. Does this need to return a `Result`? Unless I am misreading, this will always return `Ok`?

src/version_selection.rs
KristofferC: The difference is that `eprintln!` is for users and will always show, `log::debug` is for juliaup developers and will not always show.

MilesCranmer: ah i see. ok good with me. in any case would be good to have the console::style setting match the rest of juliaup (if not already)?

IanButterworth: There's a function for these prints now

src/version_selection.rs
```diff
@@ -0,0 +1,664 @@
...
+) -> Result<Option<String>> {
+    if !manifest_version_detect {
+        return Ok(None);
+    }
+
+    if let Some(required_version) = determine_project_version_spec(args)? {
+        resolve_auto_channel(required_version, versions_db).map(Some)
+    } else {
+        Ok(None)
+    }
```
MilesCranmer, line 666:
```suggestion
    if !manifest_version_detect {
        Ok(None)
    } else if let Some(required_version) = determine_project_version_spec(args)? {
        resolve_auto_channel(required_version, versions_db).map(Some)
    } else {
        Ok(None)
    }
```

src/version_selection.rs
```diff
@@ -0,0 +1,664 @@
...
+
+    log::debug!("VersionDetect::No valid project with manifest found in JULIA_LOAD_PATH");
+    Ok(None)
+}
+
+/// Check if a Julia option requires an argument (mimics getopt's required_argument)
+/// Based on jloptions.c:421-493
+pub fn julia_option_requires_arg(opt: &str) -> bool {
+    // Options with = already include their argument
```
MilesCranmer, line 188:
```suggestion
    // TODO: Rewrite with clap.rs in future
    // Options with = already include their argument
```

src/version_selection.rs
```diff
@@ -0,0 +1,664 @@
...
+    Ok(Some(path))
+}
+
+/// Search JULIA_LOAD_PATH for the first valid project with a manifest
+pub fn find_project_from_load_path(
+    load_path: &str,
+    current_dir: &Path,
+    depot_path: Option<&std::ffi::OsStr>,
+) -> Result<Option<PathBuf>> {
+    let separator = if cfg!(windows) { ';' } else { ':' };
```
MilesCranmer: I'd prefer a single source of truth for this. c.f., the tests:

```rust
#[cfg(windows)]
const LOAD_PATH_SEPARATOR: &str = ";";
#[cfg(not(windows))]
const LOAD_PATH_SEPARATOR: &str = ":";
```

Can we put that test code into this src file, and then import it to the test

MilesCranmer:
Just to check, do u want me to click resolved on comments once you give a thumbs up, or will you click resolved after implementing?

KristofferC:
> 
Just to check, do u want me to click resolved on comments once you give a thumbs up, or will you click resolved after implementing?

Not sure.

-----

I made the "feature" opt-in for now since that seems to be what people want.

src/version_selection.rs
```diff
@@ -0,0 +1,666 @@
...
+
+    /// Find the maximum version (including prerelease) for a given major.minor series
+    pub fn max_version_for_minor(&self, major: u64, minor: u64) -> Option<Version> {
+        self.available_channels
+            .keys()
+            .filter_map(|key| parse_db_version(key).ok())
+            .filter(|version| version.major == major && version.minor == minor)
+            .max()
+    }
+}
```
MilesCranmer, line 531:
```suggestion
    pub fn has_channel(&self, version: &str) -> bool {
        self.available_channels.contains_key(version)
    }
}
```

src/version_selection.rs
```diff
@@ -0,0 +1,666 @@
...
+
+pub fn resolve_auto_channel(required: String, versions_db: &JuliaupVersionDB) -> Result<String> {
+    // Check if exact version is available
+    if versions_db.available_channels.contains_key(&required) {
```
MilesCranmer:
```suggestion
    if versions_db.has_channel(&required) {
```

src/version_selection.rs
```diff
@@ -0,0 +1,666 @@
...
+    // Prereleases should use nightly channels because they represent development/testing versions
+    if !required_version.pre.is_empty() {
+        // Check if a version-specific nightly channel exists (e.g., 1.12-nightly)
+        let versioned_nightly =
+            versioned_nightly_channel(required_version.major, required_version.minor);
+
+        if versions_db
+            .available_channels
+            .contains_key(&versioned_nightly)
+        {
```
MilesCranmer:
```suggestion
        if versions_db.has_channel(&versioned_nightly) {
```

src/version_selection.rs
```diff
@@ -0,0 +1,666 @@
...
+        if let Some(project_file) =
+            find_project_from_load_path(load_path, current_dir, depot_path.as_deref())?
+        {
+            return extract_version_from_project(project_file);
+        }
+    }
+
+    // No project found
+    log::debug!("VersionDetect::No project specification found");
+    Ok(None)
```
MilesCranmer:
```suggestion
    // Resolve project file (in priority order)
    let maybe_project_file =
        // 1. --project flag or JULIA_PROJECT env
        init_active_project_impl(
            args,
            current_dir,
            julia_project.as_deref(),
            depot_path.as_deref(),
        )?
        // 2. Fallback to JULIA_LOAD_PATH
        .or_else(|| {
            julia_load_path.as_ref().and_then(|load_path| {
                find_project_from_load_path(load_path, current_dir, depot_path.as_deref())
                    .ok()      // Result<Option<PathBuf>> → Option<Option<PathBuf>>
                    .flatten() // Option<Option<PathBuf>> → Option<PathBuf>
            })
        });
    
    // If no project was found, stop here
    let Some(project_file) = maybe_project_file else {
        log::debug!("VersionDetect::No project specification found");
        return Ok(None);
    };
    
    // Extract version from the resolved project
    extract_version_from_project(project_file)
```

Just made it a bit more readable to myself

src/version_selection.rs
```diff
@@ -0,0 +1,666 @@
...
+    load_path: &str,
+    current_dir: &Path,
+    depot_path: Option<&std::ffi::OsStr>,
+) -> Result<Option<PathBuf>> {
+    for entry in load_path.split(LOAD_PATH_SEPARATOR).map(str::trim) {
+        if should_skip_load_path_entry(entry) {
+            continue;
+        }
+
+        let Ok(Some(project_file)) = load_path_expand_impl(entry, current_dir, depot_path) else {
```
MilesCranmer: If `load_path_expand_impl` returns an `Err`, it gets skipped. Is that on purpose? Or would we want it to propagate normally?

src/version_selection.rs
```diff
@@ -0,0 +1,666 @@
...
+        "--version" | "--help" | "--help-hidden" | "--interactive" | "--quiet"
+            | "--experimental" | "--lisp" | "--image-codegen" | "--rr-detach"
+            | "--strip-metadata" | "--strip-ir" | "--gc-sweep-always-full"
+            | "--trace-compile-timing"
+            // optional_argument options
+            | "--project" | "--code-coverage" | "--track-allocation" | "--optimize"
+            | "--min-optlevel" | "--debug-info" | "--worker" | "--trim" | "--trace-eval"
+    )
+}
+
```
MilesCranmer, line 221:
```suggestion

const PROJECT_FLAGS: &[&str] = &["--project", "--projec", "--proje", "--proj"];

fn match_project_flag(arg: &str) -> Option<Option<String>> {
    PROJECT_FLAGS.iter().find_map(|flag| {
        if arg == *flag {
            // --proj / --proje / --projec / --project
            Some(None)
        } else {
            // --proj=val / --proje=val / --projec=val / --project=val
            arg.strip_prefix(flag)
                .and_then(|rest| rest.strip_prefix('='))
                .map(|v| Some(v.to_string()))
        }
    })
}

```

src/version_selection.rs
```diff
@@ -0,0 +1,666 @@
...
+            break;
+        }
+
+        if ["--project", "--projec", "--proje", "--proj"].contains(&arg.as_str()) {
+            project_cli = Some(None);
+        } else if let Some(value) = ["--project=", "--projec=", "--proje=", "--proj="]
+            .iter()
+            .find_map(|prefix| arg.strip_prefix(prefix))
+        {
+            project_cli = Some(Some(value.to_string()));
```
MilesCranmer:
```suggestion
        if let Some(spec) = match_project_flag(arg) {
            project_cli = Some(spec);
```

src/version_selection.rs
```diff
@@ -0,0 +1,666 @@
...
+        } else if julia_option_requires_arg(arg) {
+            // This option consumes the next token as its argument
+            // Skip it to avoid treating the argument as a flag
+            args_iter.next();
+        }
+    }
+
+    // Determine project spec
+    let project = if let Some(spec) = project_cli {
```
MilesCranmer:
```suggestion
    let maybe_project = if let Some(spec) = project_cli {
```

src/version_selection.rs
```diff
@@ -0,0 +1,666 @@
...
+        })
+    };
+
+    let project = match project {
+        None => return Ok(None),
+        Some(p) => p,
+    };
+
+    // Expand the project spec using load_path_expand
+    load_path_expand_impl(&project, current_dir, depot_path)
```
MilesCranmer:
```suggestion
    if let Some(project) = maybe_project {
        load_path_expand_impl(&project, current_dir, depot_path)
    } else {
        Ok(None)
    }
```

src/command_config_manifestversiondetect.rs
```diff
@@ -0,0 +1,64 @@
+use anyhow::Result;
+
+pub fn run_command_config_manifestversiondetect(
+    value: Option<bool>,
+    quiet: bool,
+    paths: &crate::global_paths::GlobalPaths,
+) -> Result<()> {
+    use crate::config_file::{load_config_db, load_mut_config_db, save_config_db};
+    use crate::utils::{print_juliaup_style, JuliaupMessageType};
+    use anyhow::Context;
```
MilesCranmer: move to top of file

src/command_config_manifestversiondetect.rs
```diff
@@ -0,0 +1,64 @@
+use anyhow::Result;
+
+pub fn run_command_config_manifestversiondetect(
+    value: Option<bool>,
+    quiet: bool,
+    paths: &crate::global_paths::GlobalPaths,
```
MilesCranmer: import normally

src/version_selection.rs
MilesCranmer, line 1: General question:

If Manifest.toml is invalid, do we want Juliaup to hard error, or should it just log it and fall back?

KristofferC: Imo, I think you never want it to hard error. It should print the error and fall back to `default`.

IanButterworth:
> I made the "feature" opt-in for now since that seems to be what people want.

You could add a prompt to enable the opt-in first time in interactive mode, like the auto channel install thing.

KristofferC:
Since this is now off by default, maybe it is safe to put this in and ask people to try things out?

KristofferC:
I think this is at the point where it would be good to merge, ask people to try it out and then it can be enabled by default.

IanButterworth (approved):
Plan to try this out sounds good to me

PR #1317, opened 2025-11-02
Close config files right after read

davidanthoff:
I think this probably should work because we now have a separate lock file that handles all concurrency.

WIP at the moment.


IanButterworth:
I think I closed this accidentally. Thought it was https://github.com/JuliaLang/juliaup/pull/746

Issue #1283, opened 2025-10-20
wrong update message?


xgdgsc:
<img width="869" height="532" alt="Image" src="https://github.com/user-attachments/assets/083168d3-9919-4381-a6e5-0d29c413493b" />

which suggests a downgrade. Only happened once. After it' s fine.

KristofferC:
I just got the same (but on linux)

davidanthoff:
If this happens again, could you run `juliaup info` and report the output here?

PR #1319, opened 2025-11-03
Only alert about actual new versions

IanButterworth:
Fixes https://github.com/JuliaLang/juliaup/issues/1283

I guess the db got out of sync with the installed channels, somehow.. This seems a logical check to add.

### DIFF
--- a/src/command_status.rs
+++ b/src/command_status.rs
@@ -13,6 +13,7 @@ use cli_table::{
 };
 use itertools::Itertools;
 use numeric_sort::cmp;
+use semver::Version;
 
 fn format_linked_command(command: &str, args: &Option<Vec<String>>) -> String {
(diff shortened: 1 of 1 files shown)


davidanthoff:
I think I would prefer that we just update the error message to be correct ("Something like `release is pointing to X, but you have Y`").

With this PR we would never be able to switch a channel back to an older version, and when I designed this originally, I wanted to have the ability to for example pull a prematurely released build.

Also, it would be good to actually understand what caused the underlying situation. I can't really wrap my head around what could have cause this, and I think I would prefer that we add diagnostic code that helps us track down what is going on with the DB versions if we encounter this situation. If we merge this PR we essentially will never be able to track down what is causing the problem in the first place..

Issue #1010, opened 2024-08-12
Use dmg on Mac


davidanthoff:
@IanButterworth I looked a bit into dmg files, here are some random thoughts:
- How would we extract the dmg file? All I find is instructions on "here is how you do that with finder", but that obviously doesn't help us...
- Related: do we understand how the attached ticket is handled? If (for example) the ticket is checked somehow by finder when one moves the content of the dmg file, then this won't work for us, right?
- Also, there seems to be yet another installer tech, namely installer with pkg files. Do we know what that is?

IanButterworth:
setup-Julia used to use dmgs and had all that figured out.

IanButterworth:
https://github.com/julia-actions/setup-julia/pull/117

davidanthoff:
https://github.com/julia-actions/setup-julia/blob/f2dfec4debe6a31c20070a66aa6d33580f91e8c8/src/installer.ts#L314 seems the relevant code.

So that is essentially just mounting the dmg file with `hdiutil` and then copying the content via `bash cp`. 

So what I am wondering is: at which stage it the attached ticket going to be "recorded" by the OS? During the mounting step? Or does the `cp` command that ships with MacOS handle these tickets? I think we first need to make sure that this approach actually gets the ticket to be recognized by the OS. I could also imagine a situation where the ticket only gets handled properly if one copies the stuff out of the dmg file via finder. I have no indication it is one or the other, so we just need to make sure to find out how it works.

davidanthoff:
https://forums.developer.apple.com/forums/thread/741219 suggests that in fact `hdiutil` will not ingest the ticket, but instead we need to open the image or something like that...

davidanthoff:
https://codermite.com/t/executing-shell-commands-like-open-hdiutil-xattr-from-a-swift-application-for-dmg-management-and-notarisation/211/2 is also relevant.

IanButterworth:
I think that will mean the mounted dmg will appear in finder. Hopefully we can unmount it too and delete the dmg

davidanthoff:
https://forums.developer.apple.com/forums/thread/706442 kind of makes it sound as if Gatekeeper only ever kicks in if the app has the quarantine attribute, which I would assume it doesn't have when installed via Juliaup. Or is that outdated info?

DilumAluthge:
Fixed by https://github.com/JuliaLang/juliaup/pull/1320?

PR #1320, opened 2025-11-03
Try to use .dmg installers on macOS before .tar.gz

IanButterworth:
## Summary

Use DMG installers instead of tarballs on macOS when available. This avoids the need for notarizing the standard library since DMG-distributed `.app` bundles come with notarization tickets intact.

Also strips the `com.apple.quarantine` extended attribute after extraction (both DMG and tarball) to prevent Gatekeeper/XProtect scanning delays during installation.

## Changes

- **DMG-first installation**: On macOS, attempt to download `.dmg` files first, falling back to `.tar.gz` if unavailable
- **App bundle support**: Handle `.app` bundle directory structure (`Contents/Resources/julia/bin/julia`) when resolving Julia binary paths via `resolve_julia_binary_path`
- **Quarantine stripping**: Remove `com.apple.quarantine` xattr after extraction to avoid Gatekeeper scanning delays that were causing macOS CI to take 20-45 minutes
- **RAII DMG handling**: `DmgMountGuard` ensures DMG volumes are always detached, even on error paths
- **Progress feedback**: Show status during DMG mount/copy/unmount phases ("Mounting installer...", "Copying application...", etc.)
- **Graceful fallback**: If DMG download or mount fails, automatically fall back to tarball extraction with existing notarization check
- **Pre-computed binary paths**: Store resolved binary paths in config (`BinaryPath` field, `skip_serializing_if = "Option::is_none"` for backward compatibility), with runtime fallback for existing installations
- **PR build codesigning**: Skip codesigning for DMG installs (already signed); gate on `is_pr && !used_dmg` in `install_from_url`
- **HTTP error handling**: Add status code checks and user-friendly hints for `download_juliaup_version` and `download_versiondb` failures
- **New tests**: `symlink_uses_app_bundle_for_system_channel`, `symlink_uses_app_bundle_for_direct_download_channel`, `dmg_failure_falls_back_to_tarball`

## CI performance

macOS CI time reduced significantly due to using dmgs and quarantine stripping:
- aarch64: 25m → 8m
- x86_64: 45m → 8m

## Example

```
% juliaup add release
  Installing Julia 1.12.1+0.aarch64.apple.darwin14
         Add Installed Julia channel 'release'
% julia +release -e "@time using Pkg"
  0.328271 seconds (517.75 k allocations: 34.026 MiB)
```

## Technical notes

- The current implementation uses runtime URL manipulation (`.tar.gz` → `.dmg`) as a workaround. TODO comments document the preferred approach: a v2 database schema with multiple ranked download sources per platform
- `install_from_url` and `install_non_db_version` now return `Result<(JuliaupConfigChannel, bool)>` where the bool indicates whether a DMG was used, enabling callers to skip codesigning/notarization for DMG installs
- The notarization check for tarball installations is preserved for Julia ≥1.11 when DMG is not available
- Note: this does not affect `setup-julia` CI action — that downloads tarballs directly and handles notarization separately ([context](https://github.com/julia-actions/setup-julia/issues/105#issuecomment-2283931792))

## Related

- Previous attempt: #998
- Julia issue: https://github.com/JuliaLang/julia/issues/54366
- Discussion: #1010

### DIFF
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -352,6 +352,14 @@ pub fn download_extract_dmg(url: &str, target_path: &Path) -> Result<String> {
 }
 
 #[cfg(target_os = "macos")]
+fn strip_quarantine_attribute(path: &Path) {
+    let _ = std::process::Command::new("xattr")
+        .args(["-dr", "com.apple.quarantine"])
+        .arg(path)
+        .output();
+}
+
+#[cfg(target_os = "macos")]
 fn try_download_dmg_with_fallback(url: &url::Url, target_path: &Path) -> Result<(String, bool)> {
     let dmg_url = if url.as_str().ends_with(".tar.gz") {
         url.as_str().replace(".tar.gz", ".dmg")
@@ -361,11 +369,13 @@ fn try_download_dmg_with_fallback(url: &url::Url, target_path: &Path) -> Result<
 
     if let Ok(dmg_url) = url::Url::parse(&dmg_url) {
         if let Ok(etag) = download_extract_dmg(dmg_url.as_ref(), target_path) {
+            strip_quarantine_attribute(target_path);
             return Ok((etag, true));
         }
     }
 
     let etag = download_extract_sans_parent(url.as_ref(), target_path, 1)?;
+    strip_quarantine_attribute(target_path);
     Ok((etag, false))
 }
 
@@ -969,7 +979,9 @@ pub fn install_from_url(
     #[cfg(target_os = "macos")]
     let julia_version = if did_codesign {
         let version = query_julia_version(&julia_path)?;
-        check_stdlib_notarization(&julia_path);
+        if !used_dmg {
+            check_stdlib_notarization(&julia_path);
+        }
         version
     } else {
         String::new()


DilumAluthge:
Might be worth sticking a link to [this comment](https://github.com/julia-actions/setup-julia/issues/105#issuecomment-2283931792) as a code comment somewhere here, otherwise people (like me) might forget why the `setup-julia` issue doesn't apply here.

src/operations.rs
```diff
@@ -148,6 +148,146 @@ fn bar_style() -> ProgressStyle {
...
+
+    eprint!("\r\x1b[2K");
+    let _ = std::io::stderr().flush();
+
+    Ok(etag)
+}
+
+#[cfg(target_os = "macos")]
+fn try_download_dmg_with_fallback(url: &url::Url, target_path: &Path) -> Result<(String, bool)> {
+    let dmg_url = if url.as_str().ends_with(".tar.gz") {
```
davidanthoff, line 364: We should really handle that in the construction of the version db, right? I.e. where there is a dmg file available, the `UrlPath` path in the version DB should just point to that.

IanButterworth: I thought about that but would that be backwards compatible? The db needs to serve before and after we add dmg support

davidanthoff: The version db file itself is versioned, so when we update the schema (which this would essentially be a case of) we would publish it as version db 2 file, so any old client would just stay on the old version file. Would actually be good to test such a version db schema change with a relatively simply case like this one.

IanButterworth: I spent some time trying to figure out the database update approach and don't think I understand how to add the v2 database and keep updating the v1 database.

I think implementing that is a bit outside of my wheelhouse...

Do you think we could proceed with this approach for now, with a TODO to update the database version format appropriately?

davidanthoff: I think my original thinking for DB versions was that we stop updating v1 database once we start to publish v2 databases. The general idea for Juliaup is that the binaries need to be kept up-to-date.

I would prefer that we try to do this with a DB update, I'm happy to help with that.

One thing I don't fully understand yet: do we need this stuff for _only_ the PR type channels, or for everything?

davidanthoff: Oh, and I can help with the DB update!

IanButterworth: Thanks. 

It's needed for all official releases. I'm not sure which of the nightlies or PRs have dmgs

IanButterworth: @davidanthoff to try and help this move along I've given the database format changes a go. What do you think?

IanButterworth: I think we need to keep moving forward here, so I'm going to implement the dmg switch via the workaround for now, and we can do the db format switch when you have time @davidanthoff

IanButterworth:
Note that this speeds up macOS CI to be in line with other platforms.

aarch - 19m -> 8m
x86 - 27m -> 8m

PR #1327, opened 2025-11-06
feat: persist `JULIA_*` ENV variables used as user settings

pankgeorg:
# Add Persistence for Julia Environment Variables

One of my biggest inconveniences with Julia is that you need to set the knobs very often in many places (`startup.jl` etc) 

This PR adds automatic persistence of all the Julia environment variables except the `JULIA_PROJECT` one. When Julia is launched with environment variables set (e.g., `JULIA_NUM_THREADS=8`), they are automatically saved to `~/.julia/juliaup/juliaup.json` and reused in future launches, eliminating the need to set them every time.

This is particularly important for me because I _always_ want _some_ knobs to be set; specifically:

- `JULIA_PKG_SERVER=juliahub.com`
- `JULIA_PKG_USE_CLI_GIT=true`
- `JULIA_NUM_THREADS=8`
- `JULIA_CPU_TARGET=generic` (yes, generic :) )

and so on

## Why not `.bashrc`?

Well, `.bashrc` is not cross platform and not standard. It's also significantly harder to edit `.bashrc` than running a julia command once to set the defaults, especially if you're trying to explain this to a new julia user.

## Why not `startup.jl`?

Some julia environment variables must be set before runtime (startup.jl is too late). This feature allows users to configure their Julia environment once without needing shell scripts or manual environment variable management.

## Example Usage

```bash
# Set environment variable once
JULIA_NUM_THREADS=8 julia -e 'println(Threads.nthreads())'  # Prints: 8


# Later, without setting the variable
julia -e 'println(Threads.nthreads())'  # Still prints: 8 (persisted value used)

# Override anytime
JULIA_NUM_THREADS=4 julia -e 'println(Threads.nthreads())'  # Prints: 4
```

### Changes

**`src/config_file.rs`**
- Added `julia_env_variables: HashMap<String, String>` to `JuliaupConfigSettings`
- Variables stored in JSON under `Settings.JuliaEnvironmentVariables`
- Empty maps skipped during serialization for clean configs

**`src/utils.rs`**
- Added `get_julia_environment_variables()` listing 58 supported variables
- Excludes `JULIA_PROJECT` per Julia documentation
- Covers: parallelization, file locations, Pkg.jl, REPL formatting, debugging, etc.

**`src/bin/julialauncher.rs`**
- Added `process_julia_environment_variables()` to handle persistence
- Loads persisted values from config
- Checks current environment for Julia variables
- Persists new/changed values automatically
- Applies resolved variables to Julia subprocess via `.envs()`

### Behavior

1. When Julia launches, the launcher checks for Julia environment variables in the current environment
2. If a variable has a non-empty value, it's both used immediately and persisted to config
3. If a variable isn't set but exists in config, the persisted value is applied
4. Current environment always takes precedence over persisted values
5. Persistence happens automatically on every Julia launch with environment variables

### Config File Format

Variables appear in the Settings section:

```json
{
  "Settings": {
    "VersionsDbUpdateInterval": 1440,
    "JuliaEnvironmentVariables": {
      "JULIA_NUM_THREADS": "8",
      "JULIA_EDITOR": "vim",
      "JULIA_PKG_SERVER": "https://pkg.julialang.org"
    }
  }
}
```

## Testing

Added comprehensive test suite:

**Unit Tests** (in `src/utils.rs`):
- Validates environment variable list completeness
- Ensures no duplicates
- Verifies `JULIA_PROJECT` exclusion

**Integration Tests** (new file `tests/env_var_persistence.rs`):
- Basic persistence and retrieval
- Precedence rules (current env > persisted)
- Multiple variables handling
- Empty value handling
- Value updates
- Config file integrity
- Cross-version persistence
- Preservation of other settings

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1102,9 +1102,9 @@ checksum = "469fb0b9cefa57e3ef31275ee7cacb78f2fdca44e4765491884a2b119d4eb130"
 
 [[package]]
 name = "iri-string"
-version = "0.7.8"
+version = "0.7.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dbc5ebe9c3a1a7a5127f920a418f7585e9e758e911d0466ed004f393b0e380b2"
+checksum = "4f867b9d1d896b67beb18518eda36fdb77a32ea590de864f1325b294a6d14397"
 dependencies = [
  "memchr",
  "serde",
@@ -1830,9 +1830,9 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.34"
+version = "0.23.35"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a9586e9ee2b4f8fab52a0048ca7334d7024eef48e2cb9407e3497bb7cab7fa7"
+checksum = "533f54bc6a7d4f647e46ad909549eda97bf5afc1585190ef692b4286b198bd8f"
 dependencies = [
  "once_cell",
  "ring",
@@ -2080,9 +2080,9 @@ checksum = "13c2bddecc57b384dee18652358fb23172facb8a2c51ccc10d74c157bdea3292"
 
 [[package]]
 name = "syn"
-version = "2.0.108"
+version = "2.0.109"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "da58917d35242480a05c2897064da0a80589a2a0476c9a3f2fdc83b53502e917"
+checksum = "2f17c7e013e88258aa9543dcbe81aca68a667a9ac37cd69c9fbc07858bfe0e2f"
 dependencies = [
  "proc-macro2",
  "quote",
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -11,6 +11,7 @@ use juliaup::jsonstructs_versionsdb::JuliaupVersionDB;
 use juliaup::operations::{is_pr_channel, is_valid_channel};
 use juliaup::utils::{print_juliaup_style, JuliaupMessageType};
 use juliaup::versions_file::load_versions_db;
+use std::collections::HashMap;
 #[cfg(not(windows))]
 use nix::{
     sys::wait::{waitpid, WaitStatus},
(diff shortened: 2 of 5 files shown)


davidanthoff:
Hm, ok, this is interesting, but I also think this is too magical and probably out of scope for Juliaup...

I think at the end of the day what this amounts to is a desire to have an option to configure the things that are currently configured via env variables via a global config file. To me that kind of makes sense. But I think really this should be done at the Julia level, i.e. if we want a global config file for these things, then Julia should add support for that.

I could see Juliaup playing in a role in such a setup, for example it could provide command line options to modify these global configuration options.

KristofferC:
> This PR adds automatic persistence of all the Julia environment variables except the JULIA_PROJECT one. When Julia is launched with environment variables set (e.g., JULIA_NUM_THREADS=8), they are automatically saved to ~/.julia/juliaup/juliaup.json and reused in future launches, eliminating the need to set them every time.

Maybe I missed something but just reading this in isolation it would be very surprising behavior to me. It seems like this is a job for https://direnv.net/ maybe?

pankgeorg:
> Hm, ok, this is interesting, but I also think this is too magical and probably out of scope for Juliaup...
> 
> I think at the end of the day what this amounts to is a desire to have an option to configure the things that are currently configured via env variables via a global config file. To me that kind of makes sense. But I think really this should be done at the Julia level, i.e. if we want a global config file for these things, then Julia should add support for that.
> 
> I could see Juliaup playing in a role in such a setup, for example it could provide command line options to modify these global configuration options.

I'm happy to discuss. I would personally prefer `juliaup` as it's more flexible and feels easier to change. My mental model is that `juliaup` could be more magical (towards the magic-ness of [uv](https://docs.astral.sh/uv/)  for example), but I understand if that is not really the goal. Having said that, uv doesn't do what this PR does. But I'll argue julia has more knobs to adjust, and that many of them (e.g. `JULIA_PKG_SERVER`) do have equivalents in [the settings](https://docs.astral.sh/uv/reference/settings/#index).

> Maybe I missed something but just reading this in isolation it would be very surprising behavior to me. It seems like this is a job for [direnv.net](https://direnv.net/) maybe?

Yea, `direnv.net` seems to address something similar. But I'm coming from a place where I'm trying to target users who do not want or just cannot refine their environments to such precision and detail. So I do think that adding support for "global" julia settings _somewhere_ is important. It needs to be something simple and something that loads _before_ julia does. Again, happy to discuss possible alternatives. 

Converting to draft for now, I'm open to ideas!

KristofferC:
So you want a way to persist these type of settings in some other way than using the environment? Having such an option is ok on its own, it is the automatic persistence of these by just setting options and launching julia that didn't really sit so well with me. Having `JULIA_NUM_THREADS=4 julia` change some global setting would be quite surprising.

pankgeorg:
> So you want a way to persist these type of settings in some other way than using the environment? Having such an option is ok on its own, it is the automatic persistence of these by just setting options and launching julia that didn't really sit so well with me. Having `JULIA_NUM_THREADS=4 julia` change some global setting would be quite surprising.

Yes, like `Preferences.toml` but for things that should be set _before_ julia starts up. Basically an alternative to the `~/.bashrc` suggestion [here](https://docs.julialang.org/en/v1/manual/environment-variables/)

davidanthoff:
Yeah, I could definitely imagine a world where there are lots of other things to configure with `juliaup configure`.

But also wondering whether `juliaup` is the right entry point, or whether something like `juliaconfig` would be better...

pankgeorg:
Will be superseded by https://github.com/JuliaLang/juliaup/compare/main...pankgeorg:juliaup:server-add?expand=1

PR #1332, opened 2025-11-14
add a pre commit file for running formatting and linting checks that would otherwise fail build

KristofferC:
https://pre-commit.com/

I keep forgetting to manually run e.g. `cargo fmt`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,27 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+  - repo: https://github.com/doublify/pre-commit-rust
+    rev: v1.0
+    hooks:
+      - id: fmt
+        name: cargo fmt
+        description: Format Rust code with rustfmt
+        entry: cargo fmt --all --
+        language: system
+        types: [rust]
+        pass_filenames: false
+      - id: cargo-check
+        name: cargo check
+        description: Check Rust code with cargo check
+        entry: cargo check
+        language: system
+        types: [rust]
+        pass_filenames: false
+      - id: clippy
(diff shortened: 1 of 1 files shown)


KristofferC:
I saw this

image


which also made me think this is a good idea 😆

IanButterworth:
Yeah it's currently v painful

.pre-commit-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+        language: system
+        types: [rust]
+        pass_filenames: false
+      - id: clippy
+        name: cargo clippy
+        description: Lint Rust code with clippy
+        entry: bash -c 'cargo clippy --workspace -- -D warnings'
+        language: system
+        types: [rust]
+        pass_filenames: false
```
IanButterworth, line 27: In CI we run clippy on the 3 platforms. Does this cover all 3?

KristofferC: Nope, and I don't think it is a really easy way to fix that. But it should catch most things and CI catches the rest.

SaadiSave:
@KristofferC  Won't this fail to check non-default features? Any errors behind a feature gate will simply slip through (until they show up in CI).

KristofferC:
Yes, this package unfortunately errors if you try to enable all features (https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features).

SaadiSave:
> Yes, this package unfortunately errors if you try to enable all features (https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features).

@KristofferC I don't mean mutually exclusive features. The precommit hook doesn't enable *any* features. So any mistakes gated behind `selfupdate`, for example, will only show up during CI and won't be caught before committing and pushing.

PR #1350, opened 2025-12-04
Fix FreeBSD CI timeout

### LINKED ISSUE #1349: Tests get stuck and time out on FreeBSD CI
It seems that FreeBSD CI freezes then the build is killed after about 2.5 hours. It doesn't appear to get stuck at the same place across runs. One common element is that the failing builds have some ssh info dumped into the logs, e.g. the line right before all of the debug info is dumped is
```
  client_loop: send disconnect: Broken pipe
```
and the last two lines of a build are
```
  ssh: connect to host 192.168.122.97 port 22: No route to host
Error: The process '/usr/bin/bash' failed with exit code 255
```
This makes me wonder whether the ssh connection between the host and VM is getting messed up somehow. Or it could be a red herring.

It might be a while before I'm able to test with my old x86-64 FreeBSD machine (local environment is AArch64 now) to see whether there's actually an issue with the tests/code rather than the setup.

ararslan:
Fixes #1349 by switching from `vmactions/freebsd-vm` to `cross-platform-actions/action`. Besides not exhibiting the freezing, the latter also seems a bit faster.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -214,16 +214,16 @@ jobs:
         CARGO_TARGET_aarch64-unknown-linux-musl: ${{matrix.rustflags}}
     - name: Test FreeBSD
       if: ${{ contains(matrix.target, 'freebsd') }}
-      uses: vmactions/freebsd-vm@v1
+      uses: cross-platform-actions/action@v0.30.0
       with:
-        release: "13.4"
-        usesh: true
-        mem: 8192
-        copyback: false
-        prepare: |
-          pkg install -y curl
-          curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --default-toolchain ${{matrix.toolchain}}
+        operating_system: freebsd
+        version: '13.4'
+        architecture: x86-64
+        cpu_count: 4
+        memory: '12G'
+        shell: sh
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -214,7 +214,7 @@ jobs:
         CARGO_TARGET_aarch64-unknown-linux-musl: ${{matrix.rustflags}}
     - name: Test FreeBSD
       if: ${{ contains(matrix.target, 'freebsd') }}
-      uses: vmactions/freebsd-vm@v1
+      uses: vmactions/freebsd-vm@2a40a6ac72877d88146d947622a7e006181b4787
```
IanButterworth: Can you add the version tag as a comment after it (Dependabot will update the comment in any updates)

ararslan: We don't want it to be updated, at least until whatever the issue is is fixed upstream. I can add the tag comment though

IanButterworth: Yeah. Any dependabot PRs to update it wouldn't pass CI so I don't worry about that.

ararslan: Done. (Also realized that this SHA wasn't even for a tag, ha.)

ararslan:
Okay, something is definitely weird here: a combination of changes that had been working yesterday no longer works and I can't see any reason why that should be the case.

PR #1355, opened 2025-12-12
Add User-Agent header to HTTP requests

lamdor:
Some Web Application Firewalls (like CloudFront used by JuliaHub) block requests without a User-Agent header, returning 403 Forbidden. This caused confusing "Failed to get etag" errors when using JULIAUP_SERVER with alternate mirrors.

Add a helper function http_client() that creates a reqwest client with a proper User-Agent header (juliaup/{version}), and use it for all HTTP requests on non-Windows platforms. Likewise for the windows HTTP client.


IanButterworth:
It seems this PR has "allow maintainers to modify" disabled, so I cannot rebase. If you can rebase/update shortly please do, otherwise I'll merge in 30-mins from now.

PR #1365, opened 2026-01-05
Document `juliaup config manifestversiondetect true`

visr:
Right now the readme reads like it is enabled by default.

Would be nice to have a release with https://github.com/JuliaLang/juliaup/pull/1315.

Or I'm also fine to not merge this and set the default to true, it works well for me and seems preferable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,12 @@ The channel is used in the order listed above, using the first available option.
 
 ### Project-based Version Selection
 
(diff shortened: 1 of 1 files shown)


IanButterworth (approved):

KristofferC:
I left it disabled by default as a compromise to get it merged quickly. I'm fine swapping it to default enabled but there is no real rush I guess.

PR #1390, opened 2026-01-19
Return valid file paths for alias channels in API

### LINKED ISSUE #1376: `juliaup api getconfig1` returns invalid file paths for `juliaup link` aliases
When creating alias channels the resulting output from `juliaup api getconfig1` returns the following:

https://github.com/JuliaLang/juliaup/blob/3ec01c8161ad3d11ebb222562a1034ff598eeb06/src/command_api.rs#L125

which are not files. This is currently resulting in breakage in https://github.com/julia-vscode/julia-vscode/issues/3986 which assumes that the `file` fields will be files.

Cc @pfitzseb

IanButterworth:
Fixes #1376

- Modified command_api.rs to resolve alias channels to their target
- Aliases now return actual executable paths instead of `alias-to-{target}`
- Properly accumulate and merge arguments from alias chains
- Added comprehensive tests in command_api_test.rs to verify:
  * Basic API functionality
  * Alias channels return valid file paths
  * Alias arguments are properly included
  * Aliases to system channels work correctly

Fixes issue where julia-vscode extension would fail when using juliaup link aliases because the File field contained invalid placeholder strings instead of actual executable paths.

Before:
```
  {"Name":"myalias","File":"alias-to-release","Version":"alias to release",...}
```
After:
```
  {"Name":"myalias","File":"/path/to/julia","Version":"1.10.10",...}
```

### DIFF
--- a/src/command_api.rs
+++ b/src/command_api.rs
@@ -47,24 +47,10 @@ pub fn run_command_api(command: &str, paths: &GlobalPaths) -> Result<()> {
         // Resolve aliases to their target channels
         let (resolved_value, alias_args) = match value {
             JuliaupConfigChannel::AliasChannel { target, args } => {
-                let mut current_target = target.as_str();
-                let mut accumulated_args = args.clone().unwrap_or_default();
-
-                // Follow alias chain to find the real channel
-                loop {
-                    match config_file.data.installed_channels.get(current_target) {
-                        Some(JuliaupConfigChannel::AliasChannel {
-                            target: next_target,
-                            args: next_args,
-                        }) => {
-                            if let Some(next_args) = next_args {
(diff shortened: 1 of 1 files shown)


IanButterworth:
@MichaelHatherly @pfitzseb does this look correct?

pfitzseb:
Yes, I think so. This has some currently unnecessary complexity though, because you can't currently alias an alias as per
https://github.com/JuliaLang/juliaup/blob/9e3f5e8df488e58a83e38769bca01ec26d9e7d2b/src/command_link.rs#L36-L41

IanButterworth:
Ah right. I forgot that we removed that before merging the alias support. Thanks

MichaelHatherly:
Thanks @IanButterworth.

PR #1410, opened 2026-02-05
make manifestversiondetect config default to true

KristofferC:
This is working quite nicely for me at least.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,11 +165,11 @@ The channel is used in the order listed above, using the first available option.
 
 ### Project-based Version Selection
 
-When no explicit channel is specified via command line, environment variable, or directory override, Juliaup can automatically attempt to select an appropriate Julia version based on the active project's requirements.
-This feature is currently disabled by default, but will likely be enabled by default in the future. Options are `true`, `false`, `default`. Set it with:
+When no explicit channel is specified via command line, environment variable, or directory override, Juliaup automatically attempts to select an appropriate Julia version based on the active project's requirements.
+This feature can be configured by setting the `manifestversiondetect` config option to `true`, `false` or `default` as:
 
 ```sh
-juliaup config manifestversiondetect true
+juliaup config manifestversiondetect false
 ```
 
 If a project is specified (via `--project`, `JULIA_PROJECT`, or `JULIA_LOAD_PATH`), Juliaup reads the project's `Manifest.toml` and uses the `julia_version` field to determine which Julia version to use:
--- a/src/config_file.rs
+++ b/src/config_file.rs
@@ -96,7 +96,7 @@ impl Default for JuliaupConfigSettings {
             create_channel_symlinks: false,
             versionsdb_update_interval: default_versionsdb_update_interval(),
             auto_install_channels: None,
-            manifest_version_detect: false,
+            manifest_version_detect: true,
         }
     }
 }


visr:
I support defaulting to true after having used it for a while.
However when a new patch release comes out, I don't like that I have to override this for each project to avoid staying on older patch releases.
Would it be an idea to start the latest patch release that is installed, as long as it isn't older than the version in the manifest? There are cases where you don't want this, but I suspect they are far fewer.

IanButterworth:
I think we might want to notify of this behavior change when juliaup updates itself, which is currently silent.

IanButterworth:
Proposal for that notification system here https://github.com/JuliaLang/juliaup/pull/1411

KristofferC:
> Would it be an idea to start the latest patch release that is installed, as long as it isn't older than the version in the manifest?

I'm not sure I would want that. You need to recompile your environment, etc. Part of the niceness of this is, imo, that things kind of keep working the way they did. And if you want to force a specific julia version, you can use `JULIAUP_CHANNEL` maybe?

visr:
>  things kind of keep working the way they did

From the [priority list](https://github.com/JuliaLang/juliaup?tab=readme-ov-file#overrides) I think people mostly rely on the last one, the default channel. This comes above the default channel, so it does change things. The default channel tends to move with the patch releases and not keep older patch releases installed. 

I expect people will start running into this a lot:

```
❯ julia
Question: The Juliaup channel '1.12.3' is not installed. Would you like to install it?: No
ERROR: `1.12.3` resolved from project manifest is not installed. Please run `juliaup add 1.12.3` to install channel or version.
```

So we'd suggest them to install and hence keep using an older patch release, even though in practice we'd recommend `julia +1.12 --project -e 'using Pkg; Pkg.resolve()'`, after which it will select the latest 1.12 again.

I appreciate that the current approach is simpler, but I'm afraid it will cause people to run behind on julia patches, and/or use different patch releases at the same time without intending to. But perhaps that is the difficulty of equating the julia version the manifest last got resolved with, with the version that we want people to use by default.

davidanthoff:
How about this: when Juliaup uses a Project/Manifest for Julia selection, it also reads the `[compat]` section for `julia` from the `Project.toml`. It then checks whether there is a newer Julia version that is compatible via our normal semver rules. If that is the case, it prompts the user with something like:
```
This project is currently using Julia 1.12.1. Do you want to update the project to use Julia 1.12.5?
- Yes
- No
- No, and fix the project to Julia 1.12.1
```
If the user picks `yes`, Juliaup launches the correct Julia version, and triggers a project resolve, thus updating the manifest.
If the user picks `no`, we just launch the old Julia version
If the user picks `No, and...`, then we update the `Project.toml` `[compat]` section to read `julia="=1.12.1"`, which should then prevent future update nags.

If the update would amount to a minor version jump, we could also offer options that fix the compat to minor version upgrades etc.

davidanthoff:
In terms of changing the default, we should wait a little bit. I believe we never shipped a version of Juliaup to any Windows user that had the feature at all (because we temporarily lost access to the Windows Store listing...), so ideally we would test it for a while with Windows users before we switch things over.

davidanthoff:
We probably also need to add better support in the VS Code extension before we ship that, right now I got some very cryptic crashes with this option enabled.

PR #1420, opened 2026-02-10
Add tab completion for julia +channel selection

IanButterworth:
Add shell tab completion so users can type 'julia +<TAB>' and see available channels. Completions for both juliaup commands and julia channel selection are auto-installed for bash and zsh, and available for all 6 supported shells (bash, zsh, fish, powershell, elvish, nushell).

```
ian@Mac julia % julia +
1.10          1.11-nightly  1.11.6        1.12-nightly  1.13          1.7           1.9           n             pr60943
1.11          1.11.5        1.12          1.12.4        1.13-nightly  1.8           lts           nightly       release
```
and `julia +pr{tab}` etc.

Changes:
- Add hidden '_list-channels' subcommand that prints sorted installed channel names, used by completion scripts
- Generate per-shell completion scripts that combine juliaup subcommand completions with julia +channel completions
- Write pre-generated completion files to ~/.juliaup/completions/ during initial setup and self-update to avoid eval overhead on shell startup
- Add 'source' line to managed shell init blocks in .bashrc/.zshrc
- Add hidden '_post-update' subcommand invoked by the old binary after extracting a new one, to refresh completion files and shell init blocks
- Update README to document auto-installed completions

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -235,7 +235,7 @@ fn set_auto_install_preference(
 
     config_file.data.settings.auto_install_channels = Some(auto_install);
 
-    save_config_db(&mut config_file)
+    save_config_db(&mut config_file, paths)
         .with_context(|| "Failed to save auto-install preference to configuration.")?;
 
     print_juliaup_style(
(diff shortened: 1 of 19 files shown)


fingolfin:
Off-topic, but: Can you perhaps tell Claude to put in the notifications about it being used in a way that is not hidden by GitHub? Because `Co-authored-by:` lines are removed from the automatically generated PR descriptions (for my own work, I manually add them back each time I open a PR).

PR #1426, opened 2026-02-19
 ci: add submission monitor to msstore release

pfitzseb:
We're trying to debug why the release to the msstore fails. 
- first commit in this PR adds a `SubmissionMonitor` for the store submission, which seems like a good idea regardless (and will hopefully show what the issue is).
- second commit adds a new deployment action to exfiltrate the `CLIENTID` and `TENANTID` secrets as base64 encoded strings, because we aren't 100% sure which account is being used for this

cc @aviks

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1064,7 +1064,8 @@ jobs:
         $sec = ConvertTo-SecureString $env:CLIENTSECRET -AsPlainText -Force
         $cred = New-Object System.Management.Automation.PSCredential $env:CLIENTID, $sec
         Set-StoreBrokerAuthentication -TenantId $env:TENANTID -Credential $cred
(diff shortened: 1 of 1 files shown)


pfitzseb:
@IanButterworth wanna approve that deployment? Probably best if it gets double checked etc...

IanButterworth:
Approved

pfitzseb:
```
clientid: 05281901-f0b2-4268-8392-9fbdff15d4b7
tenantid: 8d3de454-9fdc-4091-90af-9cf79dcabe79
```

pfitzseb:
Thanks!

This now only contains the first commit which adds a submission monitor. According to the docs it should work fine, but I don't think there's any way I can reasonably test that.

IanButterworth:
Should I go ahead with a new patch release? I guess we won't know until it's fully released?

pfitzseb:
Yep, that would help.

PR #1434, opened 2026-02-21
Fail winstore CD jobs on CommitFailed submission status

IanButterworth:
The winstore release job is green although it's failing

---

Start-SubmissionMonitor prints CommitFailed but does not throw, so the job exits 0. After the monitor returns, check the submission status and throw if it is CommitFailed.

Also add monitoring to the dev and release-preview flight deployments, which previously did not wait for or check submission results.


fingolfin:
I have no idea what this code does, but I am going to trust @IanButterworth and @pfitzseb on this one ;-)

PR #1439, opened 2026-03-01
Add juliaupgui: graphical interface for juliaup

### LINKED ISSUE #54: Add a GUI application for Julia version management
We could also add a GUI application named `Juliaup` or `Julia Configuration` or something like that. This would appear in the Start menu and would show a list of all channels on the system, which ones are out-of-date, with a button to update individual or all channels.

IanButterworth:
Closes https://github.com/JuliaLang/juliaup/issues/54
Some discussion in https://github.com/JuliaLang/www.julialang.org/pull/2430

Developed with Claude.

New juliaupgui workspace crate (eframe/egui 0.29) providing a native GUI
for managing Julia installations, accessible via \`juliaup gui\`.

Installed tab:
  - Tile and list views with launch, update, set-default, remove actions
  - Whole-tile click to launch; hover highlights and pointer cursor
  - Custom launch popup with project directory, extra args, and env vars
  - Per-session persistence of project dir, args, and env vars

Available tab:
  - Filterable channel list with install buttons and PR-number prompts
  - Nested channel tree with expand/collapse; popular channels highlighted
  - Link existing Julia binaries to custom channel names

Configuration tab:
  - All juliaup settings: DB update interval, auto-install, manifest detect,
    channel symlinks
  - Directory overrides management (add/remove/prune non-existent)
  - Terminal application preference with file browser (rfd)
  - Self-update, refresh DB, and garbage collect actions

Platform integration:
  - macOS: .app bundle in /Applications with generated .icns icon
  - Linux: .desktop file and PNG icon in XDG directories
  - Shortcuts created on install, refreshed on self-update, removed on uninstall
  - Launch opens new terminal window (Terminal.app, iTerm2, or configured app)
  -  Julia three-dot logo (SVG-accurate) as window icon and app icon

Infrastructure:
  - Background worker thread for non-blocking operations
  - Subprocess streaming with ANSI stripping for output-producing ops
  - Activity log panel with collapsible history

---
Fresh install

Screenshot 2026-03-02 at 9 48 18 PM

---

Screenshot 2026-03-02 at 8 47 59 PM
Screenshot 2026-03-02 at 8 48 07 PM
Screenshot 2026-03-02 at 8 48 14 PM

Screenshot 2026-03-02 at 9 17 18 PM
Screenshot 2026-03-02 at 8 48 26 PM


ViralBShah:
Would it make sense for the UI to present a multi-level view? Show the top level releases - 1.10, 1.11, nightly etc. And then click on that to get in to see all the different RCs, patch versions, etc.

I think it would be valuable for the first screen to be newbie friendly that basically just installs/updates the latest stable version and maybe also offers the LTS version. You then click on something called "Channels" to go into all the other options.

ViralBShah:
The JuliaUp version number next to the title (1.19.9) has the risk of being confused with the Julia version number. Perhaps you should only get it through an About section.

ViralBShah:
Completely random and off-topic here - but it would be great to have the same UI manage packages and environments as well. So that you have one UI for installing and managing everything.

Would it make sense for this to be a separate repo?

davidanthoff:
This is very cool!!

> Completely random and off-topic here - but it would be great to have the same UI manage packages and environments as well. So that you have one UI for installing and managing everything.

Agreed! And apps. I generally think we should integrate apps more into the Juliaup story. I'm thinking of apps as something global per system installed, and I think centralizing those kind of things under Juliaup would make sense.

Random other note: it would _also_ be nice to have a UI like this in the VS Code extension. So maybe another idea would be to try to implement the UI in some kind of web framework, then expose it as a native client via tauri, and in the VS Code extension. Juliaup itself could just expose a JON-RCP interface that the UI code call into to do actual data operations. Main drawback of this though is that it is a heck of a lot more indirections and complicated... So I could also see an argument to just go with a simple version initially a la this PR, and then later make it more sophisticated...

StefanKarpinski:
This looks great. Vibe coding is amazing...

IanButterworth:
I think the first tab should default to a big friendly tile view, then this view be the "list" view alternative. That should work better for more basic usage.

IanButterworth:
Added a more basic default tile view and nesting in the channel list. Updated the screenshots up top

IanButterworth:
Can I request a few people try this out here. If it's looking good I'll focus on code polish before asking for review. (Claude thinks its code is good at least, with only minor findings...).

Note I updated the top post with a description of all the changes.

IanButterworth:
I've addressed review findings from Claude Opus 4.6 and Codex 5.3.

I'm planning to go ahead and merge this and put it out on releasepreview for people to try out.

davidanthoff:
Ok, just took a very quick look at the code structure. Why are we introducing this whole `app_shortcut` stuff? Isn't this just a new binary, and we should handle it exactly like all the other binaries, i.e. `julia` and `juliaup`?

IanButterworth:
This installs juliaupgui as an application on the system i.e. on macOS it creates an .app and installs it into /Applications so that the user can double-click on that to run it. I've thought less about how that works on linux/windows, but it should be done if claude is to be believed.

davidanthoff:
Hm, so at least for Windows we definitely want to handle this differently.

I'll take another look at MacOS.

On Linux, I have no idea whether there is something like "shortcuts" for programs that show up in some launch menu? But for invocation from the command line, it seems it should be handled exactly like `juliaup` etc, right?

And then in general we'll need to change Github workflow that builds everything, I don't think it will be bundled at all otherwise.

davidanthoff:
Just started to run it on Windows. The default screen I get is:

image

That color scheme is not ideal, hard to read things.

IanButterworth:
Can you try on darkmode. I'll fix light mode (I hadn't tested that)

IanButterworth:
Also I'll make it default to list view when there are many channels installed.

davidanthoff:
Dark mode looks good:

image

davidanthoff:
Might also make sense to add a config for that? Right now it picks up the Windows system setting, which seems the right thing as a default, but might be nice to have an override option.

davidanthoff:
I added the DO NOT MERGE label, not because I don't like this PR, but because @IanButterworth and I discussed a sequencing for this PR that will have it wait a bit until we have the Windows Store submission working again. So this is just informational for everyone else to please not merge at this point :)

ViralBShah:
Is the Windows store still a blocker? Just wondering if we have a GUI, then can't it download the MSI installer or even the portable zip installer directly?

Lots of people who want to use Julia in companies seem to have issues with the store in general.

davidanthoff:
This is really orthogonal to the question of how one might get Julia/Juliaup onto the user machine in the first place. What this PR does is add another executable to the existing installer options that one can then launch and use to manage Julia versions. But all of that only kicks in _after_ things have been installed successfully.

@ViralBShah I think our best answer for situations where the Julia store doesn't work is still 1) app installer and then 2) the MSI installer. I have not heard any complaints that app installer in particular is blocked anywhere, so that definitely seems like the best second option.

I am going to look again whether we can actually somehow package _all_ these options behind _one_ `winget` command that then just auto-selects the method that works. Might be possible now.

IanButterworth:
@davidanthoff lets get this out so people can try it out during JuliaCon

src/command_app_shortcut.rs
```diff
@@ -0,0 +1,398 @@
...
+        .ok_or_else(|| anyhow::anyhow!("Could not determine XDG data directory"))?;
+    Ok(data_dir
+        .join("icons")
+        .join("hicolor")
+        .join("256x256")
+        .join("apps"))
+}
+
+#[cfg(target_os = "linux")]
+fn create_linux_desktop(gui_bin: &std::path::Path) -> Result<()> {
```
davidanthoff: Does this create an icon on the desktop? That seems weird, I'm not a Linux expert, but at least on Windows I'm not a fan of apps that just put themselves there :) Wouldn't we want to appear in some kind of start menu?

src/operations.rs
```diff
@@ -1221,10 +1221,21 @@ fn pr_info_from_api_response(body: &str) -> Result<PrInfo> {
     })
 }
 
-/// Resolves a Julia pull request number to the current head commit sha of the
-/// PR branch and its lifecycle state via the GitHub API.
+fn pr_title_from_api_response(body: &str) -> Result<String> {
```
davidanthoff, line 1224: What is this doing here? Is this from some other PR?

src/operations.rs
```diff
@@ -2830,20 +2851,44 @@ fn download_direct_download_etags(
...
+
+            let mut input = String::new();
+            if io::stdin().read_line(&mut input)? == 0 {
+                bail!(
+                    "Cannot prompt for PR build codesigning because standard input is unavailable."
+                );
+            }
+            !matches!(input.trim().to_ascii_lowercase().as_str(), "n" | "no")
+        }
+    };
```
davidanthoff, line 2888: Same here, this all seems unrelated to a GUI?

davidanthoff (commented):
Ok, took a quick look. 

There seems to be some code that is unrelated to the GUI stuff?

I think my biggest worry right now is all the setup/install stuff. I don't have time to review that, but in my mind that touches _a lot_ of critical code paths, is super platform specific and makes me quite nervous. I think ideally we would devise some rollout strategy that gives us a lot of time to test that, make sure it all works etc. My gut feeling is that I wouldn't ship something like that a day before Juliacon.

Is there a world where we merge this PR without any of the app installation stuff? So for now the entry point is just `juliaup gui`? That would allow folks to play with it, we can iterate on it, but shipping all the code that does the installation stuff we could ship later.

IanButterworth:
Yeah that's what I was thinking tbh. I can reel back the installation.

src/command_app_shortcut.rs
IanButterworth: No, it wrote `~/.local/share/applications/juliaup.desktop`, which is the freedesktop *desktop entry*, i.e. the application menu entry that is the Linux equivalent of a Start Menu shortcut. Nothing ever went to `~/Desktop`.
The naming was confusing: `create_linux_desktop` meant "desktop entry", but moot now anyway, since I've removed the whole app installation layer from this PR. When it comes back we can use clearer names (`install_linux_desktop_entry`, `desktop_entry_path`, `icon_theme_dir`).

src/operations.rs
IanButterworth: This one is for the GUI. It splits the GitHub PR API call out into `github_pr_api_response` so the response can be reused, then adds `get_julia_pr_title` on top of it. The GUI calls that at `juliaupgui/src/app.rs:312` to show the PR title next to a PR channel instead of just the bare number.

src/operations.rs
IanButterworth: Also GUI driven. `prompt_and_codesign_pr_build` blocks on stdin, which the GUI has no way to answer, so it would just hang on the prompt. `JULIAUP_PR_CODESIGN` gives a non-interactive path, set by the GUI at `juliaupgui/src/app.rs:2758-2760`. The `read_line == 0` check is part of the same fix, since an EOF on stdin was previously read as "yes".

IanButterworth:
Ok, I've narrowed the scope now so people can just `juliaupgui` or `juliaup gui` to get it.

IanButterworth:
Oh and I propose we only push to `dev` during juliacon so people can opt-in if they want to try it.

davidanthoff (approved):
Let's merge it!

davidanthoff:
> Oh and I propose we only push to dev during juliacon so people can opt-in if they want to try it.

In my mind, if it is on `main` we might as well ship it on the prerelease channel? But maybe what we could do is tag a version, and release it in the dev channel. That should ship the build to my machine (I'm the only one on Windows on the dev channel), and then I can test that, and if that works we can push it to the prerelease channel?

.github/workflows/release.yml
```diff
@@ -253,6 +253,13 @@ jobs:
         CARGO_TARGET_i686-unknown-linux-musl: ${{matrix.rustflags}}
         CARGO_TARGET_aarch64-unknown-linux-musl: ${{matrix.rustflags}}
         AWS_LC_SYS_CMAKE_BUILDER: "1"
+    - name: Build juliaupgui
+      if: ${{ ! contains(matrix.target, 'freebsd') && ! contains(matrix.label, 'portable') && ! contains(matrix.target, 'musl') }}
```
ararslan, line 257: Why exclude FreeBSD? I haven't tried building the GUI locally but reading through the code it looks like it should work.

IanButterworth: Just no testing. If you can try that would be great!

IanButterworth: https://github.com/JuliaLang/juliaup/pull/1569

PR #1479, opened 2026-04-16
Switch tests from 1.11-nightly to 1.10-nightly

fingolfin:
It seems that there are no more 1.11-nightly builds, causing CI failures for juliaup.

### DIFF
--- a/tests/command_add.rs
+++ b/tests/command_add.rs
@@ -23,7 +23,7 @@ fn command_add() {
 
     env.juliaup()
         .arg("add")
-        .arg("1.11-nightly")
+        .arg("1.10-nightly")
         .assert()
         .success()
         .stdout("");


tests/command_add.rs
```diff
@@ -23,7 +23,7 @@ fn command_add() {
 
     env.juliaup()
         .arg("add")
-        .arg("1.11-nightly")
+        .arg("1.10-nightly")
```
IanButterworth, line 26: Buy us some more time?
```suggestion
        .arg("1.13-nightly")
```

fingolfin: My reasoning was that 1.10 is the LTS release, so potentially longer-lived than 1.13. Of course it needs to be updated again when the next LTS release is designated, but that happens less often than new releases are made / become unsupported...

Is there a way at that point to programatically query which version is the LTS? And/or the latest regular release? Then one could make this test self-updating...

fingolfin:
Anyway, if you really prefer to have 1.13, feel free to just edit this PR (you should be able to modify the branch or edit via code change comments, right?) -- either way it would be good to get this merged ASAP so that CI becomes useful again for other PRs :-)

tests/command_add.rs
IanButterworth: Good point. Let's just keep this

IanButterworth (approved):

PR #1500, opened 2026-05-15
CI: disable fail fast

IanButterworth:
Our CI is too flaky. It makes getting CI green annoying

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ on:
 
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
-  cancel-in-progress: false
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 jobs:
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,14 +9,14 @@ on:
 
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+  cancel-in-progress: false
 
 jobs:
 
   check-juliaup:
     runs-on: ${{ matrix.runner }}
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -9,14 +9,14 @@ on:
...
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+  cancel-in-progress: false
 
 jobs:
 
   check-juliaup:
     runs-on: ${{ matrix.runner }}
     strategy:
-      fail-fast: ${{ github.ref != 'refs/heads/main' }}
+      fail-fast: false
```
davidanthoff, line 19: This is just the compiler/linter, right? Shouldn't that stay on, there shouldn't be any false positives here, I think?

.github/workflows/test.yml
```diff
@@ -9,14 +9,14 @@ on:
 
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
-  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+  cancel-in-progress: false
```
davidanthoff: I do think we should keep some kind of concurrency cancelation, otherwise we'll completely saturate the runners. In what way would this here help with flakey tests?

davidanthoff (changes requested):
One of the changes makes a ton of sense to me, the other two less :)

IanButterworth:
On CI I find it more helpful to collect all issues to work on locally. I don't see the merit of failing fast given the amount CI is run on this repo isn't that excessive.

davidanthoff:
I can certainly be convinced that the `fail-fast` on the `cargo check` is not needed, but I still don't understand why we wouldn't want concurrency cancellation?

.github/workflows/test.yml
IanButterworth: yep, accidental

PR #1514, opened 2026-06-05
Restore julia symlinks after self-update

IanButterworth:
Self-update replaces the entire bin directory via an atomic rename, which removes the `julia` -> `julialauncher` symlink (and any per-channel `julia-<channel>` symlinks) that are not part of the juliaup tarball. This left users with "command not found: julia" after an automatic update.

Recreate these symlinks in the `_post-update` hook so `julia` keeps working after an update. Best-effort and guarded for non-Windows targets.

Fixes #1513

PR #1515, opened 2026-06-05
Release config lock before spawning post-update hook

IanButterworth:
`self update` holds the exclusive configuration lock from `load_mut_config_db` for the whole command. The `_post-update` hook runs in a child process that reads the config (to restore channel symlinks) via a shared lock, which blocked forever on the parent's exclusive lock.

Drop the lock after extraction completes, just before spawning the hook, so the child can acquire its read lock. The lock is intentionally held across the extraction to serialize concurrent self-updates.

Fixes the deadlock reported after #1514 on the releasepreview channel.


IanButterworth:
This would've been caught in tests if we had a self update integration test. Currently tests check the individual steps only.

PR #1517, opened 2026-06-06
Don't hold the configuration lock during downloads

### LINKED ISSUE #435: Unlock configuration file during download
When I run `juliaup add 1.7` it takes a long time to download because of a slow internet connection. This is okay, but when I try to launch `julia +1.8.2`  in a different terminal, I get "Juliaup configuration is locked by another process, waiting for it to unlock." It would be nice to unlock the configuration during the download process and only lock it very briefly when doing actual work with the configuration.

### LINKED ISSUE #561: Don't lock the juliaup config file when only reading

### LINKED ISSUE #1524: potential spurious lock messages in 1.20.4
Since a few days (so I am assuming this is coming from upgrading to 1.20.4) I have been getting 
```
Juliaup configuration is locked by another process, waiting for it to unlock.
```
messages. At this point the process stalls and has to be interrupted.

I am observing this when running a script that runs julia twice, as in

```sh
#!/bin/bash
VERSION=1.11.9
julia +"$VERSION" --project=. -e 'import Pkg; Pkg.instantiate()'
julia +"$VERSION" -p 6 --project=. run_estimation.jl
```

It is during the second process.

This is on a server where no other julia/juliaup process can be running, so I am assuming this is spurious. The only lock-related change I am seeing is in #1515, but I am not claiming this is related. If I can help debug this further, let me know what to do.

IanButterworth:
Closes #435
Closes #561
Closes #1524

Existing draft PRs
Closes #1238
Closes #1317
Closes #746

Claude:

---

Shrink the region covered by the global juliaup configuration lock so the network download/extract of a version no longer happens while the exclusive lock is held. The same pattern is now applied to every command that downloads:

- **`add`** — splits the old `install_version` into `download_version_to_temp` (lock-free, network-bound) and `commit_version_install` (lock-held, fast rename + config write).
- **`self update`** — reads the configured channel under a brief shared lock, releases it, downloads juliaup with no lock held, then re-acquires the exclusive lock only to record the timestamp.
- **`update`** — two-phase: snapshot the config under a shared lock, download all needed channels lock-free (`prepare_channel_update`), then take the exclusive lock once to commit them all (`commit_channel_update`).
- `update_version_db` already followed this pattern.

In every case the commit phase re-checks the config (optimistic concurrency), so two processes acting on the same version/channel resolve cleanly (the loser discards its temp dir) rather than erroring.

Also:
- The "configuration is locked by another process, waiting for it to unlock" message now only prints if the lock can't be acquired within a 1s grace period, suppressing spurious messages during the millisecond-scale commit phase.
- Adds https://github.com/JuliaLang/juliaup/blob/ib%2Freduce_lock/devdocs/locking_and_update_flows.md documenting the lock model and per-command flows with diagrams.

Together these eliminate the long-lock-across-network pathology in the launcher's background daemon path (it spawns `self update` / versiondb update), which was the cause of the spurious lock stalls in #1524.

How this honors the discussion (#435, #561):
- Implements davidanthoff's proposed design from #435: read config + release lock, download, then re-acquire and commit, instead of holding the write lock for the whole download.
- Honors LilithHafner's refinement: the commit phase re-checks whether the install is still needed.
- Keeps StefanKarpinski's shared/exclusive model: many readers coexist and only block for the millisecond-scale commit, not for a multi-minute download.

Benefit over the other open PRs:
- vs #1238 (remove file locking entirely): juliaup performs multi-file operations (install dir + config + symlinks) that a single-file atomic rename cannot keep consistent. This retains the lock for cross-file consistency while still eliminating the long-lock pathology #1238 targets.
- vs #1317 (close config files right after read): complementary — that PR shortens how long handles stay open; this shortens how long the exclusive lock is held during the download.

### DIFF
--- a/devdocs/locking_and_update_flows.md
+++ b/devdocs/locking_and_update_flows.md
@@ -1,0 +1,217 @@
+# Configuration locking & update flows
+
+This document describes how juliaup serializes access to its configuration
+(`juliaup.json`) and where the configuration lock is held during the various
+launcher and update code paths. It exists primarily to make the locking
+behaviour auditable, since holding the lock across network operations has
+historically caused spurious "configuration is locked by another process"
+stalls (see [#1524](https://github.com/JuliaLang/juliaup/issues/1524)).
+
+## The lock
+
+All access to the configuration is serialized through a single lock file
+(`paths.lockfile`) using advisory file locks (`cluFlock`). There are two lock
+modes:
+
+| Mode | Acquired by | Blocks |
+| --- | --- | --- |
+| **Shared** (read) | `get_read_lock` / `load_config_db(paths, None)` | only an exclusive holder |
+| **Exclusive** (write) | `load_mut_config_db` | any other shared or exclusive holder |
+
+Many readers can hold the shared lock simultaneously. A single writer holding
+the exclusive lock blocks everyone, including readers. The user-visible message
+
+> Juliaup configuration is locked by another process, waiting for it to unlock.
+
+is printed whenever a process cannot acquire the lock it wants within a short
+grace period (1 second) and falls back to a blocking wait. The acquisition is
+retried (polling) during the grace period, so the message is suppressed for the
+common case where another process holds the lock for only a few milliseconds
+(e.g. while committing a config change). Both `get_read_lock` and
+`load_mut_config_db` route through `lock_with_delayed_message` to get this
+behaviour.
+
+### Design rule
+
+> **Never hold the exclusive lock across a network operation.**
+
+Downloads can be slow or hang. A writer that holds the exclusive lock across a
+download blocks every concurrent `julia` / `juliaup` invocation (each of which
+needs at least a shared lock at startup) for the entire duration of the
+download. The commands below follow a common pattern to honour this rule:
+
+```mermaid
+flowchart LR
+    A["shared lock<br/>(read snapshot)"] --> B["release lock"]
+    B --> C["download / extract<br/>NO lock held"]
+    C --> D["exclusive lock<br/>(commit: rename + config write)"]
+    D --> E["release lock"]
+```
+
+The download phase produces files in a temporary directory; the commit phase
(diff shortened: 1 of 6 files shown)


davidanthoff (approved):
It all looks great to me!

I did not try to run it yet, but maybe we merge and push out something on the dev channel and see how it goes?

IanButterworth:
Great. I think we should get https://github.com/JuliaLang/juliaup/pull/1525 in and test this branch on it before merge.

topolarity:
Nice work on this @IanButterworth , it seems well thought-out

It does seem to leave `get_read_lock()` requesting [read-write permissions](https://github.com/JuliaLang/juliaup/blob/9da3103419aca7d9b03c5b1b1b772a58710612dd/src/config_file.rs#L229-L230) on the lockfile which is an issue when, e.g., using `juliaup` inside a read-only filesystem such as Claude's sandbox.

Do you think there's a chance to tighten those permissions so that read-only filesystems work?

davidanthoff:
@topolarity see https://github.com/JuliaLang/juliaup/issues/1532.

hraftery:
I guess the documentation URL should be:

https://github.com/JuliaLang/juliaup/blob/main/devdocs/locking_and_update_flows.md

Perhaps not primarily targeted at human eyes though.

IanButterworth:
Those are dev docs not user docs.

PR #1518, opened 2026-06-06
Add some more Juliaup GUI setup tweaks on Windows

davidanthoff:


fingolfin:
I'd approve this but I have absolutely no idea what this does, and no way to verify it. Perhaps you can at least quickly summarize what this does? Or if it helps, show a screenshot before and after?

IanButterworth:
I've assumed it's incomplete, otherwise David would just merge it into mine.

fingolfin:
So perhaps it should be a draft? Ah well

PR #1520, opened 2026-06-08
Remove self-update and version database update logic from julialauncher

### LINKED ISSUE #1519: Consider changing self update defaults
Automatic updates present operational and security risks, and I wonder if different defaults may be a safer starting point given the variety of different environments juliaup is used in. See https://github.com/JuliaLang/juliaup/pull/1514 for context.

The least surprising defaults to my mind, would be no self update unless explicitly asked for. An automatic self update when doing other updates to the environment may be reasonable.

I'd be keen to hear if others share the same philosophy, or maybe my use cases are different enough that this isn't a concern for most.

xgdgsc:
Vibe coded with cursor. An attempt to fix https://github.com/JuliaLang/juliaup/issues/1519 . Interesting thing is during the process. I only asked it to modify julialauncher to remove auto updating. It even auto changed juliainstaller auto updating config default to 0.

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -13,11 +13,6 @@ use juliaup::operations::{is_pr_channel, is_valid_channel};
 use juliaup::utils::{print_juliaup_style, resolve_julia_binary_path, JuliaupMessageType};
 use juliaup::version_selection::get_auto_channel;
 use juliaup::versions_file::load_versions_db;
-#[cfg(not(windows))]
-use nix::{
-    sys::wait::{waitpid, WaitStatus},
-    unistd::{fork, ForkResult},
-};
 use normpath::PathExt;
 #[cfg(not(windows))]
 use std::os::unix::process::CommandExt;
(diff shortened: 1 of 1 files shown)


davidanthoff:
That does not strike me as the right direction, we already have config options to disable auto-update. This also seems to replace the entire `fork` logic etc.

xgdgsc:
Why would launcher do anything online in the first place? It was extremely disappointing when you wake up one day and all your julia pipelines on all of your servers stuck.

davidanthoff:
Two things happen in the background: 1) the database that holds which Julia versions exist is updated, and 2) on systems where Juliaup was installed with the shell script, Juliaup itself auto-updates.

1) is needed so that we can show the kind of update prompts we show at the moment about new versions of Julia being available. For that to work we need to know about new Julia versions.

2) kind of mimics what happens on other systems where Juliaup is distributed via an OS level update mechanism. On Windows, for example, Julia is distributed via the app store and that also auto-updates.

Generally, the philosophy is that we _never_ update any Julia version itself without explicit opt-in by users and at a moment when a user wants that. But the Juliaup machinery itself is more like an app on a phone that just continuously stays up-to-date.

Issue #805, opened 2024-01-25
Add some tests that exercise an upgrade scenario


davidanthoff:
https://github.com/JuliaLang/juliaup/pull/601 had a major bug in terms of upgrades, but the core issue here really is that we don't have any tests for that. Given how important it is that things don't break when upgrading from an existing version we really need to add tests for that.

I don't really have a good idea on how to test the upgrade itself, but what we can fairly easily test is that a new version of Juliaup at least doesn't crash if it has to deal with an old schema config file (that was the problem we had with that PR). I'll take a stab at that.

davidanthoff:
Ideally we would test two scenarios: 
1. Update from a previous version
2. Update to a newer (not yet existing) version

The former seems much easier, but still involved.

One question is whether we need to set up a test-only whole story for the servers that hold the latest version info etc. I think if we really want to test end-to-end, probably yes... I think the best strategy for that would be to have the ability (via an env var) to point Juliaup to a custom server, and then as part of the tests spin up a custom server that serves stuff just for this test run. Things like SSL certificates then become an issue...

PR #1525, opened 2026-06-09
Add end-to-end self-update integration tests

IanButterworth:
Implements the first part of https://github.com/JuliaLang/juliaup/issues/805#issuecomment-4634996695

Claude:

---

Adds blind integration tests that exercise the real self-update code paths against a local mock server.

**`self_update_end_to_end`** — the explicit `juliaup self update` command:
  1. installs an isolated copy of juliaup,
  2. disables automatic self-update,
  3. sets up state (a system channel, a linked channel, an alias),
  4. mocks a newer juliaup version on a loopback server,
  5. runs `juliaup self update`,
  6. verifies juliaup/julia still work and the state survived.

**`self_update_auto_triggered_by_launcher`** — the automatic path: with `startupselfupdateinterval` elapsed, running the `julia` launcher forks a background daemon that spawns `juliaup self update` (inheriting the mock-server env) and then execs into Julia. The test waits for the detached update to finish and verifies the launcher symlink is restored and julia still runs.

To let the tests point juliaup at a local mock server, allow plain HTTP for loopback hosts (localhost / 127.0.0.1 / ::1) in `get_juliaserver_base_url` and `get_julianightlies_base_url`. HTTPS remains required for all non-loopback servers, so this only affects traffic that never leaves the machine.

The tests need the `selfupdate` feature, which changes config loading in a way incompatible with the other integration tests, so they run in their own CI job (`test-selfupdate`) on Linux and macOS.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -285,3 +285,33 @@ jobs:
           df -h
           export CARGO_INCREMENTAL=0
           cargo test --verbose --target ${{matrix.target}} --features ${{matrix.features}}
+
+  test-selfupdate:
+    # Exercises the real `juliaup self update` code path end-to-end against a
+    # local mock server (see tests/command_selfupdate_test.rs, issue #805).
+    # Runs in its own job because it needs the `selfupdate` feature, which
+    # changes config loading in a way that is incompatible with the other
+    # integration tests.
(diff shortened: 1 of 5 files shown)


IanButterworth:
I'll merge this as I think it is complete, and at least moves us in a good direction.

Issue #400, opened 2022-09-12
Don't delete running Julia install


ericphanson:
From slack:

> I updated to 1.8.1 in a terminal, but had 1.8.0 running in another terminal. Then I did a Pkg operation there and I got a process failed error when it tried to shell out to /Users/eph/.julia/juliaup/julia-1.8.0+0.aarch64/libexec/7z. IMO it would be nice to not immediately delete the old version but rather clean it up later when it is not in use.

One thought as to a solution:

> since I’m launching julia via juliaup, maybe it can be aware of which versions are running, e.g. by writing a PID file or such

davidanthoff:
> since I’m launching julia via juliaup

One complication is that clients like the VS Code extension are _not_ launching via the `julialauncher`, they are getting the direct launch command from the `juliaup` API and then bypass `julialauncher`. Maybe we should change that, we could then write such a PID file. Or we do the wait-a-week thing. Or maybe there is also just a way to list all running processes and just somehow inspect them whether they are a Julia process?

PR #1533, opened 2026-06-11
Make julialauncher config reads lock-free

IanButterworth:
Claude:

----

Follow-up to #1517. That PR shrank how long the exclusive configuration lock is held; this one removes the launcher from the lock contention picture entirely: **launching `julia` can no longer block on the configuration lock, ever.**

## Why this is safe

The launcher only ever took a *shared* lock, inside `load_config_db`, and released it before resolving the binary path and exec'ing Julia. The only thing that lock bought was a guarantee of not reading a half-written `juliaup.json`. Since all config writes go through an atomic temp-file + rename (`save_config_db`), a lock-free reader always sees either the old or the new file in full — no lock needed.

One hole had to be closed to make that invariant total: the *initial* config file was created by opening with `create(true)` and writing JSON in place, so a lock-free reader racing first-time setup could observe an empty or partial file. The initial config is now also created via temp-file + rename (`create_initial_config_file`), so an incomplete `juliaup.json` can never be observed on disk.

## Changes

- New `load_config_db_lockfree`; `julialauncher` uses it for both config reads (startup and post-auto-install reload). The locked `load_config_db` is unchanged for all `juliaup` commands.
- `load_mut_config_db` creates the initial config atomically instead of writing in place (it still holds the exclusive lock while doing so).
- A zero-length config file (possible leftover from an interrupted initial setup by an *older* juliaup) is treated like a missing file rather than a parse error.
- `devdocs/locking_and_update_flows.md` documents the lock-free read path and adds the design rule that all `juliaup.json` writes must be atomic replacements, since that invariant is now load-bearing.
- Unit tests, including a regression test that a lock-free read succeeds while the exclusive lock is held.

## What this does and doesn't fix

With this change the launcher cannot be stalled even by the remaining long exclusive-lock holders (`remove`/`gc` deleting a large install under the lock). It does *not* change the situation in #400 (an in-use installation being deleted) — but it also doesn't meaningfully widen that race: the lock was already released before exec, so the unprotected window between config read and launch existed before and is unchanged. #400 remains the complementary fix, planned separately along with moving directory deletion out from under the exclusive lock (rename-to-trash).

### DIFF
--- a/devdocs/locking_and_update_flows.md
+++ b/devdocs/locking_and_update_flows.md
@@ -43,6 +43,12 @@ path (`load_config_db_lockfree`, used by `julialauncher`) safe — a reader
 always observes either the old or the new file in full, never a torn write.
 Never write to `juliaup.json` in place.
 
+On Windows the rename must go through `std::fs::rename`
+(`FILE_RENAME_FLAG_POSIX_SEMANTICS`), not `tempfile`'s `persist()`
+(`MoveFileExW`): the latter fails while any process has the destination open,
+and with lock-free readers a writer's commit can now overlap a read. See
+`persist_atomically` in `config_file.rs`.
+
 > **Never hold the exclusive lock across a network operation.**
 
 Downloads can be slow or hang. A writer that holds the exclusive lock across a
--- a/devdocs/locking_and_update_flows.md
+++ b/devdocs/locking_and_update_flows.md
@@ -17,6 +17,7 @@ modes:
 | --- | --- | --- |
 | **Shared** (read) | `get_read_lock` / `load_config_db(paths, None)` | only an exclusive holder |
 | **Exclusive** (write) | `load_mut_config_db` | any other shared or exclusive holder |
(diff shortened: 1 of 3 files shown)


topolarity:
Hooray! You may need to use `FILE_RENAME_FLAG_POSIX_SEMANTICS` to avoid readers blocking the rename on Windows and causing `persist()` to fail (open read handles on the destination file block renames on Windows by default)

davidanthoff (approved):
Very nice!

IanButterworth:
@topolarity look good now?

topolarity:
Yes, I think that should resolve it!

Only other thought is that despite what the OP says w.r.t. #400, it does seem like the new behavior makes that window somewhat worse since you will no longer be blocked from loading an install during the time that its files are being deleted

But that may be handled best in a follow-up tackling that specific issue?

IanButterworth:
Yeah. I do think we want to close #400 but the issue with vscode poking through the juliaup api to binaries directly makes it harder to solve. Need @davidanthoff's input there.

We can merge this for now at least.

PR #1535, opened 2026-06-12
Wrap more into interactive startup updates in julialauncher

### LINKED ISSUE #1519: Consider changing self update defaults

xgdgsc:
- Introduced a new function `run_startup_updates` to handle both version database updates and self-updates when the launcher is invoked in interactive mode.
- Modified the process for forking and executing the Julia process to ensure updates are only triggered in interactive sessions.
- Added tests to verify that self-updates are correctly triggered in interactive mode and not in non-interactive invocations (e.g., `julia -e`).
- Enhanced the test suite to simulate interactive terminal sessions for more accurate testing of the launcher behavior.

fix https://github.com/JuliaLang/juliaup/issues/1519

### DIFF
--- a/src/bin/julialauncher.rs
+++ b/src/bin/julialauncher.rs
@@ -143,32 +143,31 @@ fn run_selfupdate(_config_file: &juliaup::config_file::JuliaupReadonlyConfigFile
 }
 
 fn is_interactive() -> bool {
-    // First check if we have TTY access - this is a prerequisite for interactivity
-    if !std::io::stdin().is_terminal() || !std::io::stderr().is_terminal() {
-        return false;
-    }
-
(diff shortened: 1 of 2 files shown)


davidanthoff:
I don't agree with this approach. If version db and self-updates are configured to happen, then it should be triggered regardless of whether the launcher is started in interactive mode or not.

PR #1538, opened 2026-06-12
Resolve PR channels via the ephemeral staging bucket

Keno:
This is a WIP PR prepared by Claude to go along with the CI infrastructure refactor - I have not yet reviewed it in detail. Claude commit message follows:

Julia's CI is moving PR builds off the nightlies bucket (JuliaCI/julia-buildkite#544): pull request builds are now staged write-once into a dedicated ephemeral bucket, keyed by the head commit sha of the PR:

    https://julialang-ephemeral-pr.s3.amazonaws.com/bin/<head sha>/julia-<10-char short sha>-<os>-<arch>.tar.gz

Teach `juliaup add pr<N>` (and `juliaup update`) to resolve the PR number to its head sha via the GitHub API (honoring GITHUB_TOKEN / GH_TOKEN for authenticated requests) and to download from that location. If the staged build does not exist (pre-crossover PRs), fall back to the legacy `julia-pr<N>-<os>-<arch>.tar.gz` scheme in the nightlies bucket; the fallback is clearly marked for removal once those objects have aged out. If neither location has binaries, explain that PR builds expire (~90 days) and that re-running CI stages fresh ones.

Because the staged location moves with every push to the PR, the update check now re-resolves PR channel URLs instead of reusing the URL recorded at install time, and `update_version_db` records the re-resolved URL alongside the refreshed etag.

The new scheme also adds linux-i686 PR channel support (staged builds exist for every platform CI builds), and a JULIAUP_PR_SERVER override analogous to JULIAUP_NIGHTLY_SERVER.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ The available system provided channels are:
 - `rc`: same as `beta`, but only starts with release candidate versions.
 - `nightly`: always points to the latest build from the `master` branch in the Julia repository.
 - `x.y-nightly`: always points to the latest build from the `release-x.y` branch in the Julia repository, e.g. `1.11-nightly` gives the latest build on the `release-1.11` branch`.
-- `pr{number}` (e.g. `pr123`): points to the latest successful build of a PR branch (https://github.com/JuliaLang/julia/pull/{number}). Only available if CI has recently and successfully built Julia on that branch.
+- `pr{number}` (e.g. `pr123`): points to the latest successful build of a PR branch (https://github.com/JuliaLang/julia/pull/{number}). Only available if CI has successfully built Julia on that branch within roughly the last 90 days: PR builds expire, but re-running CI on the pull request uploads fresh ones.
 - specific versions, e.g. `1.5.4`.
 - minor version channels, e.g. `1.5`.
 - major version channels, e.g. `1`.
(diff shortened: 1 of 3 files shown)


davidanthoff:
Also didn't review it carefully, but maybe the right thing to do is to generally move to this kind of strategy for all direct downloads, where instead of relying on `etag` info to detect stale versions we always check git commits for a given branch/PR, store that in the local config and then we can decide whether a new version exists based on that. The `etag` stuff is not great and has been a bit fiddly, and also doesn't work great with mirrors etc.

IanButterworth:
What remains for this? It'd be nice to have the pr functionality back.

Keno:
We haven't turned on the new julia-pr pipeline yet, so these builds aren't being staged yet.

Keno:
julia-pr pipeline is now enabled, so we should start thinking about this

fingolfin (approved):
This looks quite sensible to me. I won't claim to have read every line, but looked the broad strokes and several specific parts of the code.

The code grows a lot in size, which my impression is due to added error checks: before we just constructed a URL and fired it up, hoping for the best; now, a lot more effort is done to do things cleanly and properly, and to catch errors at intermediate stages. Hopefully this will improve the user experience (and making debugging easier, too).

There is some minor code duplication (e.g. for handling the HTTP status codes, once under Windows and once elsewhere), but I think that's harmless and it's not clear to me that trying to refactor this will be beneficial.

Thanks!

Keno:
Alright, let's try it. I think the people want their pr channels back.

PR #1547, opened 2026-07-06
Report channel symlink destinations

### LINKED ISSUE #1431: Where does `channelsymlinks` install symlinks / how to configure this?
I was wondering where `channelsymlinks` creates symlinks. I found no documentation, so I just tried it:
```
./juliaup config channelsymlinks true
    Creating symlink julia-release for Julia 1.12.5+0.x64.linux.gnu
    Creating symlink julia-1.12 for Julia 1.12.5+0.x64.linux.gnu
   Configure Property 'channelsymlinks' set to 'true'
```
As in my other recent issues, it would be so nice if this actually told me *where* it created those...

But I eventually found out: in ` ~/.local/bin/`, and their names are `julia-CHANNEL`. Fine. But is this always the path it uses? Can it be influenced somehow?

fingolfin:
Document channel symlink names and JULIAUP_BIN_DIR selection.

Co-authored-by: Codex <codex@openai.com>

Resolves #1431

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,18 @@ To launch the default Julia version simply run `julia` in your terminal.
 
 To launch a specific Julia version, say in channel `release`, run `julia +release`.
 
+On Unix-like platforms, `juliaup config channelsymlinks true` creates a
+separate executable name for each installed channel. These entries are named
+`julia-<CHANNEL>`, for example `julia-release`, and point to the Julia version
+for that channel.
+
+Juliaup creates these channel symlinks in the directory returned by
+`JULIAUP_BIN_DIR` when that environment variable is set. If `JULIAUP_BIN_DIR`
+contains multiple entries separated like `PATH`, Juliaup uses the first entry.
+When `JULIAUP_BIN_DIR` is not set, Juliaup uses the directory containing the
+running `juliaup` executable if that directory is inside the user's home
+directory; otherwise it uses `~/.local/bin`.
+
 ## Overrides
 
 The Julia launcher `julia` automatically determines which specific version of Julia to launch. There are several ways to control and override which Juliaup channel should be used:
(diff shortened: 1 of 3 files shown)


IanButterworth (approved):

README.md
```diff
@@ -153,6 +153,18 @@ To launch the default Julia version simply run `julia` in your terminal.
...
+separate executable name for each installed channel. These entries are named
+`julia-<CHANNEL>`, for example `julia-release`, and point to the Julia version
+for that channel.
+
+Juliaup creates these channel symlinks in the directory returned by
+`JULIAUP_BIN_DIR` when that environment variable is set. If `JULIAUP_BIN_DIR`
+contains multiple entries separated like `PATH`, Juliaup uses the first entry.
+When `JULIAUP_BIN_DIR` is not set, Juliaup uses the directory containing the
+running `juliaup` executable if that directory is inside the user's home
+directory; otherwise it uses `~/.local/bin`.
```
fingolfin, line 166: In light of the discussion at #1556, it seems `JULIAUP_BIN_DIR` is a rather obscure feature and may intentionally be undocumented (and perhaps should be removed outright?)

It'd be good to hear from @davidanthoff and @ghyatzo ...?

fingolfin: ping @davidanthoff @ghyatzo

ghyatzo: I don't think I have the authority to say anything final in this matter. At the time I had just pointed out that it was still undocumented so it was unclear if it was intended to stay or to be deprecated. I think that it's been enough time that probably even if undocumented it would be too breaking removing this.

The name is pretty confusing imho, but maybe we can use the same location as an installation location also for juliaup and julialauncher binaries, but haven't explored this idea very thoroughly yet.

PR #1548, opened 2026-07-07
Document more installer self-update options

### LINKED ISSUE #1432: Make "secret" installer options public and documented?
Searching through the sources here, I discovered I can do this:
```
curl -fsSL https://install.julialang.org | sh -s -- \
    --path $JULIAUP_DEPOT_PATH \
    --yes \
    --add-to-path=no \
    --background-selfupdate=0 \
    --startup-selfupdate=0 
```
The last three options are undocumented. I'd love to use them, but am worried that they might go away one day, as they are undocumented.

It would be great if those could be declared "official" (i.e. won't be removed over night), and documented. I can provide a PR for the latter, but obviously project leadership has to do the former.

fingolfin:
Document `--add-to-path`, `--background-selfupdate` and `--startup-selfupdate` options in the README; improve their documentation in the usage text printed by the script.

Co-authored-by: Codex <codex@openai.com>

Resolves #1432


IanButterworth:
I think this is a call for @davidanthoff

davidanthoff (approved):
Looks good to me!

The language here now is slightly different from the install manual (https://docs.julialang.org/en/v1/manual/installation/#Command-line-arguments), maybe we could copy the language from this PR over to there as well?

PR #1556, opened 2026-07-18
Centralize where paths are discovered v2

ghyatzo:
Thanks to @fingolfin that reminded me that I had licked the cookied but hadn't eaten it yet.

Replace the scattered get_juliaup_path() / current_exe() calls with two new GlobalPaths fields — juliaupselfexecfolder and juliaupselfexec — computed once via dunce::canonicalize in get_paths() and reused everywhere.

Supersedes: #915, #886

some additional small changes on top of #915 regarding `JULIAUP_BIN_DIR`. I renamed the function `get_bin_dir` to `get_channel_link_dir` to clarify it's real purpose.

There is a bit of a mismatch with having the ability to specify `JULIAUP_DEPOT_PATH` through  env var, but not the binary location (other than sourcing .bashrc) but that can be a separate discussion, along with keep relying or not on `current_exe` semantics instead of more external authorities (would help with the whole `julialauncher` calling `juliaup` trick).

Discussions on whether the undocumented `JULIAUP_BIN_DIR` still has its place or not are also out of scope, although it ties pretty close with the above since the name is pretty confusing. should it be a thing on its own? all binaries and symlinks in the same folder?

### DIFF
--- a/src/bin/juliainstaller.rs
+++ b/src/bin/juliainstaller.rs
@@ -432,9 +432,9 @@ pub fn main() -> Result<()> {
         }
     }
 
-    let juliaupselfbin = install_choices.install_location.join("bin");
+    let juliaupselfexecfolder = install_choices.install_location.join("bin");
 
-    trace!("Set juliaupselfbin to `{:?}`", juliaupselfbin);
+    trace!("Set juliaupselfexecfolder to `{:?}`", juliaupselfexecfolder);
 
     println!("Now installing Juliaup");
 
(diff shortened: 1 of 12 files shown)


fingolfin:
@ghyatzo CI fails, already the `rustfmt` check, but there is also
```
error[E0063]: missing fields `juliaupselfexec` and `juliaupselfexecfolder` in initializer of `global_paths::GlobalPaths`
Error:    --> src/config_file.rs:750:9
    |
750 |         GlobalPaths {
    |         ^^^^^^^^^^^ missing `juliaupselfexec` and `juliaupselfexecfolder`
```

ghyatzo:
Yeah, I had a misconfigured IDE without the proper features enabled.
Hadn't had the time to revisit yet.
Will do within the week. Apologies.

ghyatzo:
done

PR #1573, opened 2026-08-17
Fix concurrent add dropping downloaded version

KristofferC:
This has gotten little review of my own but putting it up at @IanButterworth request.

## Summary

- make the post-download `juliaup add` conflict check version-aware
- commit and register the downloaded version when a concurrent system channel points at a different version, then move the channel to the requested version
- discard true duplicate downloads while preserving concurrent linked, direct-download, and alias channel name claims
- add regression tests for each conflict policy and document the updated locking flow

This fixes the race introduced by #1517 where `add` could print `Installing ...` followed by `already installed`, delete the completed temporary install, and leave the channel on an older version.


IanButterworth (approved):
Thanks

PR #1591, opened 2026-09-10
Add startmenu items for Juliaup GUI to Windows installers

davidanthoff:
I'm going to merge right away, the only way to test this is to actually try to release a new version.


DilumAluthge:
I don't suppose we can also create a shortcut on the desktop?

davidanthoff:
> I don't suppose we can also create a shortcut on the desktop?

We really shouldn't :) MS has discouraged that forever, and the modern day installer tools based on MSIX don't support it at all.

PR #1592, opened 2026-09-11
fix: stream Windows downloads so progress bar updates live

xgdgsc:
## Summary
- Fix Windows download progress appearing only after the transfer finished (#1281) by using `HttpCompletionOption::ResponseHeadersRead` so the body streams during unpack.
- Use `Partial` stream reads and read `Content-Length` for a fixed progress total (avoids `TryComputeLength` returning 0 under streaming, which made `{total_bytes}` grow with each chunk).
- Draw the bar at `0/total` immediately and finish it cleanly after unpack.

## Test plan
- [x] Built `juliaup` for `x86_64-pc-windows-msvc` and ran `juliaup add` on Windows
- [x] Confirmed progress bar appears during network transfer (not only after download completes)
- [x] Confirmed `{bytes}/{total_bytes}` uses a fixed total from `Content-Length` while downloaded bytes climb
- [ ] CI passes on this PR

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/operations.rs
+++ b/src/operations.rs
@@ -507,8 +507,15 @@ pub fn download_extract_sans_parent(
     let request_uri = windows::Foundation::Uri::CreateUri(&HSTRING::from(url))
         .with_context(|| "Failed to convert url string to Uri.")?;
 
+    // Use ResponseHeadersRead so the body streams as we unpack. The default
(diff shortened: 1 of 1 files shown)


IanButterworth:
Closes #1282

IanButterworth (approved):
Thanks!

PR #1595, opened 2026-09-11
Skip download in `add` when the version is already installed

KristofferC:
`juliaup add` only checked whether the *channel* name was installed before downloading, never whether the *version* it resolves to already is. So after `juliaup update release` installed 1.13.0, both `juliaup add 1.13` and `juliaup add 1.13.0` downloaded the full archive again, only for `commit_version_install` to silently discard it. This also hits the launcher's channel auto-install, which shells out to `juliaup add`.

Now `add` checks `installed_versions` during the existing shared-lock pre-check and, when the version is already installed, skips the download and only commits the channel pointer. If the version is removed concurrently (e.g. `juliaup gc`) between that check and the exclusive-lock commit, it drops the lock, downloads after all, and re-acquires, so the lock is still never held during a download.

### DIFF
--- a/src/command_add.rs
+++ b/src/command_add.rs
@@ -10,7 +10,7 @@ use crate::operations::{
 };
 use crate::utils::{print_juliaup_style, JuliaupMessageType};
 use crate::versions_file::load_versions_db;
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use regex::Regex;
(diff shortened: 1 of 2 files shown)


IanButterworth:
I think it used to be this way.. if so I'm surprised we didn't have tests that caught it.

IanButterworth (approved):
Seems this regressed in #1517 (v1.20.5). I could've sworn we had tests that checked for this.

PR #1596, opened 2026-09-11
ci: ship juliaupgui in the Linux archives

### LINKED ISSUE #1593: GUI binary not found on homebrew installed juliaup
I run juliaup 1.22.3 installed via brew on macOS 15.7.8 but the GUI does not launch. 

```
> juliaup gui
Error: GUI binary not found at '/opt/homebrew/bin/juliaupgui'. The juliaupgui component may not be installed.
```

IanButterworth:
Claude:

---

Linux installs of juliaup have never included the GUI. `juliaup-init.sh` always selects the musl build, and the release workflow skipped `juliaupgui` for musl targets because the static musl binary cannot dlopen the X11/Wayland/OpenGL libraries the GUI needs at runtime. The gnu archives did carry a GUI, but it was built in the cross `:edge` images (Ubuntu 24.04) and so requires glibc 2.39, which would not have helped most users either.

This builds the Linux GUI in pinned manylinux_2_28 containers instead, natively on x86_64 and arm64 runners, so it runs on any distribution with glibc >= 2.28 (2018 onwards, RHEL 8 included). A new composite action does the build and fails if the binary's glibc floor creeps above 2.28. The release workflow runs it as a job and packs the result into the gnu, musl and portable Linux archives, so the installer and `juliaup self update` deliver it. The same action runs in `test.yml` on both architectures.

The binary produced in CI on this branch needs only `libc`, `libm`, `libdl`, `libpthread` and `libgcc_s`; everything else is loaded on demand.

Also in this change: the portable macOS and Windows archives are no longer excluded from the GUI build, the cross images drop the X11 packages that nothing builds with anymore, and `juliaup self uninstall` removes `juliaupgui`.

The release workflow only runs on tags, so the packaging steps are reviewed by reading rather than exercised.

Fixes #1593

### DIFF
--- a/.github/actions/build-juliaupgui-linux/action.yml
+++ b/.github/actions/build-juliaupgui-linux/action.yml
@@ -1,0 +1,47 @@
+name: Build juliaupgui for Linux
+description: >
+  Builds juliaupgui against an old glibc and checks the result. Must run in a
+  job whose `container` is a manylinux_2_28 image (AlmaLinux 8, glibc 2.28)
+  matching the target architecture.
+
+  Linux installs of juliaup use the static musl build of the CLI, which cannot
+  dlopen the X11/Wayland/OpenGL libraries the GUI needs at runtime. The GUI is
+  therefore built dynamically against glibc here and shipped alongside the
+  musl binaries, so it has to keep running on any glibc >= 2.28 distribution.
+inputs:
+  target:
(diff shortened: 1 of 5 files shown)


IanButterworth:
Planning to merge this when green and get it tried out on releasepreview along with the other fixes ASAP.

PR #1600, opened 2026-09-14
Discover nightly builds and variants through `nightlies.json`

maleadt:
Use the catalog introduced by [VersionsJSONUtil.jl#107](https://github.com/JuliaLang/VersionsJSONUtil.jl/pull/107) to discover nightly channels and their downloads. Juliaup currently constructs nightly URLs and lists an `x.y-nightly` placeholder; the catalog lets it show the channels and builds available for the user's platform and report unavailable selections before downloading Julia.

Build variants become ordinary named channels:

```sh
juliaup add nightly+opt
julia +nightly+opt
juliaup add 1.13-nightly+nogpl~x64
```

`juliaup list` shows the available variants, such as `opt`, `assert`, and `nogpl`. The grammar is `<channel>[+<variant>...][~<arch>]`; combined variants select a matching build if the catalog provides one. Variant order does not affect which artifact is selected, while installed channels retain the name supplied by the user. Release and PR variants are not supported yet.

The suffix makes the build choice usable with the existing launcher, defaults, overrides, and aliases, without changing the configuration format. It follows [uv's `+variant` convention](https://docs.astral.sh/uv/concepts/python-versions/#requesting-a-version) and keeps the architecture last, as [rustup does with the host](https://rust-lang.github.io/rustup/concepts/toolchains.html).

The catalog is cached atomically and refreshed by installs and update checks. Failed refreshes preserve the cache; listing falls back to generic placeholders when no catalog is available. Manifest detection uses the cache to select a release-series nightly with a compatible standard build, falling back to `nightly`. Launching does not fetch the catalog synchronously.

Mirrors serve `bin/nightlies.json` through `JULIAUP_SERVER`. `JULIAUP_NIGHTLY_SERVER` redirects official nightly URLs; artifacts on other hosts keep their catalog URLs. ETag support is checked on the actual artifact, so mirrors need not provide directory listings.

Nightly updates still use the URL recorded at installation. On macOS, installation retains the existing DMG attempt with tarball fallback. Resolving changed nightly URLs on update and selecting DMGs directly from the catalog remain follow-up work.

Assisted by: Fable 5.1, Astra 6


maleadt:
My personal motivation for this all:

```
Benchmark 1: julia +nightly -e 'Base.compilecache(Base.identify_package("Pkg"))'
  Time (mean ± σ):     23.225 s ±  0.128 s    [User: 63.916 s, System: 1.104 s]
  Range (min … max):   23.098 s … 23.463 s    10 runs

Benchmark 2: julia +nightly+opt -e 'Base.compilecache(Base.identify_package("Pkg"))'
  Time (mean ± σ):     19.759 s ±  0.190 s    [User: 51.534 s, System: 1.154 s]
  Range (min … max):   19.575 s … 20.119 s    10 runs
```

But I've heard people would really like an `+asserts` build, as well as `+nogpl`, all of which now work. Over time, we'll want to have this for releases too, but let's start with nightlies.

IanButterworth:
Is opt with BOLT?

maleadt:
BOLT, PGO and LTO, so we'll be able to roll some of that out to other platforms too over time.

DilumAluthge:
Could you add an example to the PR body for the syntax of multiple variants?

E.g. is it `julia +nightly+assert+opt`? Or some other syntax?

IanButterworth:
Food for thought (fable 5.1.. funny how different agents with different context find different things)

---

Bugs worth fixing before merge

- Manifest detection regresses without a cache (src/version_selection.rs:610). On main, a manifest stamped with a too-new patch version picks x.y-nightly unconditionally. Now it picks master nightly whenever nightlies.json is absent, since the launcher only reads the cache and never downloads. Fresh installs, offline machines, and mirrors without the catalog hit this, even when x.y-nightly is already installed. With auto-install on, the wrong channel gets installed. Keeping the versioned channel when the catalog is unknown would fix it.
- Variant spelling is not canonicalized at install (src/command_add.rs:241). add nightly+opt+assert stores the verbatim key, but list and the GUI render the sorted canonical form nightly+assert+opt, so the installed build shows as not installed, julia +nightly+assert+opt prompts to install it again, and a second spelling creates a second folder for the same artifact. Using the canonical name as the config key fixes it.
- One bad catalog key breaks list and the GUI (src/operations.rs:1083). Keys are parsed with ?, so any key the grammar rejects (empty, containing ~, a trailing +) aborts juliaup list and empties the GUI's channel view. The module doc promises lenient parsing. Skipping unparseable keys, or building ChannelName directly rather than rendering strings and re-parsing, removes the failure.
- Launcher offers installs that always fail (src/bin/julialauncher.rs:380). The auto-install condition uses the new is_pr_channel, which accepts pr123+opt, while is_valid_channel and add reject PR variants. So julia +pr123+opt prompts, spawns an install that bails, and with auto-install on does so on every launch. channel_valid alone already covers PR and nightly syntax.

Behavior changes to weigh

- Existing JULIAUP_SERVER mirrors break (src/operations.rs:1652). Nightly installs now require bin/nightlies.json from the mirror with no fallback to the old URL scheme. Previously mirrors only needed the versiondb. README documents it, but it needs at least a release note; a fallback when the catalog is missing would avoid the break.
- list does synchronous network I/O with no timeout on Windows (src/operations.rs:761). With no cache, list fetches the catalog. Non-Windows caps it at 30 s, the Windows path has no timeout at all. Pre-PR list was offline.
- ETag check moved after the full download (src/operations.rs:1176). A mirror without ETags now downloads and extracts the whole tarball, plus a DMG attempt on macOS, before failing. A HEAD on the artifact URL first keeps both properties.
- Loopback http mirrors rejected (src/operations.rs:1677). The scheme check refuses non-https catalog URLs, but the base-URL helpers explicitly allow loopback http. The PR's own test keeps the official https URL to sidestep this.

Minor

- GUI tree does not nest +variant rows under their base channel, since + is not in the separator set (juliaupgui/src/app.rs:1516).
- list never refreshes an existing cache, so it can advertise channels that add, which always refreshes, then rejects. Bounded to about a day by the launcher's background refresh.

davidanthoff:
Thanks for this — the general direction of discovering nightlies through a catalog instead of hardcoding URL patterns is one I like a lot. Two design-level concerns before the details get locked in:

**1. Catalog download frequency**

We went to some lengths to keep the recurring background traffic minimal: the periodic `update-version-db` run first fetches only the tiny `RELEASECHANNELDBVERSION` stamp and downloads the full versions db only when it has actually changed, and the direct-download ETag checks only do network work for channels the user actually has installed.

As written, this PR adds an unconditional full download of `nightlies.json` to every `update_version_db` run with `update_direct_downloads = true` — which includes the background refresh that every juliaup installation performs on its regular schedule. So every user downloads the complete catalog every time, even if they've never installed a nightly, with no conditional request. The file is small, but multiplied across the install base that's real traffic, and it abandons the check-before-download discipline the rest of the update path follows.

I think this needs at least one of (ideally the first plus one of the others):

- gate the background refresh on the user actually having a nightly channel installed (`juliaup add` already fetches fresh, so discovery still works),
- a conditional GET using a stored ETag, and/or
- a version stamp alongside the catalog, mirroring the `RELEASECHANNELDBVERSION` mechanism.

The cost of gating is only that `juliaup list`'s variant display and manifest auto-resolution see a staler cache for users without nightlies, which seems fine to me.

**2. Who decides what channel names exist**

The PR puts a fair amount of policy client-side: the variant grammar in `channel_name.rs` (sorting/dedup so `nightly+opt+assert` equals `nightly+assert+opt`), the artifact-matching in `NightliesDb::select`, and the hardcoded triplet whitelist. That means a new platform or a twist in variant naming needs a juliaup release even though the catalog already describes it, and the client's canonicalization rules can drift from the server's vocabulary. I also don't think the offline story justifies keeping this client-side: if someone is offline with no cached catalog, "go online" is an acceptable answer.

I'd prefer the model we use for regular channels, where the name is essentially an opaque identifier the server hands out. And we already have the infrastructure for exactly this: there is an existing process that takes the VersionsJSONUtil.jl output and transforms it into juliaup's own format. We could run the nightlies catalog through that same pipeline — take what VersionsJSONUtil.jl produces today, generate the channel names server-side exactly how we want them (per channel, the available variant suffixes in a canonical order, e.g. `nightly` declaring `+opt`, `+assert`, `+assert+opt`, `+nogpl` with their artifacts), and publish the result alongside the other juliaup-generated files rather than consuming a raw `bin/nightlies.json`. That's also just cleaner layout-wise: everything juliaup consumes lives in one place, in one format regime we control, and it requires no changes to VersionsJSONUtil.jl at all.

I would still publish it as a separate nightly version db story, so that users that don't use nightlies are completely shielded from the version churn there.

The client then just splits base + suffix and exact-matches against the published list — a lookup, not a grammar. That keeps the server in charge of what exists, dissolves the permutation-equivalence question (the canonical order is whatever the server publishes and `juliaup list` shows), avoids the cross-product bloat that enumerating full flat names would cause if variants later extend to release channels, and leaves only the trivial base-name split and `pr<number>` handling client-side.

The one new requirement is a way to regenerate that file when nightlies change, since that happens independently of versions.json updates — but a simple workflow-triggered GitHub Action would probably do.

maleadt:
@DilumAluthge yes, `julia +nightly+assert+opt` is the syntax, if that combination is published. I'll document the ordering too: use the exact spelling shown by `list`; another permutation won't be accepted. No combined build is currently published.

---

@IanButterworth I'll fix the manifest fallback for too-new patch versions, unsupported PR-variant install prompts, malformed metadata breaking listing, embedded loopback URLs, GUI nesting, and the Windows timeout.

Server-defined names will make listing and installation use the same spelling. Listing won't reparse those names, and a bad refresh will preserve the previous cache. I'll refresh listing metadata after 24 hours, with a five-second timeout and cached fallback.

For ETags, we can check GET headers inside the downloaders before extraction or mounting, including the DMG attempt. That avoids both the wasted download and an extra HEAD request. Mirrors will need the new metadata; I'll document that requirement rather than keep the old URL-discovery fallback.

---

@davidanthoff Sure. I'll move naming and platform selection into juliaup's publishing pipeline, with a separate target-specific nightly database retaining the base/suffix structure you suggest. New variants on supported platforms will flow through generically, with ABI compatibility checked server-side.

I'll gate periodic nightly checks on installed nightlies and use a tiny version stamp plus immutable database files, following the release-db approach. `add` will check the stamp; `list` will check missing or aged metadata. Installed channels remain launchable offline.

Juliaup will publish metadata daily, with manual dispatch available and unchanged output skipped. This affects discovery of available builds; ordinary nightly rebuilds are still detected through artifact ETags independently of metadata publication. The existing deployment environments require approval, so I'll arrange unattended nightly deployment and publish the initial database before shipping the client change. The generator can read the upstream S3 origin to avoid relying on CDN invalidation.

davidanthoff:
@maleadt I thought a bit more about this :) I actually want to rework the entire version db story a bit down the road, and then we could presumably integrate the nightlies stuff into that as well. So, I think actually maybe for now it would be easier to just make sure the download only happens when someone actually has nightlies installed, but we could for now also just key of the existing jsonversion generated stuff, and then at a later point in one unified step do the updates that I want to make for the general version db and the nightlies thing. 

I think that would make it easier and less involved to get this PR to merge, right?

maleadt:
For those wanting to test this out: Optimized binaries are available for macOS now too:

```
Benchmark 1: JULIA_OBJCACHE=0 JULIA_DEPOT_PATH=$(mktemp -d): ./target/release/julia +nightly -e 'Base.compilecache(Base.identify_package("Pkg"))'
  Time (mean ± σ):     26.477 s ±  1.025 s    [User: 53.086 s, System: 1.285 s]
  Range (min … max):   25.223 s … 28.300 s    10 runs

Benchmark 2: JULIA_OBJCACHE=0 JULIA_DEPOT_PATH=$(mktemp -d): ./target/release/julia +nightly+opt -e 'Base.compilecache(Base.identify_package("Pkg"))'
  Time (mean ± σ):     23.576 s ±  0.938 s    [User: 44.200 s, System: 1.696 s]
  Range (min … max):   22.206 s … 25.355 s    10 runs
```

maleadt:
> I think that would make it easier and less involved to get this PR to merge, right?

Yes. The branch keeps consuming the existing nightlies.json directly. Periodic catalog downloads only happen when a nightly is installed, and fresh metadata is reused for 24 hours. That keeps this PR smaller and leaves the broader version-db redesign for later.

Should be ready for review again. Eager to get this out into people's hands!

DilumAluthge:
> yes, `julia +nightly+assert+opt` is the syntax, if that combination is published. I'll document the ordering too: use the exact spelling shown by `list`; another permutation won't be accepted. No combined build is currently published.

Could we allow a different permutation, as long as everything else is the same? E.g. could we allow both `julia +nightly+assert+opt` and `julia +nightly+opt+assert` (and have them mean the same thing)?

maleadt:
> Could we allow a different permutation, as long as everything else is the same?

Sure. It's pretty theoretical right now, since we don't actually have builds that combine variants, but I pushed a change that canonicalizes the variants everywhere so that we should be able to use differently-ordered strings everywhere.

maleadt:
Let's land https://github.com/JuliaLang/juliaup/pull/1615 first. Marked draft until I rebase on top of that.

davidanthoff:
I merged the FreeBSD thing.

I think I would prefer to not allow different permutations for the channel names. Eventually I hope to move this work into the generic version database stuff we have, and there the idea is very much that Juliaup does not know anything about the format of channel names, they are just opaque identifiers. If we allowed permutations, that would clash with that design.

And just for completeness: the platform syntax that we use (i.e. `release~x86`) would be added at the end, right?

